Issue #1, opened 2018-02-04
Can't build/run on macOS


get9:
Can't seem to build or run this project on macOS/darwin due to the `syscall.BindToDevice`. Searching around leads me to the SO_BINDTODEVICE sockopt which isn't present on darwin. Instead, others have used a variety of workarounds:
* https://trac.transmissionbt.com/ticket/2313
* http://atastypixel.com/blog/the-making-of-talkie-multi-interface-broadcasting-and-multicast/

insomniacslk:
Thanks for opening this issue. I haven't found an obvious solution as it seems to depend entirely on platform implementations. In the case of Linux we have the best guarantee, i.e. binding to a specific network device, while on macOS and FreeBSD the most we can do appears to be `bind`ing to a specific socket. This is a weaker contract than `SO_BINDTODEVICE` but I guess we can live with the limitation. I will try to get this out this week

insomniacslk:
This might help (as suggested by Marc, who has a better google-fu than I have), https://lists.apple.com/archives/darwin-dev/2014/Jan/msg00004.html .

From the post:
```
you can use IP_BOUND_IF to bind socket to the particular interface index.

something like
int idx = if_nametoindex( "en1");
setsockopt(s, IPPROTO_TCP, IP_BOUND_IF, &idx, sizeof(idx) );
```

insomniacslk:
this seems to work. I'll do some more testing and post a patch here

insomniacslk:
hey @get9 and @rb2k, I've created https://github.com/insomniacslk/dhcp/pull/2 to fix this. I am still testing, but feel free to have a preliminary look

PR #2, opened 2018-02-07
Added OS-specific implementations for binding to a network device

insomniacslk:
macOS has a different implementation than Linux in order to bind packets to a specific device. Added OS-specific implementations. Fixes https://github.com/insomniacslk/dhcp/issues/1 .


get9:
Checking out this branch, I was able to build and run:
```
package main

import (
	"fmt"

)

func main() {
	client := dhcpv4.Client{}
	conv, err := client.Exchange("en0", nil)
	if err != nil {
		fmt.Println(err)
	}

	for _, m := range conv {
		fmt.Print(m.Summary())
	}
}
```

I could view the whole DORA exchange, so I think we're good to go. Thanks for the quick fix!

PR #3, opened 2018-03-03
Add BSDP client support to DHCPv4

get9:
This PR adds client-side support for Apple's Boot Service Discovery Protocol (BSDP), which is kind of like bootp except different because of course it is. It uses DHCP option 43 (Vendor-specific information) plus a 4-packet exchange of information to bootstrap netbooting macOS hosts so they know what image to boot over the network.

The canonical reference for BSDP can be found [here](https://opensource.apple.com/source/bootp/bootp-359.30.2/Documentation/BSDP.doc), but can also be gleaned from the [source code for bootp](https://opensource.apple.com/source/bootp/bootp-359.30.2/).

Unit tests have been supplied to ensure correctness of the protocol implementation.

On a running macOS host with active BSDP servers on the network (identifying information removed):
```
DHCPv4
  opcode=BootRequest
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=0
  transactionid=0x32a98af5
  numseconds=0
  flags=Unicast (0x00)
  clientipaddr=192.168.0.42
  youripaddr=0.0.0.0
  serveripaddr=0.0.0.0
  gatewayipaddr=0.0.0.0
  clienthwaddr=<MAC>
  serverhostname=
  bootfilename=
  options=
    DHCP Message Type -> [8]
    Vendor Specific Information -> [1 1 1 2 2 1 1]
    Parameter Request List -> [43 60]
    Maximum DHCP Message Size -> [5 220]
    Class Identifier -> [65 65 80 76 66 83 68 80 67 47 105 51 56 54 47 77 97 99 109 105 110 105 55 44 49]
    End -> []

DHCPv4
  opcode=BootReply
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=0
  transactionid=0x32a98af5
  numseconds=0
  flags=Unicast (0x00)
  clientipaddr=192.168.0.42
  youripaddr=192.168.0.42
  serveripaddr=0.0.0.0
  gatewayipaddr=0.0.0.0
  clienthwaddr=<MAC>
  serverhostname=
  bootfilename=
  options=
    Subnet Mask -> [255 255 255 0]
    Router -> [192 168 0 1]
    Vendor Specific Information -> [1 1 1 4 2 128 128 7 4 129 0 16 0 9 18 129 0 16 0 13 102 97 99 101 98 111 111 116 45 98 115 100 112]
    DHCP Message Type -> [5]
    Server Identifier -> <SERVER ID>
    Class Identifier -> [65 65 80 76 66 83 68 80 67]
    End -> []

DHCPv4
  opcode=BootRequest
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=0
  transactionid=0x32a98af5
  numseconds=0
  flags=Unicast (0x00)
  clientipaddr=0.0.0.0
  youripaddr=0.0.0.0
  serveripaddr=0.0.0.0
  gatewayipaddr=0.0.0.0
  clienthwaddr=<MAC>
  serverhostname=
  bootfilename=
  options=
    Class Identifier -> [65 65 80 76 66 83 68 80 67 47 105 51 56 54 47 77 97 99 109 105 110 105 55 44 49]
    Parameter Request List -> [1 3 67 43 60]
    DHCP Message Type -> [8]
    Vendor Specific Information -> [1 1 2 2 2 1 1 8 4 129 0 16 0 3 4 10 102 120 23]
    End -> []

DHCPv4
  opcode=BootReply
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=2
  transactionid=0x32a98af5
  numseconds=0
  flags=Broadcast (0x8000)
  clientipaddr=192.168.0.42
  youripaddr=192.168.0.42
  serveripaddr=0.0.0.0
  gatewayipaddr=0.0.0.0
  clienthwaddr=<MAC>
  serverhostname=
  bootfilename=
  options=
    Subnet Mask -> [255 255 255 0]
    Router -> [192 168 0 1]
    Vendor Specific Information -> [1 1 2 8 4 129 0 16 0 130 10 78 101 116 98 111 111 116 48 48 49]
    DHCP Message Type -> [5]
    Server Identifier -> <SERVER ID>
    Class Identifier -> [65 65 80 76 66 83 68 80 67]
    End -> []
```


insomniacslk:
Thanks for your PR @get9 !

I quickly looked through the code and thought that it could be better to separate the BSDP code into its own separate subpackage. It it not part of DHCPv4 itself but it's a vendor extension, and there is an additional client implementation for the second quartet of packets. I think that a separate package would be easier to follow and it would set a clear distinction between the base protocol and the extensions. What do you think?

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+// BootImageID describes a boot image ID - whether it's an install image and
+// what kind of boot image (e.g. OS 9, macOS, hardware diagnostics)
+type BootImageID struct {
+	isInstall bool
+	imageKind byte
+	index     uint16
+}
+
+// toBytes serializes a BootImageID to network-order bytes.
+func (b BootImageID) toBytes() (bytes []byte) {
```
insomniacslk: any reason not to expose this method as public?

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	bytes[0] |= b.imageKind
+
+	// Index
+	binary.BigEndian.PutUint16(bytes[2:], b.index)
+	return
+}
+
+// BootImageIDFromBytes deserializes a collection of 4 bytes to a BootImageID.
+func bootImageIDFromBytes(bytes []byte) BootImageID {
```
insomniacslk: `golint` here will likely complain about the mismatch between function name and first word in the comment. By the way this seems a good candidate for a public function too, any reason to keep it private?

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+
+// BootImage describes a boot image - contains the boot image ID and the name.
+type BootImage struct {
+	ID BootImageID
+	// This is a utf-8 string.
+	Name string
+}
+
+// toBytes converts a BootImage to a slice of bytes.
+func (b *BootImage) toBytes() (bytes []byte) {
```
insomniacslk: same here about being public

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+	idBytes := b.ID.toBytes()
+	bytes = append(bytes, idBytes[:]...)
+	bytes = append(bytes, byte(len(b.Name)))
+	bytes = append(bytes, []byte(b.Name)...)
+	return
+}
+
+// BootImageFromBytes returns a deserialized BootImage struct from bytes as well
+// as the number of bytes read from the slice.
+func bootImageFromBytes(bytes []byte) (*BootImage, int, error) {
```
insomniacslk: I tend to prefer short lists of return values - what about adding a `Length` or `Size` method to `BootImage` and use that below instead of returning the size here?

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+// option and returns a list of BootImages.
+func parseBootImagesFromBSDPOption(data []byte) ([]BootImage, error) {
+	// Should at least have the # bytes of boot images.
+	if len(data) < 4 {
+		return nil, fmt.Errorf("invalid length boot image list")
+	}
+
+	readByteCount := 0
+	start := data
+	var bootImages []BootImage
```
insomniacslk: nit (feel free to ignore): the above could be more Go-ish, something like:

```Go
var (
    readByteCount int
    start = data
    bootImages []BootImages
)

File: dhcpv4/bsdp.go [hunk 6]
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	return bootImages, nil
+}
+
+// parseVendorOptionsFromOptions extracts the sub-options list of the vendor-
+// specific options from the larger DHCP options list.
+func parseVendorOptionsFromOptions(options []Option) []Option {
+	var vendorOpts []Option
+	var err error
```
insomniacslk: nit: same as above:

```Go
var (
    vendorOpts []Option
    err error
)
```

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+	for _, opt := range options {
+		if opt.Code == OptionVendorSpecificInformation {
+			vendorOpts, err = OptionsFromBytes(opt.Data)
+			if err != nil {
+				return []Option{}
```
insomniacslk: I prefer the error not to be skipped silently, can you `log` it?

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+	}
+
+	return bootImages, nil
+}
+
+// NewInformListForInterface creates a new INFORM packet for interface ifname
+// with configuration options specified by config.
+func NewInformListForInterface(iface string, replyPort uint16) (*DHCPv4, error) {
+	d, err := NewInformForInterface(iface /* needsBroadcast */, false)
```
insomniacslk: can you remove the `/* needsBroadcast */` comment if it's a leftover?

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+			}
+			bootImages = append(bootImages, images...)
+		}
+	}
+
+	return bootImages, nil
+}
+
+// NewInformListForInterface creates a new INFORM packet for interface ifname
+// with configuration options specified by config.
```
insomniacslk: can you add information about `replyPort` in the function comment?

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+		Option{
+			Code: BSDPOptionVersion,
+			Data: BSDPVersion1_1,
+		},
+	}
+
+	// If specified, replyPort MUST be a priviledged port.
+	if replyPort != 0 && replyPort != ClientPort {
+		if replyPort >= 1024 {
+			return nil, fmt.Errorf("replyPort must be a priviledged port (< 1024)")
```
insomniacslk: typo "privileged"

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+		},
+		Option{
+			Code: BSDPOptionSelectedBootImageID,
+			Data: append([]byte{4}, selectedImage.ID.toBytes()...),
+		},
+	}
+
+	// Find server IP address
+	var serverIP []byte
+	for _, opt := range ack.options {
```
insomniacslk: ouch, my bad, I haven't implemented `GetOption` and `GetOneOption` in the dhcpv4 package like in the dhcpv6 oen :(

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+		if opt.Code == OptionServerIdentifier {
+			serverIP = make([]byte, 4)
+			copy(serverIP, opt.Data)
```
insomniacslk: if I recall correctly this does not behave as expected, since `net.IP` is a slice (used for both IPv4 and IPv6) and the last four bytes are used for IPv4, not the first four. This will probably yield a `0.0.0.0` address, see https://play.golang.org/p/ofkyBXT3tTG.

You can use a simpler syntax, just `serverIP = net.IP(opt.Data)`

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+	}
+	if len(serverIP) == 0 {
```
insomniacslk: a stronger check is `if serverIP.To4() == nil`

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+			Data: BSDPVersion1_1,
+		},
+		Option{
+			Code: BSDPOptionSelectedBootImageID,
+			Data: append([]byte{4}, selectedImage.ID.toBytes()...),
+		},
+	}
+
+	// Find server IP address
+	var serverIP []byte
```
insomniacslk: you can change this to be a `net.IP`, and below (line 291) do `serverIP = net.IP(opt.Data)`

dhcpv4/bsdp.go
```diff
@@ -0,0 +1,328 @@
...
+		return nil, fmt.Errorf("could not parse server identifier from ACK")
+	}
+	vendorOpts = append(vendorOpts, Option{
+		Code: BSDPOptionServerIdentifier,
+		Data: serverIP,
+	})
+
+	// Validate replyPort if requested.
+	if replyPort != 0 && replyPort != ClientPort {
+		// replyPort MUST be a priviledged port.
```
insomniacslk: same typos here and below

dhcpv4/bsdp_client.go
```diff
@@ -0,0 +1,109 @@
...
+	err = syscall.SetsockoptInt(fd, syscall.SOL_SOCKET, syscall.SO_BROADCAST, 1)
+	if err != nil {
+		return conversation, err
+	}
+	err = BindToInterface(fd, ifname)
+	if err != nil {
+		return conversation, err
+	}
+
+	daddr := syscall.SockaddrInet4{Port: ClientPort, Addr: [4]byte{255, 255, 255, 255}}
```
insomniacslk: you can use `net.IPv4bcast` instead of the `[4]byte{..}`

dhcpv4/bsdp_client.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return conversation, err
+	}
+	err = syscall.Sendto(fd, packet, 0, &daddr)
+	if err != nil {
+		return conversation, err
+	}
+
+	// ACK 1
+	conn, err := net.ListenUDP("udp4", &net.UDPAddr{IP: net.IPv4(0, 0, 0, 0), Port: ClientPort})
```
insomniacslk: you can use `net.IPv4zero` instead of the `[4]byte{..}`

dhcpv4/bsdp_client.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return conversation, err
+	}
+	// TODO match the packet content
+	// TODO check that the peer address matches the declared server IP and port
+	conversation = append(conversation, *ack1)
+
+	// Parse boot images sent back by server
+	bootImages, err := ParseBootImageListFromAck(*ack1)
+	fmt.Println(bootImages)
```
insomniacslk: I guess this is a debugging leftover

dhcpv4/dhcpv4.go
```diff
@@ -37,6 +38,33 @@ type DHCPv4 struct {
 	options        []Option
 }
 
+// interfaceV4Addr obtains the currently-configured IPv4 address for iface.
```
insomniacslk: this function actually gets the *first* IPv4 address configured for the interface. I'd rather return a list of `net.IP` instead, and leave to the caller to pick the first one

dhcpv4/dhcpv4.go
```diff
@@ -500,6 +568,18 @@ func (d *DHCPv4) Summary() string {
 	)
 	ret += "  options=\n"
 	for _, opt := range d.options {
+		// Parse and display sub-options
+		if opt.Code == OptionVendorSpecificInformation {
```
insomniacslk: I understand that because of this block of code, moving the BSDP specific code above may be difficult. Instead of giving up though, I'd rather modify the signature of `RequestFromOffer` to be variadic and accept optional functions that will modify the DHCP options. Something like this, https://play.golang.org/p/DrYLFweJNjJ but for DHCP options

dhcpv4/options.go
```diff
@@ -64,20 +75,27 @@ func OptionsFromBytes(data []byte) ([]Option, error) {
...
 	}
 	return options, nil
 }
 
-func OptionsToBytes(options []Option) []byte {
-	// Convert a list of options to a wire-format representation. This will
-	// include the Magic Cookie
+// OptionsToBytesWithMagicCookie converts a list of options to a wire-format
+// representation with the DHCP magic cookie prepended.
+func OptionsToBytesWithMagicCookie(options []Option) []byte {
```
insomniacslk: I'd rather rename this function back to `OptionsToBytes`, since it's more general and going to be used more frequently, and the one below to become `OptionsToBytesWithoutMagicCookie`

dhcpv4/bsdp.go
get9: Not necessarily, I just didn't think it'd need to be public given that all the serialization is handled by the exchange functions. I can make it public though to make it a bit more flexible.

dhcpv4/bsdp.go
get9: Mmm, no real reason. I'm still learning what bits to make public and what to make private.

dhcpv4/bsdp.go
get9: I can, but there currently aren't any other usages of logging in the package. Should I use the global logger?

dhcpv4/bsdp.go
get9: I was going to leave it there to clarify what the `false` param is for (since go doesn't have kwargs :( )

get9:
As I kept adding more and more, it seemed like it should be a separate package since it has a lot of custom stuff in it. Should it have its own types then?

dhcpv4/dhcpv4.go
get9: I'm not sure I follow - do you mean modify the `Summary()` function instead?

dhcpv4/bsdp.go
insomniacslk: yeah, it'd be nice to make it publi public. The `BootImageID` object itself is public so anything that makes it easy to operate with it (e.g. serialization and deserialization) sounds like a good candidate for being public. The end goal for this library is to be reusable, so if someone wants to just use the BSDP subpackage but not all the rest, with public methods they can

dhcpv4/bsdp.go
insomniacslk: yes, the global logger is OK here

dhcpv4/bsdp.go
insomniacslk: ah, so I guess we can steal from C++ here and make something like

```Go
    d, err := NewInformForInterface(iface, /* needsBroadcast = */ false)
```

dhcpv4/bsdp.go
insomniacslk: can you help my memory and add a TODO here before the loop?

```Go
// TODO replace this loop with `ack.GetOneOption(BSDPOptionBootImageList)`
```
I will implement it after your PR is merged, to mimic the [DHCPv6 interface](https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/dhcpv6.go#L17)

dhcpv4/dhcpv4.go
insomniacslk: Sorry, I was tricked by the compressed view on GitHub :( let me re-read this

insomniacslk: Yeah, I thought I was reading `RequestFromOffer`, not `Summary`.

By the way, same principle - `BSDPString()` is too vendor-specific to live in the `dhcpv4 `package. In the `dhcpv6` package I made `Options` an interface, not a struct, which makes it a lot more flexible - you just call `myopt.String()` and it will print your option-specific code. Unfortunately the `dhcpv4` package is a bit behind.

In this case I'd rather remove the BSDPString method and the loop below that prints the suboption, and wait until, after your PR, we have `Option` as interface. I commit to do this as soon as this PR is merged so we can quickly re-add this feature.

To summarize: remove the BSDPString method and the loop that prints the suboption in `Summary`, and please add a TODO for me to implement `Option` as interface

dhcpv4/bsdp/types.go
```diff
@@ -0,0 +1,62 @@
...
+	MessageTypeSelect
+	MessageTypeFailed
+)
+
+// Different types of BootImages - e.g. for different flavors of macOS.
+const (
+	BootImageTypeMacOS9 byte = iota
+	BootImageTypeMacOSX
+	BootImageTypeMacOSXServer
+	BootImageTypeHardwareDiagnostics
```
insomniacslk: can you define a type `BootImageType`? I.e.

`type BootImageType byte`

and use this throughout the code? It will require more casting around, but it will also make the code safer against erroneous uses of the `ImageType` field that is simply `byte`

dhcpv4/bsdp/bsdp.go
```diff
@@ -0,0 +1,315 @@
...
+		bytes[0] |= 0x80
+	}
+	bytes[0] |= b.ImageType
+	binary.BigEndian.PutUint16(bytes[2:], b.Index)
+	return
+}
+
+// BootImageIDFromBytes deserializes a collection of 4 bytes to a BootImageID.
+func BootImageIDFromBytes(bytes []byte) BootImageID {
+	return BootImageID{
```
insomniacslk: if the slice `bytes` is less than one byte long this will panic. Can you enforce a minimum length requirement here?

dhcpv4/bsdp/bsdp.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// BootImage describes a boot image - contains the boot image ID and the name.
+type BootImage struct {
+	ID   BootImageID
+	Name string
+}
+
+// ToBytes converts a BootImage to a slice of bytes.
+func (b *BootImage) ToBytes() (bytes []byte) {
```
insomniacslk: nit: for consistency with the rest of the code can you return `bytes` as a non-named type?

dhcpv4/bsdp/bsdp.go
```diff
@@ -0,0 +1,315 @@
...
+// specific options from the larger DHCP options list.
+func ParseVendorOptionsFromOptions(options []dhcpv4.Option) []dhcpv4.Option {
+	var (
+		vendorOpts []dhcpv4.Option
+		err        error
+	)
+	for _, opt := range options {
+		if opt.Code == dhcpv4.OptionVendorSpecificInformation {
+			vendorOpts, err = dhcpv4.OptionsFromBytesWithoutMagicCookie(opt.Data)
+			if err != nil {
```
insomniacslk: this error should be logged

dhcpv4/bsdp/bsdp.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+	}
+
+	return bootImages, nil
+}
+
+// NewInformListForInterface creates a new INFORM packet for interface ifname
+// with configuration options specified by config.
+func NewInformListForInterface(iface string, replyPort uint16) (*dhcpv4.DHCPv4, error) {
+	d, err := dhcpv4.NewInformForInterface(iface /* needsBroadcast = */, false)
```
insomniacslk: the comma should be before the commented flag:

```
d, err := dhcpv4.NewInformForInterface(iface, /* needsBroadcast = */ false)
```

dhcpv4/bsdp/bsdp.go
```diff
@@ -0,0 +1,315 @@
...
+	d.AddOption(dhcpv4.Option{
+		Code: dhcpv4.OptionParameterRequestList,
+		Data: []byte{
+			dhcpv4.OptionVendorSpecificInformation,
+			dhcpv4.OptionClassIdentifier,
+		},
+	})
+
+	u16 := make([]byte, 2)
+	binary.BigEndian.PutUint16(u16, 1500)
```
insomniacslk: I'd make this 1500 as a global constant

dhcpv4/bsdp/bsdp.go
```diff
@@ -0,0 +1,315 @@
...
+	if serverIP.To4() == nil {
+		return nil, fmt.Errorf("could not parse server identifier from ACK")
+	}
+	vendorOpts = append(vendorOpts, dhcpv4.Option{
+		Code: OptionServerIdentifier,
+		Data: serverIP,
+	})
+
+	// Validate replyPort if requested.
+	if replyPort != 0 && replyPort != dhcpv4.ClientPort {
```
insomniacslk: this block is duplicated from `NewInformListForInterface`, can you make it a private function?

dhcpv4/bsdp/client.go
```diff
@@ -0,0 +1,113 @@
...
+	conversation := make([]dhcpv4.DHCPv4, 1)
+	var err error
+
+	// INFORM[LIST]
+	if d == nil {
+		d, err = NewInformListForInterface(ifname, dhcpv4.ClientPort)
+	}
+	conversation[0] = *d
+
+	fd, err := syscall.Socket(syscall.AF_INET, syscall.SOCK_RAW, syscall.IPPROTO_RAW)
```
insomniacslk: I guess I will move this code to a function in a future commit since it's basically duplicated from `dhcpv4.Client`, but not now. Could you add a TODO here for me please?

dhcpv4/dhcpv4.go
```diff
@@ -37,6 +38,34 @@ type DHCPv4 struct {
 	options        []Option
 }
 
+// InterfaceV4Addr obtains the currently-configured IPv4 address for iface.
```
insomniacslk: s/address/addresses/, and also non-loopback

dhcpv4/types.go
```diff
@@ -3,6 +3,18 @@ package dhcpv4
 // values from http://www.networksorcery.com/enp/protocol/dhcp.htm and
 // http://www.networksorcery.com/enp/protocol/bootp/options.htm
 
+// DHCP message types
+const (
+	MessageTypeDiscover byte = iota + 1
```
insomniacslk: please define a `MessageType` type here (same reasons as above)

dhcpv4/bsdp/types.go
get9: I did try that, but it was a lot more annoying due to all the casting. If you think that's better, can do.

dhcpv4/bsdp/bsdp.go
get9: Weird I thought I fixed that, must have gotten lost in rebasing.

dhcpv4/bsdp/types.go
get9: This would be a lot easier if Go supported enums...

insomniacslk: no that's fine, it's not worth the effort

dhcpv4/bsdp/bsdp.go
get9: Huh, goimports/gofmt seems to prefer the other way...

insomniacslk: What error do you get? I just tried with a simple program and golint didn't complain

```Go
package main

func blah(a string, b bool) {
}

func main() {
        blah("test", /* b = */ true)
}
```

get9: Not sure how much I can do on this one.

insomniacslk: just ignore it for now. Out of curiousity I'll have a look at it once merged

PR #7, opened 2018-03-07
Add Travis CI configs to run tests, gofmt, and govet

get9:
Inspired by what [stretchr/testify](https://github.com/stretchr/testify) does for their CI. This runs tests, runs gofmt, and runs govet over dhcpv4 and dhcpv6 on each change uploaded to github.


get9:
@insomniacslk: I think you'll also have to set up and enable this project in travis CI.

insomniacslk:
Thanks for the contribution :)

I am not convinced by CI running gofmt/govet for a couple of reasons:
1) failing tests on formatting is a strong opinion, and I believe a better place for gofmt is in the pre-commit hooks rather than in CI (an example is provided at https://golang.org/misc/git/pre-commit )
2) different versions of gofmt will yield to different results, especially across versions.

I also remember reading on the golang issue tracker strong advise against using gofmt in tests for similar reasons, but I can't find that again.

get9:
Understandable for gofmt, I agree that it would be better-implemented as a commit hook.

Do you think govet should still be part of CI? It’s less flaky than gofmt in terms of changes and it’s pretty high signal from what I know.

insomniacslk:
I am not entirely sure about govet either, since it's a tool that reports warnings, not errors. More philosophically, should a test fail for a warning? Without hard feelings, I think not.

Alternatively, can public CI systems report warnings without blocking? After a quick search it seems not, but I'm happy if we find one.

(however we definitely need the `go test` one)

get9:
So would you want to enforce contributors run gofmt/govet before merging their PR into trunk? If so, how do you ensure they do so? You can't expect that everyone will have githooks to run those right? I'm thinking these kinds of things would be more for warnings in the CI that says "hey, this doesn't pass this check" but still allow you to merge it in if you want.

insomniacslk:
as discussed offline we will look into github bots that will raise linter warnings as comments

insomniacslk:
Looks good :)  thanks!

(and as discussed let's tackle the linter in a separate issue)

PR #11, opened 2018-03-09
Added stickler config for golint bot

insomniacslk:
Using Stickler-ci.com for golint bot


insomniacslk:
cc @get9 - I added a bot to run `golint` and suggest fixes for the code

insomniacslk:
now we just have to try it with some bad code :)

PR #15, opened 2018-03-10
dhcpv4: fix client regression

insomniacslk:
The previous commit introduced a regression in the DHCPv4 client, due to the way `net.IP` works. The erroneous code was:

```
	BroadcastAddressBytes = [4]byte{
		net.IPv4bcast[0],
		net.IPv4bcast[1],
		net.IPv4bcast[2],
		net.IPv4bcast[3],
	}
```

The first four bytes are all zeros, because `net.IP` can be stored as a 16-bytes slice even for IPv4 addresses. The safest way is to always use `.To4()` to make sure to use the right bytes.

### DIFF
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -17,14 +17,6 @@ const (
 )
 
 var (
-	// BroadcastAddressBytes are the bytes representing net.IPv4bcast.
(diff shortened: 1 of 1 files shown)


get9:
Looks good to me!

PR #17, opened 2018-03-13
Add more specific dhcpv4 options

get9:
This PR adds some more specific options to dhcpv4. Currently, only the ones used by the main dhcpv4 package will be added.

### DIFF
--- a/dhcpv4/option_class_identifier.go
+++ b/dhcpv4/option_class_identifier.go
@@ -1,9 +1,11 @@
 package dhcpv4
 
 import (
-	"errors"
 	"fmt"
 )
+
+// This option implements the Class Identifier option
+// https://tools.ietf.org/html/rfc2132
 
 // OptClassIdentifier represents the DHCP message type option.
 type OptClassIdentifier struct {
@@ -15,7 +17,7 @@ type OptClassIdentifier struct {
 func ParseOptClassIdentifier(data []byte) (*OptClassIdentifier, error) {
 	// Should at least have code and length
 	if len(data) < 2 {
-		return nil, errors.New("too short of byte stream")
+		return nil, ErrShortByteStream
 	}
 	code := OptionCode(data[0])
 	if code != OptionClassIdentifier {
@@ -23,7 +25,7 @@ func ParseOptClassIdentifier(data []byte) (*OptClassIdentifier, error) {
 	}
 	length := int(data[1])
 	if len(data) < 2+length {
-		return nil, fmt.Errorf("short bytestream: expected %v bytes, got %v", 2+length, len(data))
+		return nil, ErrShortByteStream
 	}
 	return &OptClassIdentifier{Identifier: string(data[2 : 2+length])}, nil
 }
--- a/dhcpv4/option_class_identifier.go
+++ b/dhcpv4/option_class_identifier.go
@@ -1,0 +1,50 @@
+package dhcpv4
+
+import (
+	"errors"
+	"fmt"
+)
+
+// OptClassIdentifier represents the DHCP message type option.
+type OptClassIdentifier struct {
+	Identifier string
+}
+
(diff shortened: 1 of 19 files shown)


dhcpv4/option_class_identifier.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+)
+
+// This option implements the Class Identifier option
+// https://tools.ietf.org/html/rfc2132
+
+// OptClassIdentifier represents the DHCP message type option.
+type OptClassIdentifier struct {
+	Identifier string
```
insomniacslk: should this be `string` or `[]byte` ? Looking at the RFC I'd say `[]byte`, then all the string conversions should be done elsewhere

dhcpv4/option_class_identifier.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// ToBytes returns a serialized stream of bytes for this option.
+func (o *OptClassIdentifier) ToBytes() []byte {
+	return append([]byte{byte(o.Code()), byte(o.Length())}, []byte(o.Identifier)...)
+}
+
+// String returns a human-readable string for this option.
+func (o *OptClassIdentifier) String() string {
+	return fmt.Sprintf("Class Identifier -> %v", o.Identifier)
```
insomniacslk: e.g. if the above becomes `[]byte`, this should be converted to string first

dhcpv4/option_maximum_dhcp_message_size.go
```diff
@@ -0,0 +1,58 @@
...
+type OptMaximumDHCPMessageSize struct {
+	Size uint16
+}
+
+// ParseOptMaximumDHCPMessageSize constructs an OptMaximumDHCPMessageSize struct from a sequence of
+// bytes and returns it, or an error.
+func ParseOptMaximumDHCPMessageSize(data []byte) (*OptMaximumDHCPMessageSize, error) {
+	// Should at least have code, length, and message type.
+	if len(data) < 4 {
+		return nil, errors.New("too short of byte stream")
```
insomniacslk: use `ErrShortByteStream` here

dhcpv4/options.go
```diff
@@ -27,23 +35,26 @@ func ParseOption(data []byte) (Option, error) {
...
-	switch code {
+	switch OptionCode(data[0]) {
+	case OptionDHCPMessageType:
+		opt, err = ParseOptMessageType(data)
+	case OptionParameterRequestList:
+		opt, err = ParseOptParameterRequestList(data)
+	case OptionRequestedIPAddress:
+		opt, err = ParseOptRequestedIPAddress(data)
+	case OptionServerIdentifier:
+		opt, err = ParseOptServerIdentifier(data)
```
insomniacslk: here you didn't `switch` on all the options you implemented, e.g. `OptMaximumDHCPMessageSize` and others are missing

dhcpv4/option_class_identifier.go
get9: I was going to make the specific opts more API/human-friendly, then handle serialization in the ToBytes method.

insomniacslk:
This looks good to me. I'll wait for CI to finish the tests before squashing and merging

dhcpv4/option_class_identifier.go
insomniacslk: I think that the API should be as friendly as possible, as long as it doesn't create other kinds of risk. The string could be a multi-byte string when decoded, which may cause issues with `Length`, and the decoding of all the following options as a consequence

insomniacslk: as you showed me offline with https://play.golang.org/p/sXdfZg0987B, your code is good, and I learned something new, thanks!

insomniacslk:
@get9 is this ok to merge or do you plan to add more commits?

get9:
Good to merge!

PR #18, opened 2018-03-17
Added OptionBroadcastAddress

insomniacslk:


insomniacslk:
The next options will be `Router`, `Subnet Mask` and `Domain Name Server`. These are literally duplicates. Maybe we should implement a common class for options that contain a `net.IP`, a `string` and so on, and reuse that code?

```
2018/03/17 12:23:08 DHCPv4
  opcode=BootReply
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=0
  transactionid=0x7fbe08d5
  numseconds=0
  flags=Broadcast (0x8000)
  clientipaddr=0.0.0.0
  youripaddr=10.255.69.155
  serveripaddr=10.255.68.1
  gatewayipaddr=0.0.0.0
  clienthwaddr=9c:b6:d0:d4:04:d9
  serverhostname=
  bootfilename=
  options=
    DHCP Message Type -> OFFER
    Server Identifier -> 10.255.68.1
    IP Addresses Lease Time -> [0 0 1 44]
    Renew Time Value -> [0 0 0 150]
    Rebinding Time Value -> [0 0 1 6]
    Subnet Mask -> [255 255 252 0]
    Broadcast Address -> 10.255.71.255
    Router -> [10 255 68 1]
    Domain Name Server -> [10 255 68 1]
    Domain Name -> lan
    End -> []
```

get9:
That’s what I was thinking. Most options either contain a string, IP, or int of some specific size.

How would you structure this in Go?

insomniacslk:
An option is to use anonymous fields, something like https://play.golang.org/p/sUZ2yX9WQOb , but that would still require to implement some methods like `String` and `ToBytes`. I am not sure if this is worth though

insomniacslk:
This is probably the closest to what we need, https://play.golang.org/p/PR0GR3X-NXf .

We would still have to implement tests for each option

insomniacslk:
This is how it would look like, https://github.com/insomniacslk/dhcp/pull/19 . Worth doing?

PR #19, opened 2018-03-18
Added OptGenericIP to simplify creation of options that wrap an IP address

insomniacslk:
OptGenericIP is a fake option that simplifies the creation of other options that wrap an IP address.

### DIFF
--- a/dhcpv4/option_broadcast_address.go
+++ b/dhcpv4/option_broadcast_address.go
@@ -1,36 +1,23 @@
 package dhcpv4
 
-import (
-	"fmt"
-	"net"
-)
+import "fmt"
 
-// This option implements the server identifier option
+// This option implements the broadcast address option
 // https://tools.ietf.org/html/rfc2132
 
-// OptBroadcastAddress represents an option encapsulating the server identifier.
+// OptBroadcastAddress represents an option encapsulating the broadcast address.
 type OptBroadcastAddress struct {
-	BroadcastAddress net.IP
+	OptGenericIP
 }
 
 // ParseOptBroadcastAddress returns a new OptBroadcastAddress from a byte
 // stream, or error if any.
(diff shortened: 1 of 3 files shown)


insomniacslk:
@get9 this is how a generic IP option could look like (https://github.com/insomniacslk/dhcp/pull/18 for background). Do you think it's worth doing?

insomniacslk:
Another disadvantage of this approach is that the field would always be named `IP`, not `BroadcastAddress`

insomniacslk:
it would also require exposing this piece of internals to the caller (e.g. `RequestFromOffer` in `dhcp4.go`). Even less sure if I like this or not

get9:
What if instead we make OptGenericIP look like the following:
```
type OptGenericIP struct {
    name string
    code OptionCode
    IP net.IP
}
```

and then you create one like:
```
newOpt := OptGenericIP{
    name: "BroadcastAddress",
    code: OptionBroadcastAddress,
    IP: myCoolIP,
}
```

Then the `ToString()` looks something like:
```
func (o *OptGenericIP) ToString() string {
    return fmt.Sprintf("%s -> %v", o.name, o.IP)
}
```

This way we can fully parameterize the option, but still share the implementation since it's _exactly_ the same for all of these options.

Alternatively, we could just copypasta the implementations and switch out names, maybe use codegen or something?

insomniacslk:
in that case you would miss the type checking because there will not be a specialized type for the various IP-based options. I think we'd rather copy-pasta until all the options are implemented :(

nemith:
May be able to accomplish this with embedding instead.

insomniacslk:
that was the idea - https://github.com/insomniacslk/dhcp/pull/19/files#diff-530d8bda11b58b48549f7b6391d40041R10 . Is that what you mean?

insomniacslk:
closing this because the pros are less than the cons

PR #21, opened 2018-03-23
Add some more specific bsdp options + vendor specific implementation

get9:
Added more BSDP specific options + an implementation of the vendor specific options.

### DIFF
--- a/dhcpv4/bsdp/bsdp_option_boot_image_list.go
+++ b/dhcpv4/bsdp/bsdp_option_boot_image_list.go
@@ -1,0 +1,86 @@
+// +build darwin
+
+package bsdp
+
+import (
+	"fmt"
+
+	"github.com/insomniacslk/dhcp/dhcpv4"
+)
+
+// Implements the BSDP option listing the boot images.
+
+// OptBootImageList contains the list of boot images presented by a netboot
+// server.
+type OptBootImageList struct {
+	Images []BootImage
+}
+
+// ParseOptBootImageList constructs an OptBootImageList struct from a sequence
+// of bytes and returns it, or an error.
+func ParseOptBootImageList(data []byte) (*OptBootImageList, error) {
+	// Should have at least code + length
+	if len(data) < 2 {
+		return nil, dhcpv4.ErrShortByteStream
+	}
+	code := dhcpv4.OptionCode(data[0])
+	if code != OptionBootImageList {
+		return nil, fmt.Errorf("expected option %v, got %v instead", OptionBootImageList, code)
+	}
+	length := int(data[1])
+	if len(data) < length+2 {
+		return nil, fmt.Errorf("expected length %d, got %d instead", length, len(data))
+	}
+
+	// Offset from code + length byte
+	var bootImages []BootImage
+	idx := 2
+	for {
+		if idx >= len(data) {
+			break
+		}
+		image, err := BootImageFromBytes(data[idx:])
+		if err != nil {
+			return nil, fmt.Errorf("parsing bytes stream: %v", err)
+		}
+		bootImages = append(bootImages, *image)
+
+		// 4 bytes of BootImageID, 1 byte of name length, name
+		idx += 4 + 1 + len(image.Name)
+	}
+
+	return &OptBootImageList{bootImages}, nil
+}
+
+// Code returns the option code.
+func (o *OptBootImageList) Code() dhcpv4.OptionCode {
+	return OptionBootImageList
+}
+
+// ToBytes returns a serialized stream of bytes for this option.
+func (o *OptBootImageList) ToBytes() []byte {
+	bs := make([]byte, 0, 2+o.Length())
+	bs = append(bs, []byte{byte(o.Code()), byte(o.Length())}...)
+	for _, image := range o.Images {
+		bs = append(bs, image.ToBytes()...)
+	}
+	return bs
+}
+
+// String returns a human-readable string for this option.
+func (o *OptBootImageList) String() string {
+	s := "BSDP Boot Image List ->"
+	for _, image := range o.Images {
+		s += "\n  " + image.String()
+	}
+	return s
+}
+
+// Length returns the length of the data portion of this option.
+func (o *OptBootImageList) Length() int {
+	length := 0
+	for _, image := range o.Images {
+		length += 4 + 1 + len(image.Name)
+	}
+	return length
+}
--- a/dhcpv4/bsdp/bsdp_option_default_boot_image_id.go
+++ b/dhcpv4/bsdp/bsdp_option_default_boot_image_id.go
@@ -1,0 +1,59 @@
+// +build darwin
+
+package bsdp
+
+import (
+	"fmt"
+
+	"github.com/insomniacslk/dhcp/dhcpv4"
+)
+
+// Implements the BSDP option default boot image ID, which tells the client
+// which image is the default boot image if one is not selected.
+
+// OptDefaultBootImageID contains the selected boot image ID.
+type OptDefaultBootImageID struct {
+	ID BootImageID
+}
+
+// ParseOptDefaultBootImageID constructs an OptDefaultBootImageID struct from a sequence of
+// bytes and returns it, or an error.
+func ParseOptDefaultBootImageID(data []byte) (*OptDefaultBootImageID, error) {
+	if len(data) < 6 {
+		return nil, dhcpv4.ErrShortByteStream
+	}
+	code := dhcpv4.OptionCode(data[0])
+	if code != OptionDefaultBootImageID {
+		return nil, fmt.Errorf("expected option %v, got %v instead", OptionDefaultBootImageID, code)
+	}
+	length := int(data[1])
+	if length != 4 {
+		return nil, fmt.Errorf("expected length 4, got %d instead", length)
+	}
+	id, err := BootImageIDFromBytes(data[2:6])
+	if err != nil {
+		return nil, err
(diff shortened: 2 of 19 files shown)


get9:
One curious thing - because of the way the options are parsed, the vendor-specific options that come back in the ACKs from the server aren't parsed as `bsdp.OptVendorSpecificInformation`. Instead, they're parsed as normal `dhcpv4.OptionGeneric`. I can't change the implementation in dhcpv4 because that would add a circular dep to dhcpv4 (depending on bsdp). However, how can I modify the ACK so that it parses this option correctly?

get9:
Sample test, now with pretty-printing:
```
DHCPv4
  opcode=BootRequest
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=0
  transactionid=0xea746944
  numseconds=0
  flags=Unicast (0x00)
  clientipaddr=192.168.0.53
  youripaddr=0.0.0.0
  serveripaddr=0.0.0.0
  gatewayipaddr=0.0.0.0
  clienthwaddr=<MAC>
  serverhostname=
  bootfilename=
  options=
    DHCP Message Type -> INFORM
    Vendor Specific Information ->
      BSDP Message Type -> LIST
      BSDP Version -> 1.1
    Parameter Request List -> [Vendor Specific Information, Class Identifier]
    Maximum DHCP Message Size -> 1500
    Class Identifier -> AAPLBSDPC/i386/Macmini7,1
    End -> []

DHCPv4
  opcode=BootReply
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=2
  transactionid=0xea746944
  numseconds=0
  flags=Unicast (0x00)
  clientipaddr=192.168.0.53
  youripaddr=192.168.0.53
  serveripaddr=0.0.0.0
  gatewayipaddr=0.0.0.0
  clienthwaddr=<MAC>
  serverhostname=
  bootfilename=
  options=
    Subnet Mask -> [255 255 255 0]
    Router -> [192 168 0 1]
    Vendor Specific Information ->
      BSDP Message Type -> LIST
      BSDP Server Priority -> 32896
      BSDP Default Boot Image ID -> [4096] installable macOS image
      BSDP Boot Image List ->
        image1 [4096] installable macOS image
    DHCP Message Type -> OFFER
    Server Identifier -> <IP>
    Class Identifier -> AAPLBSDPC
    End -> []

DHCPv4
  opcode=BootRequest
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=0
  transactionid=0xea746944
  numseconds=0
  flags=Unicast (0x00)
  clientipaddr=0.0.0.0
  youripaddr=0.0.0.0
  serveripaddr=0.0.0.0
  gatewayipaddr=0.0.0.0
  clienthwaddr=<MAC>
  serverhostname=
  bootfilename=
  options=
    Class Identifier -> AAPLBSDPC/i386/Macmini7,1
    Parameter Request List -> [Subnet Mask, Router, Bootfile Name, Vendor Specific Information, Class Identifier]
    DHCP Message Type -> INFORM
    Vendor Specific Information ->
      BSDP Message Type -> SELECT
      BSDP Version -> 1.1
      BSDP Selected Boot Image ID -> [4096] installable macOS image
      BSDP Server Identifier -> <IP>
    End -> []

DHCPv4
  opcode=BootReply
  hwtype=Ethernet
  hwaddrlen=6
  hopcount=2
  transactionid=0xea746944
  numseconds=0
  flags=Broadcast (0x8000)
  clientipaddr=192.168.0.53
  youripaddr=192.168.0.53
  serveripaddr=0.0.0.0
  gatewayipaddr=0.0.0.0
  clienthwaddr=<MAC>
  serverhostname=
  bootfilename=
  options=
    Subnet Mask -> [255 255 255 0]
    Router -> [192 168 0 1]
    Vendor Specific Information ->
      BSDP Message Type -> SELECT
      BSDP Selected Boot Image ID -> [4096] installable macOS image
      BSDP Machine Name -> Netboot001
    DHCP Message Type -> ACK
    Server Identifier -> <IP>
    Class Identifier -> AAPLBSDPC
    End -> []
```

Not sure why the second packet is an OFFER...

dhcpv4/bsdp/bsdp_option_boot_image_list.go
```diff
@@ -0,0 +1,86 @@
...
+		bs = append(bs, image.ToBytes()...)
+	}
+	return bs
+}
+
+// String returns a human-readable string for this option.
+func (o *OptBootImageList) String() string {
+	s := "BSDP Boot Image List ->"
+	for _, image := range o.Images {
+		s += "\n  " + image.String()
```
insomniacslk: I would normally expect `String()` to return a single-line string, and for multiline strings I normally implement `Summary()`, which is my own convention, not a Go convention. Do you think it would be too much clutter to have all the file names on a single line?

dhcpv4/bsdp/bsdp_option_default_boot_image_id.go
```diff
@@ -0,0 +1,59 @@
...
+	ID BootImageID
+}
+
+// ParseOptDefaultBootImageID constructs an OptDefaultBootImageID struct from a sequence of
+// bytes and returns it, or an error.
+func ParseOptDefaultBootImageID(data []byte) (*OptDefaultBootImageID, error) {
+	if len(data) < 6 {
+		return nil, dhcpv4.ErrShortByteStream
+	}
+	code := dhcpv4.OptionCode(data[0])
```
insomniacslk: I only now realized that we are using a different contract for DHCPv4 options, compared to DHCPv6 options. In the `dhcpv6` package, the `Parse*` methods don't require option code and length, as these are checked by the caller. While this may sound unintuitive, it removes the need to check the option twice.

I'd like to have a unique interface in the future - let's go ahead with what we have now, but then let's uniform the two

insomniacslk: I've opened https://github.com/insomniacslk/dhcp/issues/22

dhcpv4/bsdp/bsdp_option_server_priority.go
```diff
@@ -0,0 +1,58 @@
...
+	if len(data) < 4 {
+		return nil, dhcpv4.ErrShortByteStream
+	}
+	code := dhcpv4.OptionCode(data[0])
+	if code != OptionServerPriority {
+		return nil, fmt.Errorf("expected code %v, got %v", OptionServerPriority, code)
+	}
+	length := int(data[1])
+	if length != 2 {
+		return nil, fmt.Errorf("unexepcted length: expected 2, got %v", length)
```
insomniacslk: typo, "unexepcted" -> "unexpected"

dhcpv4/bsdp/client.go
```diff
@@ -22,6 +22,20 @@ func NewClient() *Client {
...
 }
 
+func castVendorOpt(ack *dhcpv4.DHCPv4) {
+	opts := ack.Options()
+	for i := 0; i < len(opts); i++ {
+		if opts[i].Code() == dhcpv4.OptionVendorSpecificInformation {
+			vendorOpt, err := ParseOptVendorSpecificInformation(opts[i].ToBytes())
+			// Oh well, we tried
+			if err != nil {
+				return
```
insomniacslk: returning here means that if an option decoding failed, all the subsequent options won't be tried. What about logging the error and continuing to the next?

dhcpv4/bsdp/option_vendor_specific_information.go
```diff
@@ -100,9 +111,14 @@ func (o *OptVendorSpecificInformation) ToBytes() []byte {
 
 // String returns a human-readable string for this option.
 func (o *OptVendorSpecificInformation) String() string {
-	s := "Vendor Specific Information ->\n"
+	s := "Vendor Specific Information ->"
 	for _, opt := range o.Options {
-		s += "  " + opt.String() + "\n"
+		optString := opt.String()
+		// If this option has sub-structures, offset them accordingly.
```
insomniacslk: if you decide to go to single line above, do it here too

insomniacslk:
on the circular import - I think your use of `castVendorOpt` is a good compromise

dhcpv4/bsdp/client.go
get9: There will nominally be only one Vendor specific option. This is a best-effort kind of thing that's only used for pretty-printing. If it fails, oh well.

dhcpv4/bsdp/bsdp_option_boot_image_list.go
insomniacslk: In my view, the purpose of `String` is to give a hint of what an object is, not describing it fully. What about truncating the output when it's too long, and printing the number of items? I used this trick in other libraries too. E.g. if the image length is longer than N characters, truncate and add "...", and if there are more than K entries don't print them but just add "..." to the list representation. Something like this:

MyObject -> ["Image With Long Na...", "Another Too Lon...", ...<5 entries total>]

dhcpv4/bsdp/client.go
insomniacslk: ok, sounds good

dhcpv4/bsdp/bsdp_option_boot_image_list.go
get9: Eh, I'd still prefer to actually print the full option. Particularly in one like this when you're listing all the images the server is providing.

insomniacslk: ok, let's go ahead

PR #24, opened 2018-03-26
Moved examples from external repo to examples dir

insomniacslk:
I originally used a separate repository for the examples, `github.com/insomniacslk/exdhcp`. This doesn't make much sense so I imported it here


insomniacslk:
I'll abandon this PR, it would import too much stuff in our internal repos and it's not worth it

PR #27, opened 2018-03-30
Remove build tags

get9:
Removes build tags from BSDP and breaks vendor class identifier into
OS-specific implementations so it is easier to integrate bsdp with other
libs.

### DIFF
--- a/dhcpv4/bsdp/vendor_class_identifier.go
+++ b/dhcpv4/bsdp/vendor_class_identifier.go
@@ -1,0 +1,9 @@
+// +build !darwin
+
+package bsdp
+
+// MakeVendorClassIdentifier calls the sysctl syscall on macOS to get the
+// platform model.
+func MakeVendorClassIdentifier() (string, error) {
+	return DefaultMacOSVendorClassIdentifier, nil
+}
--- a/dhcpv4/bsdp/vendor_class_identifier.go
+++ b/dhcpv4/bsdp/vendor_class_identifier.go
@@ -2,8 +2,8 @@
 
 package bsdp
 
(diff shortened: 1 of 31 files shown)


dhcpv4/bsdp/vendor_class_identifier_darwin.go
```diff
@@ -0,0 +1,17 @@
+package bsdp
```
insomniacslk: shouldn't this have the darwin build tag?

dhcpv4/bsdp/vendor_class_identifier.go
```diff
@@ -0,0 +1,9 @@
+// +build !darwin
+
+package bsdp
+
+// MakeVendorClassIdentifier calls the sysctl syscall on macOS to get the
```
insomniacslk: the comment should be updated, this doesn't use sysctl and is not called on darwin

dhcpv4/bsdp/vendor_class_identifier_darwin.go
get9: Nope because it's got the _darwin tag at the end. As long as the other (normal) file has the `// +build !darwin`, it's fine.

insomniacslk: relying on the other file having the `!darwin` tag seems at risk of inconsistence for future changes. This file would only work on darwin anyway - I think it's reasonable to add the `darwin` build tag anyway (unless there's something wrong in doing that and I'm missing the point)

insomniacslk: ok, I didn't know that `_darwin` tags the file for the build :D let's merge

PR #28, opened 2018-04-04
fix DHCPv6Relay ToBytes

pmazzini:


insomniacslk:
Thanks for the fix! This patch also highlights that the test of `ToBytes` is too naive, because the link and the peer addresses are the same, so the test wasn't failing. Can you also update `TestDHCPv6RelayToBytes` in  `dhcpv6relay_test.go` to use two different values for link and peer addrs?

pmazzini:
Thanks for the quick review. Test updated with different addresses.

dhcpv6/dhcpv6relay_test.go
```diff
@@ -84,7 +84,7 @@ func TestDHCPv6RelayToBytes(t *testing.T) {
 	r := DHCPv6Relay{
 		messageType: RELAY_FORW,
 		hopCount:    1,
-		linkAddr:    net.IPv6loopback,
+		linkAddr:    net.IPv6zero,
```
insomniacslk: I'd prefer something less trivial than all-zeros and all-zeros-except-one, can you scramble up some digit? :)

insomniacslk:
waiting for tests to complete

PR #29, opened 2018-04-06
fix OptInterfaceId

pmazzini:


insomniacslk:
Thanks for the fix! It looks good. Just one thing - there are more commits than there should be (they belong to another PR of yours). Can you remove those commits so I can merge?

PR #31, opened 2018-04-08
fix OptIAPrefix

pmazzini:


insomniacslk:
this option (the code that I wrote, not your fix) is a clear example of the horrors that copy-paste can do..

PR #33, opened 2018-04-10
add GetInnerPeerAddr

pmazzini:


dhcpv6/dhcpv6relay.go
```diff
@@ -143,3 +143,27 @@ func (d *DHCPv6Relay) GetInnerMessage() (DHCPv6, error) {
...
+// header, this is typically the IP address of the client making the request.
+func (r *DHCPv6Relay) GetInnerPeerAddr() (net.IP, error) {
+	var (
+		p   DHCPv6
+		err error
+	)
+	p = r
+	hops := r.HopCount()
+	addr := r.PeerAddr()
+	for i := 0; i < int(hops); i++ {
```
insomniacslk: rather than `int(hops)` let's iterate on the original type and declare `i` as uint8 (so either you pre-declare it a line before, or do `i := uint8(0)` in the loop, your choice)

insomniacslk: It's not just a nitpick, it's a safer way of running a loop and being sure that it can never run for longer than it should, on constrained data types

PR #37, opened 2018-04-18
Fixed DHCPv4 listener

insomniacslk:
There were two bugs in the DHCPv4 client:
* the listener was called *after* the sender
* the listener was not binding to the same interface as the sender, but listening for every UDP packet

This is now fixed.

### DIFF
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"net"
+	"os"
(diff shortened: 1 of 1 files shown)


insomniacslk:
@pmazzini @get9 mind to have a look at this?

insomniacslk:
ah yeah, bsdp (so darwin), that's why they didn't fail on my machine

pmazzini:
it looks good to me

Issue #34, opened 2018-04-11
Update constant names to be CamelCase


insomniacslk:
The Go linter prefers CamelCase rather than UPPERCASE_AND_UNDERSCORE names. A lot of variables in the code follow the latter. Let's make golint happier

PR #38, opened 2018-04-18
DHCPv6 client now waits for specific packet types

insomniacslk:
Before this patch the client just took whatever DHCP message it
received, without checking if it's what it wants. Now it checks for the
message type.
If the message is not a relay, also check the transaction ID

### DIFF
--- a/dhcpv6/types.go
+++ b/dhcpv6/types.go
@@ -5,6 +5,8 @@ package dhcpv6
 type MessageType uint8
 
 const (
+	// MSGTYPE_NONE is used internally and is not part of the RFC
+	MSGTYPE_NONE        MessageType = 0
 	SOLICIT             MessageType = 1
 	ADVERTISE           MessageType = 2
 	REQUEST             MessageType = 3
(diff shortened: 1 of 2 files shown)


insomniacslk:
actually, this is wrong. I have to use the input packet and get the xid and the type from the input packet

dhcpv6/types.go
```diff
@@ -5,6 +5,8 @@ package dhcpv6
 type MessageType uint8
 
 const (
+	// MSGTYPE_NONE is used internally and is not part of the RFC
+	MSGTYPE_NONE        MessageType = 0
```
get9: Should these be something like `MsgTypeNone` per go const style? Probably doesn't belong in this PR but a follow-up.

insomniacslk:
This needs to be implemented in dhcpv4 too (in `BroadcastSendReceive`)

dhcpv6/types.go
insomniacslk: yes, it's tracked in https://github.com/insomniacslk/dhcp/issues/34

PR #41, opened 2018-04-19
dhcpv6: added modifiers

insomniacslk:
Added support for packet modifiers, i.e. functions that can arbitrarily
manipulate a DHCPv6 packet. These modifiers are used by NewMessage,
NewSolicitForInterface, NewRequestForAdvertise, and can be used by other
packet creation functions.
A bunch of sample modifiers have been added under modifiers.go , too.
With the introduction of modifiers I also removed some options that
should not necessarily be in a standard DHCPv6 message.

In order to implement modifiers in a consistent manner, this patch also
modifies the return value of the `New*` functions, from `DHCPv6Message`
to `DHCPv6`.

### DIFF
--- a/dhcpv6/client.go
+++ b/dhcpv6/client.go
@@ -37,8 +37,10 @@ func NewClient() *Client {
 }
 
 // Exchange executes a 4-way DHCPv6 request (SOLICIT, ADVERTISE, REQUEST,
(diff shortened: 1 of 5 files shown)


insomniacslk:
cc @pmazzini

pmazzini:
lgtm

PR #44, opened 2018-04-20
dhcpv6: option userclass: supporting multiple user classes

insomniacslk:
Currently OptUserClass only allows for one user class. This patch adds the ability to specify multiple user classes

### DIFF
--- a/dhcpv6/option_userclass.go
+++ b/dhcpv6/option_userclass.go
@@ -5,12 +5,14 @@ package dhcpv6
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
+	"strings"
 )
 
 // OptUserClass represent a DHCPv6 User Class option
 type OptUserClass struct {
-	UserClass []byte
+	UserClasses [][]byte
 }
 
 // Code returns the option code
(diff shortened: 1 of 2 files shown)


dhcpv6/option_userclass.go
```diff
@@ -23,29 +25,46 @@ func (op *OptUserClass) ToBytes() []byte {
...
+			break
+		}
+		if len(data) < 2 {
+			return nil, errors.New("ParseOptUserClass: short data: missing length field")
+		}
+		ucLen := int(binary.BigEndian.Uint16(data[2:]))
+        if len(data) < ucLen +2 {
+			return nil, fmt.Errorf("ParseOptUserClass: short data: less than %d bytes", ucLen +2)
+        }
+		opt.UserClasses = append(opt.UserClasses, data[2:ucLen])
```
pmazzini: Isn't it missing something like `data = data[2+ucLen:]` to avoid an endless loop?

dhcpv6/option_userclass.go
```diff
@@ -23,29 +25,46 @@ func (op *OptUserClass) ToBytes() []byte {
 	buf := make([]byte, 6)
 	binary.BigEndian.PutUint16(buf[0:2], uint16(OPTION_USER_CLASS))
 	binary.BigEndian.PutUint16(buf[2:4], uint16(op.Length()))
-	// user-class-data has an internal data length field too..
-	binary.BigEndian.PutUint16(buf[4:6], uint16(len(op.UserClass)))
-	buf = append(buf, op.UserClass...)
+	for _, uc := range op.UserClasses {
+		binary.BigEndian.PutUint16(buf[4:6], uint16(len(uc)))
```
pmazzini: I think this will always write to buf[4:6] which is not the desider behaviour.

dhcpv6/option_userclass.go
insomniacslk: ...yes, it is. I had it in mind while writing the code and forgot it :( good catch!

dhcpv6/option_userclass.go
insomniacslk: this is a copy-pasta error. Thanks for spotting :)

dhcpv6/option_userclass.go
```diff
@@ -23,29 +25,46 @@ func (op *OptUserClass) ToBytes() []byte {
 	buf := make([]byte, 6)
```
pmazzini: I think this should now be `make([]byte, 4)`

pmazzini:
looks good

PR #58, opened 2018-05-05
Added tests for OptClientId and using net.HardwareAddr for DUID

insomniacslk:
Two things are changed in this diff:
* a breaking change: Duid.LinkLayerAddr is not a `[]byte` anymore but a more appropriate `net.HardwareAddr` (defined as a `[]byte`)
* tests for OptClientId

### DIFF
--- a/dhcpv6/duid.go
+++ b/dhcpv6/duid.go
@@ -3,6 +3,7 @@ package dhcpv6
 import (
 	"encoding/binary"
 	"fmt"
+	"net"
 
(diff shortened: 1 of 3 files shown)


insomniacslk:
cc @pmazzini

insomniacslk:
note: I changed the type for Duid.LinkLayerAddr when writing the tests for OptClientId - I realized net.HardwareAddr is more appropriate. I can split it into two separate PRs if necessary.

PR #61, opened 2018-05-08
OptStatusCode: public fields, proper codes, and tests

insomniacslk:
Three changes:
* status code and status message are now public - no more accessors, like in the other options
* defined status codes in the `iana` package as per rfc3315
* added unit tests


insomniacslk:
I have to add more coverage to this before merging

PR #62, opened 2018-05-10
OptDNSRecursiveNameServer gets public fields and tests, removes setter and getter

insomniacslk:
* Made public the NameServers field like in the other options.
* Removed setter and getter
* Added tests


pmazzini:
lgtm

insomniacslk:
failed for network timeout :(

PR #67, opened 2018-05-25
OptUserClass does not allow empty user classes

insomniacslk:
As noted by @pmazzini , there must be at least one user class defined if this option is used. The RFC says:

```
The data area of the user class option MUST contain one or more
instances of user class data.  Each instance of the user class data
is formatted as follows:
```

This patch enforces and tests this behaviour.

### DIFF
--- a/dhcpv6/option_userclass.go
+++ b/dhcpv6/option_userclass.go
@@ -69,5 +69,8 @@ func ParseOptUserClass(data []byte) (*OptUserClass, error) {
 		opt.UserClasses = append(opt.UserClasses, data[2:ucLen+2])
 		data = data[2+ucLen:]
 	}
+	if len(opt.UserClasses) < 1 {
+		return nil, errors.New("ParseOptUserClass: at least one user class is required")
+	}
 	return &opt, nil
 }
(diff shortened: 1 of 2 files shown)


insomniacslk:
Uh? The tests break on go 1.8 and 1.9 only, for something that seems broken in the stl:

```
# golang.org/x/sys/unix
../../../golang.org/x/sys/unix/ztypes_linux_amd64.go:1405: syntax error: unexpected = in type declaration
../../../golang.org/x/sys/unix/ztypes_linux_amd64.go:1407: syntax error: non-declaration statement outside function body
github.com/insomniacslk/dhcp/dhcpv4/bsdp
```

insomniacslk:
The code that breaks in the golang library is

```
type Nfgenmsg = struct 
      Nfgen_family uint8 
      Version      uint8 
      Res_id       uint16
} 
```

and the linter reports `ztypes_linux_amd64.go|1405 col 15 warning| expected type, found '='`

insomniacslk:
This was broken in `golang.org/x/sys/unix` commit ID 31355384c89b50e6faeffdb36f64a77a8210188e from yesterday :(

insomniacslk:
@pmazzini I think we can merge this anyway - the tests pass for 1.10 and tip, and they break for that syntax error on 1.8 and 1.9 on something that has to be fixed upstream.

insomniacslk:
well, this has been fixed 50 minutes ago :) in commit d4cedef3467ee719ea35470faeefbc6d0fa18e4b

insomniacslk:
there we go :) builds are succeeding again :)

PR #69, opened 2018-05-25
add NewRelayReplFromRelayForw

pmazzini:


dhcpv6/dhcpv6relay.go
```diff
@@ -180,3 +181,53 @@ func (r *DHCPv6Relay) GetInnerPeerAddr() (net.IP, error) {
 	}
 	return addr, nil
 }
+
+// NewRelayReplFromRelayForw creates a RELAY_REPLY packet based on a RELAY_FORW
```
insomniacslk: should be RELAY_REPL (because it was cheaper than RELAY_REPLY in the RFC :-(   )

dhcpv6/dhcpv6relay.go
```diff
@@ -180,3 +181,53 @@ func (r *DHCPv6Relay) GetInnerPeerAddr() (net.IP, error) {
...
+// packet containing the passed DHCPv6 message as payload.
```
insomniacslk: just to be sure that I understand: this will create a suitable relay reply based on the relay-forward, and will replace the inner packet with what's passed as second argument?

If so, can you be a bit more explicit in this comment on the inner packet being replaced?

dhcpv6/dhcpv6relay.go
```diff
@@ -180,3 +181,53 @@ func (r *DHCPv6Relay) GetInnerPeerAddr() (net.IP, error) {
...
+		err                error
+		linkAddr, peerAddr []net.IP
+		optiids            []Option
+	)
+
+	if relayForw == nil {
+		return nil, errors.New("RELAY_FORW cannot be nil")
+	}
+	if relayForw.Type() != RELAY_FORW {
+		return nil, errors.New("The passed RELAY_FORW must have RELAY_FORW type set")
```
insomniacslk: what about "The passed packet is not of type RELAY_FORW" ?

dhcpv6/dhcpv6relay.go
```diff
@@ -180,3 +181,53 @@ func (r *DHCPv6Relay) GetInnerPeerAddr() (net.IP, error) {
...
+	}
+	if msg == nil {
+		return nil, errors.New("The passed message cannot be nil")
+	}
+	if msg.IsRelay() {
+		return nil, errors.New("The passed message cannot be a relay")
+	}
+
+	relay := relayForw.(*DHCPv6Relay)
```
insomniacslk: I would move this type casting at line 197 together with the check on RELAY_FORW, and check that it's not nil (I know, I'm paranoid)

dhcpv6/dhcpv6relay.go
pmazzini: That is exactly the case. I will update the comment.

dhcpv6/dhcpv6relay.go
```diff
@@ -182,29 +182,27 @@ func (r *DHCPv6Relay) GetInnerPeerAddr() (net.IP, error) {
...
 	)
-
 	if relayForw == nil {
 		return nil, errors.New("RELAY_FORW cannot be nil")
 	}
 	if relayForw.Type() != RELAY_FORW {
-		return nil, errors.New("The passed RELAY_FORW must have RELAY_FORW type set")
+		return nil, errors.New("The passed packet is not of type RELAY_FORW")
 	}
+	relay := relayForw.(*DHCPv6Relay)
```
insomniacslk: what I actually meant is to do

```
relay, ok := relayForw.(*DHCPv6Relay)
if !ok {
    return nil, errors.New("Not a DHCPv6Relay")
}
if relay.Type() != RELAY_FORW {
    return nil, errors.New("The passed packet is not of type RELAY_FORW")
}

pmazzini: ok, thanks for the clarification, I updated it

dhcpv6/dhcpv6relay.go
pmazzini: right! :)

PR #70, opened 2018-06-07
add IsNetboot and IsUsingUEFI

pmazzini:


dhcpv6/utils.go
```diff
@@ -0,0 +1,59 @@
...
+	//               3    DEC Alpha
+	//               4    Arc x86
+	//               5    Intel Lean Client
+	//               6    EFI IA32
+	//               7    EFI BC
+	//               8    EFI Xscale
+	//               9    EFI x86-64
+	if opt := msg.GetOneOption(OPTION_CLIENT_ARCH_TYPE); opt != nil {
+		optat := opt.(*OptClientArchType)
+		if optat.ArchType == EFI_BC || optat.ArchType == EFI_X86_64 {
```
insomniacslk: can you add a comment that explains the choice of those two arch types? (and if necessary a TODO to investigate if other types are appropriate)

pmazzini: I will add a TODO.

PR #71, opened 2018-06-08
Pass through unknown DUIDs

stapelberg:
I’m building on a program which works with user-supplied DUIDs. Before this
change, the code would panic when users provide a DUID which is not (yet)
implemented.

In addition to this being a better failure mode, the behavior is also mandated
by the DHCPv6 RFC: https://tools.ietf.org/html/rfc3315#section-9 states:

    Clients and servers MUST treat DUIDs as opaque values and MUST only
    compare DUIDs for equality.

### DIFF
--- a/dhcpv6/duid_test.go
+++ b/dhcpv6/duid_test.go
@@ -104,3 +104,17 @@ func TestDuidUuidToBytes(t *testing.T) {
 		t.Fatalf("Invalid ToBytes result. Expected %v, got %v", expected, toBytes)
 	}
 }
+
+func TestOpaqueDuid(t *testing.T) {
+	duid := []byte("\x00\x0a\x00\x03\x00\x01\x4c\x5e\x0c\x43\xbf\x39")
+	d, err := DuidFromBytes(duid)
+	if err != nil {
+		t.Fatalf("DuidFromBytes: unexpected error: %v", err)
+	}
+	if got, want := d.Length(), len(duid); got != want {
+		t.Errorf("Length: unexpected result: got %d, want %d", got, want)
+	}
+	if got, want := d.ToBytes(), duid; !bytes.Equal(got, want) {
+		t.Fatalf("ToBytes: unexpected result: got %x, want %x", got, want)
+	}
+}
--- a/dhcpv6/duid.go
+++ b/dhcpv6/duid.go
@@ -33,6 +33,7 @@ type Duid struct {
 	EnterpriseNumber     uint32 // for DUID-EN. Ignored otherwise
 	EnterpriseIdentifier []byte // for DUID-EN. Ignored otherwise
 	Uuid                 []byte // for DUID-UUID. Ignored otherwise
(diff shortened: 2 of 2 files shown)


insomniacslk:
Hi @stapelberg , thanks for your contribution! I have limited connectivity for the next 3 days and can't make a proper review right now. @pmazzini if you have time, could you please have a look at this? Otherwise I will do on Monday.

One thing to do for sure is to add unit tests, as we want as much coverage as possible. Thanks!

pmazzini:
@stapelberg Thanks for opening a PR. It looks good to me. As @insomniacslk mentioned, would you be able to add unit tests? Thanks.

stapelberg:
Done

dhcpv6/duid_test.go
```diff
@@ -104,3 +104,17 @@ func TestDuidUuidToBytes(t *testing.T) {
 		t.Fatalf("Invalid ToBytes result. Expected %v, got %v", expected, toBytes)
 	}
 }
+
+func TestOpaqueDuid(t *testing.T) {
+	duid := []byte("\x00\x0a\x00\x03\x00\x01\x4c\x5e\x0c\x43\xbf\x39")
+	d, err := DuidFromBytes(duid)
+	if err != nil {
```
insomniacslk: We are now using `require` for unit tests, and gradually migrating to it. Can you use it instead of manual checks?

E.g. this would become
```
require.NoError(t, err)
require.Equal(t, len(duid), d.Length())
require.Equal(t, duid, d.ToBytes())
```

Thanks!

stapelberg: Sorry, no. Assert libraries are not idiomatic Go:

* they obscure control flow which is otherwise easily apparent
* they stop tests early (tests should keep going to require as few runs as possible)
* they encourage bad error messages

If you’re dead set on using an assert library regardless, please accept my commit as-is and change the test afterwards.

Thanks for understanding,

insomniacslk: No strong opinions on this, I will merge your PR as it is. Would you recommend a reading with more insight so we can discuss switching back away from assert libraries?

stapelberg: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures plays into it, but isn’t exactly focused on assert libraries. I’ll ask around.

stapelberg: I’m not aware of a public document on this topic, aside from https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures, which I mentioned in my previous comment, sorry. Maybe it’s time to start one :)

insomniacslk: it would be a nice idea. It will be easier to push people to avoid common mistakes. Thanks!

insomniacslk: oh, if you do please let me know here or over email. Thanks!

PR #74, opened 2018-06-28
Added OptDomainNameServer for DHCPv4

insomniacslk:
Added option decoder for the Domain Name Server option in DHCPv4

(tests coming in the next commit)

### DIFF
--- a/dhcpv4/option_domain_name_server.go
+++ b/dhcpv4/option_domain_name_server.go
@@ -8,7 +8,8 @@ import (
 // This option implements the domain name server option
 // https://tools.ietf.org/html/rfc2132
 
-// OptDomainNameServer represents an option encapsulating the server identifier.
+// OptDomainNameServer represents an option encapsulating the domain name
+// servers.
 type OptDomainNameServer struct {
 	NameServers []net.IP
 }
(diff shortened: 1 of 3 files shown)


dhcpv4/option_domain_name_server.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"fmt"
+	"net"
+)
+
+// This option implements the domain name server option
+// https://tools.ietf.org/html/rfc2132
+
+// OptDomainNameServer represents an option encapsulating the server identifier.
```
pmazzini: server identifier -> domain name servers

dhcpv4/option_domain_name_server.go
```diff
@@ -0,0 +1,71 @@
...
+func ParseOptDomainNameServer(data []byte) (*OptDomainNameServer, error) {
+	if len(data) < 2 {
+		return nil, ErrShortByteStream
+	}
+	code := OptionCode(data[0])
+	if code != OptionDomainNameServer {
+		return nil, fmt.Errorf("expected code %v, got %v", OptionDomainNameServer, code)
+	}
+	length := int(data[1])
+	if length%4 != 0 {
```
pmazzini: I think we should also check if the lenght is > 0, the minimun length should be 4.

dhcpv4/option_domain_name_server.go
insomniacslk: silly me. Good catch! I used another option file as template and didn't replace everything

dhcpv4/option_domain_name_server.go
insomniacslk: you are right - according to RFC2132 par3.8 the minimum length has to be four octets (one NS)

insomniacslk:
@pmazzini can you reapprove?

PR #75, opened 2018-07-05
DHCPv6.Exchange now applies modifiers correctly for both solicit and request

insomniacslk:
The logic that applies the modifiers was just wrong in Exchange():

1. it was not applied to SOLICIT packets
2. it was applied on the REQUEST *after* it was sent (wth did I have in mind?)

This patch fixes the modifiers misbehaviour


insomniacslk:
Increasing test coverage of client.go requires sending actual DHCP packets out. I will add dependency injection to mock the network requests in a future PR anyway, so for now let's just merge it with the tests we ran on our infra

PR #76, opened 2018-07-06
NewSolicitWithCID for DHCPv6

earlgreyz:
NewSolicit provides functionality between `NewMessage` and `NewSolicitForInterface`, allowing users to create a `SOLICIT` message without the need to provide an interface. This is crucial for generating fake DHCPv6 messages (ex. for a MAC address not associated with a card on the actual device).

### DIFF
--- a/dhcpv6/dhcpv6_test.go
+++ b/dhcpv6/dhcpv6_test.go
@@ -1,8 +1,10 @@
 package dhcpv6
 
 import (
+	"net"
 	"testing"
 
+	"github.com/insomniacslk/dhcp/iana"
 	"github.com/stretchr/testify/require"
 )
 
@@ -148,6 +150,38 @@ func TestNewReplyFromRebind(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, rep.(*DHCPv6Message).TransactionID(), reb.TransactionID())
 	require.Equal(t, rep.Type(), REPLY)
+}
+
+func TestNewSolicitWithCID(t *testing.T) {
+	hwAddr, err := net.ParseMAC("24:0A:9E:9F:EB:2B")
+	require.NoError(t, err)
+
+	duid := Duid{
+		Type:          DUID_LL,
+		HwType:        iana.HwTypeEthernet,
+		LinkLayerAddr: hwAddr,
+	}
+
+	s, err := NewSolicitWithCID(duid)
+	require.NoError(t, err)
+
+	require.Equal(t, s.Type(), SOLICIT)
+	// Check CID
+	cidOption := s.GetOneOption(OPTION_CLIENTID)
+	require.NotNil(t, cidOption)
+	cid, ok := cidOption.(*OptClientId)
+	require.True(t, ok)
+	require.Equal(t, cid.Cid, duid)
+
+	// Check ORO
+	oroOption := s.GetOneOption(OPTION_ORO)
+	require.NotNil(t, oroOption)
+	oro, ok := oroOption.(*OptRequestedOption)
+	require.True(t, ok)
+	opts := oro.RequestedOptions()
+	require.Contains(t, opts, DNS_RECURSIVE_NAME_SERVER)
+	require.Contains(t, opts, DOMAIN_SEARCH_LIST)
+	require.Equal(t, len(opts), 2)
 }
 
 // TODO test NewSolicit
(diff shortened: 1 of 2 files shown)


dhcpv6/dhcpv6message.go
```diff
@@ -65,14 +65,42 @@ func GetTime() uint32 {
 	return uint32((now.Nanoseconds() / 1000000000) % 0xffffffff)
 }
 
+// NewSolicit creates a new SOLICIT message without populating the CID field.
+func NewSolicit(modifiers ...Modifier) (DHCPv6, error) {
```
pmazzini: How about passing a Client DUID as parameter so that it returns a valid SOLICIT?

insomniacslk: I asked him to intentionally skip the CID - this function will be used for packets where we want custom CIDs so they can be added either via modifier, or later by the user. However we still should specify an empty CID (e.g. DUID_LL with 00:00:00:00:00:00 link-local addr) so that the packet is still valid, even if with an invalid CID. It is easy to detect the null address in a pcap dump if the user forgets to set it.

In addition, we could add `NewSolicitWithCID` but I am generally against adding functions unless necessary

dhcpv6/dhcpv6_test.go
```diff
@@ -150,5 +150,21 @@ func TestNewReplyFromRebind(t *testing.T) {
...
+	require.Equal(t, s.Type(), SOLICIT)
+
+	oroOption := s.GetOneOption(OPTION_ORO)
+	require.NotNil(t, oroOption)
+	oro, ok := oroOption.(*OptRequestedOption)
+	opts := oro.RequestedOptions()
+	require.True(t, ok)
+	require.Contains(t, opts, DNS_RECURSIVE_NAME_SERVER)
+	require.Contains(t, opts, DOMAIN_SEARCH_LIST)
+	require.Equal(t, len(opts), 2)
```
insomniacslk: after adding the empty CID we also need to test that it exists and has a zeroed LL addr

dhcpv6/dhcpv6message.go
```diff
@@ -65,14 +65,42 @@ func GetTime() uint32 {
...
+	}
+	return d, nil
+}
+
 // NewSolicitForInterface creates a new SOLICIT message with DUID-LLT, using the
 // given network interface's hardware address and current time
 func NewSolicitForInterface(ifname string, modifiers ...Modifier) (DHCPv6, error) {
-	d, err := NewMessage()
+	// We want modifiers to be applied later
+	d, err := NewSolicit()
```
insomniacslk: you can further simplify this function by computing the DUID before, and applying all the modifiers via NewSolicit instead of the loop at the end:

```
// get the CID for the specified interface
iface, err := net.InterfaceByName(iface)
if err ...

modifiers = append(modifiers, WithClientID(duid))
d, err := NewSolicit(modifiers)
```

insomniacslk: something we may want to decide here is whether the CID modifier should take precedence on the other modifiers or not. For example, the caller of `NewSolicitForInterface` also passes `WithClientID(...)`. Do we want this to take precedence, or to give precedence to the `WithClientID` passed inside the function?

dhcpv6/dhcpv6message.go
```diff
@@ -65,49 +65,50 @@ func GetTime() uint32 {
...
 		DNS_RECURSIVE_NAME_SERVER,
 		DOMAIN_SEARCH_LIST,
 	})
-	d.AddOption(&oro)
-	d.AddOption(&OptElapsedTime{})
+	d.AddOption(oro)
+	d.AddOption(new(OptElapsedTime))
 	// FIXME use real values for IA_NA
-	iaNa := OptIANA{}
+	iaNa := new(OptIANA)
```
insomniacslk: note: there is no advantage or disadvantage in using `new` vs `&OptIana{}` here - in this specific case it's just style preference. This one can go, but in general I prefer we don't make unnecessary style changes

dhcpv6/dhcpv6_test.go
```diff
@@ -150,5 +150,20 @@ func TestNewReplyFromRebind(t *testing.T) {
 	require.Equal(t, rep.Type(), REPLY)
 }
 
+func TestNewSolicitWithCID(t *testing.T) {
```
insomniacslk: you should also check for the presence of `OptClientID`, and use a non-zero-value DUID, so the tests fails if it's not set to the right value

insomniacslk:
I can't actually merge this because the code coverage went down

PR #77, opened 2018-07-07
add DelOption to OptIANA

pmazzini:


dhcpv6/dhcpv6.go
```diff
@@ -117,6 +117,16 @@ func getOption(options []Option, code OptionCode) Option {
 	return opts[0]
 }
 
+func delOption(options []Option, code OptionCode) []Option {
+	for i := 0; i < len(options); i++ {
+			if options[i].Code() == code {
+			options = append(options[:i], options[i+1:]...)
```
insomniacslk: if you modify the options slice while you iterate the index has to be adjusted. This should not work with contiguous options (e.g. two subsequent ClientID).
It's easier (and also uses less allocations) to use another array instead, with preallocated capacity.

E.g.

```
newOpts := make([]Option, 0, len(options))
for _, opt := range options {
    if opt.Code() == code {
        newOpts = append(newOpts, opt)
    }
}
return newOpts
```

dhcpv6/option_nontemporaryaddress_test.go
```diff
@@ -39,6 +39,15 @@ func TestOptIANAParseOptIANAInvalidOptions(t *testing.T) {
 	require.Error(t, err)
 }
 
+func TestOptIANADelOption(t *testing.T) {
```
insomniacslk: You should also test the case of multiple options, subsequent and non-subsequent

PR #80, opened 2018-07-09
Asynchronous client for DHCPv6

earlgreyz:


dhcpv6/async.go
```diff
@@ -0,0 +1,214 @@
...
+	"context"
+	"fmt"
+	"log"
+	"net"
+	"sync"
+	"time"
+)
+
+// ClientAsync implements an asynchronous DHCPv6 client
+type ClientAsync struct {
```
insomniacslk: nit/language: what about `AsyncClient` ?

dhcpv6/async.go
```diff
@@ -0,0 +1,214 @@
...
+	ReadTimeout  time.Duration
+	WriteTimeout time.Duration
+	LocalAddr    net.Addr
+	RemoteAddr   net.Addr
```
insomniacslk: these four fields above are also all of the fields in `Client`. What about embedding it?

dhcpv6/async.go
```diff
@@ -0,0 +1,214 @@
...
+
+// NewClientAsync creates an asynchronous client
+func NewClientAsync() *ClientAsync {
+	return &ClientAsync{
+		ReadTimeout:  DefaultReadTimeout,
+		WriteTimeout: DefaultWriteTimeout,
+	}
+}
+
+// OpenForInterface starts the client on the specified interface
```
insomniacslk: mention that it requires a link-local address to work, and that it overrides the listening port to 546

dhcpv6/async.go
```diff
@@ -0,0 +1,214 @@
...
+	go c.listen(ctx)
+	go c.broadcast(ctx)
+
+	return nil
+}
+
+// Close stops the client
+func (c *ClientAsync) Close() {
+	c.cancel()
+	<-c.stopping
```
insomniacslk: this looks like a good candidate for `sync.WaitGroup` rather than a channel

dhcpv6/async.go
```diff
@@ -0,0 +1,214 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case packet := <-c.receiveQueue:
+			c.receive(packet)
+		}
+	}
+}
+
+func (c *ClientAsync) broadcast(ctx context.Context) {
```
insomniacslk: this function should be called differently, there is no broadcasting involved (and if we really want, at most it's multicasting). I think a more appropriate name could be `senderLoop`. Similarly `listen` above should be named `receiverLoop` or similar

dhcpv6/async.go
```diff
@@ -0,0 +1,214 @@
...
+		case <-ctx.Done():
+			return
+		case packet := <-c.sendQueue:
+			c.send(packet)
+		}
+	}
+}
+
+func (c *ClientAsync) send(packet DHCPv6) {
+	transactionID, _ := GetTransactionID(packet)
```
insomniacslk: don't ignore the error, log it and return instead. The side effect of ignoring the error is that you will use 0 as transaction ID in the rest of the function

dhcpv6/async.go
```diff
@@ -0,0 +1,214 @@
...
+
+	c.connection.SetReadDeadline(time.Now().Add(c.ReadTimeout))
+	for {
+		buffer := make([]byte, maxUDPReceivedPacketSize)
+		n, _, _, _, err := c.connection.ReadMsgUDP(buffer, oobdata)
+		if err != nil {
+			if err, ok := err.(net.Error); ok && err.Timeout() {
+				return
+			}
+			log.Fatalf("Error receiving the message: %s", err)
```
insomniacslk: don't use `log.Fatalf` in library code - it will log the message and call `os.Exit`, basically ignoring any defer or other cleanup operation, and abruptly terminating the program.

Rather, log the error and return. It would be nice to have an errors channel

dhcpv6/async.go
earlgreyz: This error cannot occur here, we have checked for this packet having a transaction id before puting it to the queue `async.go:191`

insomniacslk: you can still defensively log it, as you don't know how the code will evolve in the future - maybe with `log.Printf("Warning: this should never happen, there is no transaction ID")`

dhcpv6/async.go
earlgreyz: This might be confusing as it will give the access to synchronous methods on an async client

insomniacslk: you're absolutely right, I didn't think of that

PR #81, opened 2018-07-09
GetGlobalAddr utility function

earlgreyz:
A complementary function for `GetLinkLocalAddr`, but allows to get a global address for the interface. Helpful to setup a client as a relay forwarder.

### DIFF
--- a/dhcpv6/async/client.go
+++ b/dhcpv6/async/client.go
@@ -45,7 +45,7 @@ func (c *Client) OpenForInterface(ifname string, bufferSize int) error {
 	if err != nil {
 		return err
 	}
-	c.LocalAddr = &net.UDPAddr{IP: *addr, Port: dhcpv6.DefaultClientPort, Zone: ifname}
+	c.LocalAddr = &net.UDPAddr{IP: addr, Port: dhcpv6.DefaultClientPort, Zone: ifname}
 	return c.Open(bufferSize)
 }
 
--- a/dhcpv6/client.go
+++ b/dhcpv6/client.go
@@ -100,7 +100,7 @@ func (c *Client) sendReceive(ifname string, packet DHCPv6, expectedType MessageT
 		if err != nil {
 			return nil, err
 		}
-		laddr = net.UDPAddr{IP: *llAddr, Port: DefaultClientPort, Zone: ifname}
+		laddr = net.UDPAddr{IP: llAddr, Port: DefaultClientPort, Zone: ifname}
 	} else {
 		if addr, ok := c.LocalAddr.(*net.UDPAddr); ok {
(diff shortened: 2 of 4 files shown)


insomniacslk:
This function is almost entirely the same as GetLinkLocalAddr, basically just one character is changed. Can you refactor it in a backend function and two wrappers around it?

Also please increase test coverage, the branch is protected against diffs that reduce the test coverage

earlgreyz:
`iputils` have no tests at all, the actual tests would require the assumptions about the machines which run the tests... (It loops directly the interfaces on the machine)

insomniacslk:
I know that there are are no tests, which is why we enforced that coverage doesn't go down :)

The way to do that without a mocking framework is to inject dependencies, in this case through a global var since there's no struct involved. If you are not interested in doing that, I can temporarily remove branch protection, merge this, and implement the test it myself

stapelberg:
This commit broke https://github.com/rtr7/router7:

```
github.com/rtr7/router7/internal/dhcp6/dhcp6.go:92:10: invalid indirect of llAddr (type net.IP)
```

Maybe this would be a good time to start using Go modules, so that breaking changes get a bump to the import path?

insomniacslk:
ouch, my fault for not realizing that this would have broken the API, sorry @stapelberg !

On versioned modules: I did an attempt on https://github.com/insomniacslk/dhcp/pull/141 , but I think there may be a few more (potentially breaking) changes similar to the one you saw here, before we go with a version 1 with a stable API.

What if instead I enforce that we don't break router7 builds, as a stop-gap? I added a few lines in the Travis-CI build that fetch router7 and build `./cmds/...`. Would that be good enough for you until we switch to v1 with versioned modules?

stapelberg:
That seems like a good stop-gap solution. Let me know once you go with modules, though :)

insomniacslk:
Will do :)

insomniacslk:
hey @stapelberg , while testing that the router7 builds don't break because of this library's changes, I found the following:

```
../../rtr7/router7/cmd/captured/captured.go:48:18: undefined: pcapgo.OpenEthernet
```

Apparently that function doesn't exist anymore in gopacket/pcapgo . For some reason the build is not failing on router7's CI. This seems something to look into on your side for the `captured` command, would you mind?

stapelberg:
Have a look at https://github.com/rtr7/router7/blob/aa404c3c54d9ad655479d7978ed18e81fe6ca05c/.travis.yml#L14

Let me try to escalate the pcapgo pull request to get this fixed properly.

insomniacslk:
thanks - I'll do the same here for now, so I can merge this PR

PR #83, opened 2018-07-12
add DecapsulateRelayIndex

pmazzini:


dhcpv6/dhcpv6relay.go
```diff
@@ -158,28 +158,30 @@ func (d *DHCPv6Relay) GetInnerMessage() (DHCPv6, error) {
...
-		err error
-	)
-	p = r
-	hops := r.HopCount()
-	addr := r.PeerAddr()
-	for i := uint8(0); i < hops; i++ {
-		p, err = DecapsulateRelay(p)
+// GetInnerRelay recurses into a relay message and extract and return the inner
+// DHCPv6Relay. Return nil if none found (e.g. not a relay message).
+func (r *DHCPv6Relay) GetInnerRelay() (DHCPv6, error) {
```
insomniacslk: do we really need GetInnerRelay and GetInnerPeerAddr? The former is too specialized (it returns the last relay - what about the other ones?), and the latter probably doesn't belong to the library at all. I'm saying this because I'm afraid to overcomplicate the API with features that are not general or useful enough.

What I'm thinking instead:
* let's generalize DecapsulateRelay to accept an integer as index (e.g. if 0 return the outermost relay, 1 returns the second, etc, and -1 returns the last)
* let's keep GetInnerPeerAddr out of the library: it's simple enough (and too specialized) to be implemented in the user code rather than here.

What do you think?

pmazzini: Yes, it sounds good.

pmazzini:
Ok.
> This because relay.HopCount() may not be reliable (e.g. if it is set to 5 in a message with a single nested relay)
Shouldn't we return an error in that case instead of trying to handle a non RFC compliant packet?

insomniacslk:
lgtm, thanks!

PR #86, opened 2018-07-16
Netboot: don't fail if no netboot url is returned

insomniacslk:
Currently a netboot attempt will fail if no netboot  URL is returned by the DHCPv6 server.

This patch doesn't make it fail anymore - just the returned URL will be an empty string

### DIFF
--- a/netboot/netboot.go
+++ b/netboot/netboot.go
@@ -87,7 +87,9 @@ func ConversationToNetconf(conversation []dhcpv6.DHCPv6) (*NetConf, string, erro
 			return nil, "", errors.New("no bootfile URL option found in ADVERTISE")
 		}
 	}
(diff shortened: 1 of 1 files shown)


netboot/netboot.go
```diff
@@ -87,7 +87,9 @@ func ConversationToNetconf(conversation []dhcpv6.DHCPv6) (*NetConf, string, erro
 			return nil, "", errors.New("no bootfile URL option found in ADVERTISE")
 		}
 	}
-	obf := opt.(*dhcpv6.OptBootFileURL)
-	bootfile = string(obf.BootFileURL)
+	if opt != nil {
```
pmazzini: I would change this for an else since it is already being checked that opt == nil in line 71.

PR #90, opened 2018-07-24
Added UserClass option to DHCPv4

mooneyow:
Added support for UserClass options for DHCPv4 as per the [RFC](https://tools.ietf.org/html/rfc3004)

### DIFF
--- a/dhcpv4/option_userclass.go
+++ b/dhcpv4/option_userclass.go
@@ -69,7 +69,7 @@ func ParseOptUserClass(data []byte) (*OptUserClass, error) {
 	for i := 0; i < totalLength; {
 		ucLen := int(data[i])
 		if ucLen == 0 {
-			return nil, errors.New("User Class value has invalid lenght of 0")
+			return nil, errors.New("User Class value has invalid length of 0")
 		}
 		base := i + 1
 		if len(data) < base+ucLen {
--- a/dhcpv4/option_userclass.go
+++ b/dhcpv4/option_userclass.go
@@ -55,21 +55,22 @@ func ParseOptUserClass(data []byte) (*OptUserClass, error) {
 	if code != OptionUserClassInformation {
 		return nil, fmt.Errorf("expected code %v, got %v", OptionUserClassInformation, code)
 	}
-	data = data[2:]
 
-	for {
-		if len(data) == 0 {
-			break
+	totalLength := int(data[1])
+	data = data[2:]
+	if len(data) < totalLength {
+		return nil, fmt.Errorf("ParseOptUserClass: short data: length is %d but got %d bytes",
+			totalLength, len(data))
+	}
+
+	for i := 0; i < totalLength; {
+		ucLen := int(data[i])
+		opaqueDataIndex := i + ucLen + 1
+		if len(data) < opaqueDataIndex {
+			return nil, fmt.Errorf("ParseOptUserClass: short data: less than %d bytes", opaqueDataIndex)
 		}
-		if len(data) < 1 {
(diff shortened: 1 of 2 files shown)


dhcpv4/option_userclass.go
```diff
@@ -0,0 +1,79 @@
...
+			totalLength, len(data))
+	}
+
+	for i := 0; i < totalLength; {
+		ucLen := int(data[i])
+		opaqueDataIndex := i + ucLen + 1
+		if len(data) < opaqueDataIndex {
+			return nil, fmt.Errorf("ParseOptUserClass: short data: less than %d bytes", opaqueDataIndex)
+		}
+		opt.UserClasses = append(opt.UserClasses, data[i+1:opaqueDataIndex])
```
insomniacslk: nit: for clarity I prefer that we express the slice indexes in terms of `[something:something+something_else]`, i.e. don't use `opaqueDataIndex`, but just use `i` instead. E.g.
```
base := i + 1
if len(data) < base + ucLen {
    // error
}
opt.UserClasses = append(opt.UserClasses, data[base:base+ucLen])
```

dhcpv4/option_userclass.go
```diff
@@ -0,0 +1,79 @@
...
+	if len(data) < totalLength {
+		return nil, fmt.Errorf("ParseOptUserClass: short data: length is %d but got %d bytes",
+			totalLength, len(data))
+	}
+
+	for i := 0; i < totalLength; {
+		ucLen := int(data[i])
+		opaqueDataIndex := i + ucLen + 1
+		if len(data) < opaqueDataIndex {
+			return nil, fmt.Errorf("ParseOptUserClass: short data: less than %d bytes", opaqueDataIndex)
```
insomniacslk: also print the actual size, it will simplify troubleshooting:
```
return nil, fmt.Errorf("ParseOptUserClass: short data: %d bytes; want: %d", len(data), opaqueDataIndex)
```

(the "got.. want" comes from Go's code review advice, https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures )

dhcpv4/option_userclass_test.go
```diff
@@ -0,0 +1,92 @@
...
+		4, 't', 'e', 's', 't',
+	}
+	opt, err := ParseOptUserClass(expected)
+	require.NoError(t, err)
+	require.Equal(t, len(opt.UserClasses), 2)
+	require.Equal(t, []byte("linuxboot"), opt.UserClasses[0])
+	require.Equal(t, []byte("test"), opt.UserClasses[1])
+}
+
+func TestParseOptUserClassNone(t *testing.T) {
```
insomniacslk: also write a function to test zero-length user classes, e.g. when
```
data := []byte{0}
```
and
```
data := []byte{10, 'l', 'i', 'n', 'u', 'x', 'b', 'o', 'o', 't', 0}
```

dhcpv4/option_userclass_test.go
```diff
@@ -0,0 +1,92 @@
...
+	_, err := ParseOptUserClass(expected)
+	require.Error(t, err)
+}
+
+func TestParseOptUserClass(t *testing.T) {
+	expected := []byte{
+		77, 10, 9, 'l', 'i', 'n', 'u', 'x', 'b', 'o', 'o', 't',
+	}
+	opt, err := ParseOptUserClass(expected)
+	require.NoError(t, err)
```
insomniacslk: also require that `opt.Code() == OptionUserClassInformation` and `opt.Length() == 10`

mooneyow:
Updated with requested changes 😁

PR #91, opened 2018-07-24
Add support for Bootfile name option

mooneyow:
[RFC](https://tools.ietf.org/html/rfc2132)

### DIFF
--- a/dhcpv4/option_bootfile_name.go
+++ b/dhcpv4/option_bootfile_name.go
@@ -39,16 +39,16 @@ func ParseOptBootfileName(data []byte) (*OptBootfileName, error) {
 	}
 	code := OptionCode(data[0])
 	if code != OptionBootfileName {
-		return nil, fmt.Errorf("expected code %v, got %v", OptionBootfileName, code)
+		return nil, fmt.Errorf("ParseOptBootfileName: invalid code: %v; want %v", code, OptionBootfileName)
 	}
 	length := int(data[1])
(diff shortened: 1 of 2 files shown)


dhcpv4/option_bootfile_name.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// ParseOptBootfileName returns a new OptBootfile from a byte stream or error if any
+func ParseOptBootfileName(data []byte) (*OptBootfileName, error) {
+	if len(data) < 3 {
+		return nil, ErrShortByteStream
+	}
+	code := OptionCode(data[0])
+	if code != OptionBootfileName {
+		return nil, fmt.Errorf("expected code %v, got %v", OptionBootfileName, code)
```
insomniacslk: use the same error message format as you do on line 50

(i.e. `"ParseOptBootfileName: invalid code: %v; want: %v", code, OptionBootfileName`)

dhcpv4/option_bootfile_name.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	length := int(data[1])
+	if length < 1 {
+		return nil, fmt.Errorf("Bootfile name has invalid length of %d", length)
+	}
+	data = data[2:]
```
insomniacslk: use a different name for the new variable so the code is easier to read

mooneyow:
Updated with changes 😁

PR #92, opened 2018-07-25
Add TFTP server name option

mooneyow:
[RFC](https://tools.ietf.org/html/rfc2132)


dhcpv4/options.go
```diff
@@ -60,6 +60,12 @@ func ParseOption(data []byte) (Option, error) {
 		opt, err = ParseOptDomainNameServer(data)
 	case OptionVendorIdentifyingVendorClass:
 		opt, err = ParseOptVIVC(data)
+	case OptionTFTPServerName:
+		opt, err = ParseOptTFTPServerName(data)
+	case OptionBootfileName:
```
insomniacslk: thanks for adding these ones too!

PR #98, opened 2018-07-27
Fix indentation

lestrozi:


get9:
I'm surprised gofmt didn't fix this at some point.

insomniacslk:
and I'm surprised that Stickler bot didn't catch it either :(

PR #100, opened 2018-07-28
add NewReplyFromRequest

pmazzini:
add NewReplyFromRequest which builds a DHCPv4 reply from a request.

Also add more logic to the AddOption to insert options before the OptionEnd.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -571,8 +571,9 @@ func (d *DHCPv4) SetOptions(options []Option) {
 	d.options = options
 }
 
-// AddOption appends an option to the existing ones.
-// If the last option is an OptionEnd, it will be inserted before that.
+// AddOption appends an option to the existing ones. If the last option is an
+// OptionEnd, it will be inserted before that. It does not deal with End
+// options that appead before the end, like in malformed packets.
 func (d *DHCPv4) AddOption(option Option) {
 	if len(d.options) == 0 || d.options[len(d.options)-1].Code() != OptionEnd {
 		d.options = append(d.options, option)
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -217,8 +217,8 @@ func RequestFromOffer(offer DHCPv4) (*DHCPv4, error) {
 	return d, nil
 }
 
-// NewReply builds a DHCPv4 reply from a request.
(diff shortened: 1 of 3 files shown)


dhcpv4/dhcpv4.go
```diff
@@ -214,11 +214,26 @@ func RequestFromOffer(offer DHCPv4) (*DHCPv4, error) {
 	d.AddOption(&OptMessageType{MessageType: MessageTypeRequest})
 	d.AddOption(&OptRequestedIPAddress{RequestedAddr: offer.YourIPAddr()})
 	d.AddOption(&OptServerIdentifier{ServerID: serverIP})
-	// the End option has to be added explicitly
-	d.AddOption(&OptionGeneric{OptionCode: OptionEnd})
 	return d, nil
 }
 
+// NewReply builds a DHCPv4 reply from a request.
+func NewReply(request *DHCPv4) (*DHCPv4, error) {
```
insomniacslk: shall we use the usual naming `NewReplyFromSomething` ?

dhcpv4/dhcpv4.go
```diff
@@ -557,8 +572,15 @@ func (d *DHCPv4) SetOptions(options []Option) {
 }
 
 // AddOption appends an option to the existing ones.
+// If there is an OptionEnd in the list, it will insert it before that option.
```
insomniacslk: more precisely - if the last option is an OptionEnd, it will be inserted before that. A malformed packet may contain more than one OptionEnd, or more in general it may not be the last. We should either update this comment, or scan for the first OptionEnd and insert the new option there

pmazzini: Ok, I will update the comment.

dhcpv4/dhcpv4.go
pmazzini: Ok, does `NewReplyFromRequest` sounds good?

dhcpv4/dhcpv4.go
```diff
@@ -557,8 +572,15 @@ func (d *DHCPv4) SetOptions(options []Option) {
 }
 
 // AddOption appends an option to the existing ones.
+// If the last option is an OptionEnd, it will be inserted before that.
```
insomniacslk: Thanks for updating the comment. What I meant here is to state explicitly that this doesn't deal with End options that appear before the end, like in malformed packets. Not stating that may be implicitly misleading

PR #101, opened 2018-07-29
add OptSubnetMask

pmazzini:


pmazzini:
Thanks!

dhcpv4/option_subnet_mask.go
```diff
@@ -0,0 +1,56 @@
+package dhcpv4
+
+import (
+	"fmt"
+	"net"
+)
+
+// This option implements the subnet mask option
+// https://tools.ietf.org/html/rfc2132
```
insomniacslk: nit: I'd have liked the text link here for consistency with other options

dhcpv4/option_subnet_mask.go
```diff
@@ -0,0 +1,56 @@
...
+
+// OptSubnetMask represents an option encapsulating the subnet mask.
+type OptSubnetMask struct {
+	SubnetMask net.IPMask
+}
+
+// ParseOptSubnetMask returns a new OptSubnetMask from a byte
+// stream, or error if any.
+func ParseOptSubnetMask(data []byte) (*OptSubnetMask, error) {
+	if len(data) < 2 {
```
insomniacslk: nit: the length should be always 6 here. This would remove the need for the redundant check below that `len(data) == 6`. From the RFC:

> The code for the subnet mask option is 1, and its length is 4 octets.

dhcpv4/option_subnet_mask.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Code returns the option code.
+func (o *OptSubnetMask) Code() OptionCode {
+	return OptionSubnetMask
+}
+
+// ToBytes returns a serialized stream of bytes for this option.
+func (o *OptSubnetMask) ToBytes() []byte {
+	ret := []byte{byte(o.Code()), byte(o.Length())}
+	return append(ret, o.SubnetMask[:4]...)
```
insomniacslk: nit: this should be `.To4()` instead of `[:4]`. It's a theoretical issue here, because we expect this to be set by ParseOptSubnetMask, but more in general this only works correctlyif len(o.SubnetMask) is 4

PR #102, opened 2018-07-29
Add support for modifiers to dhcpv4

mooneyow:


pmazzini:
If possible please add support for modifiers to the recently added `NewReplyFromRequest` #100 :)

mooneyow:
Done 😁

pmazzini:
Thanks!

dhcpv4/client.go
```diff
@@ -140,10 +140,14 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4) ([]DHCPv4, error) {
 
 	// Discover
 	if discover == nil {
-		discover, err = NewDiscoveryForInterface(ifname)
+		discover, err = NewDiscoveryForInterface(ifname, modifiers...)
```
pmazzini: I wouldn't call `NewDiscoveryForInterface` with modifiers.
I would apply the modifiers once we have the discover packet.
I think we can also remove the else statement in that case.

dhcpv4/client.go
```diff
@@ -124,7 +124,7 @@ func MakeListeningSocket(ifname string) (int, error) {
 // ordered as Discovery, Offer, Request and Acknowledge. In case of errors, an
 // error is returned, and the list of DHCPv4 objects will be shorted than 4,
 // containing all the sent and received DHCPv4 messages.
-func (c *Client) Exchange(ifname string, discover *DHCPv4) ([]DHCPv4, error) {
+func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier) ([]DHCPv4, error) {
```
get9: If I'm understanding this correctly, this will apply `modifiers` to both the DISCOVER and REQUEST packets. Are all potential modifiers applicable to both packets? How would you restrict which modifiers get applied to one or the other? Is that even desired functionality?

mooneyow: Yes it will apply to both. 
*Are all potential modifiers applicable to both packets?* I'm not sure.
*How would you restrict which modifiers get applied to one or the other?* With the code in this pull request you can't.
*Is that even desired functionality?* I'm trying to be consistent with [how the DHCPv6 client does it](https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/client.go#L43)

get9: Ah, didn't realize this was already in dhcpv6.

get9:
Also probably worth adding this to the `bsdp` subpackage as well.

insomniacslk:
> Also probably worth adding this to the bsdp subpackage as well. 

I've created https://github.com/insomniacslk/dhcp/issues/116 and assigned to you :P

PR #107, opened 2018-07-30
DHCPv6: Other ALL_CAPS constants to CamelCase

get9:
This changes all other ALL_CAPS constants in various other files in
dhcpv6 to CamelCase. These are a bit more controversial changes and IMO
don't enhance readability (in some cases, it makes it worse)... Feel
free to comment whether you agree or disagree with these changes.

### DIFF
--- a/dhcpv6/duid.go
+++ b/dhcpv6/duid.go
@@ -12,17 +12,17 @@ type DuidType uint16
 
 // DUID types
 const (
-	DUID_LLT  DuidType = 1
-	DUID_EN   DuidType = 2
-	DUID_LL   DuidType = 3
-	DUID_UUID DuidType = 4
+	DuidLLT  DuidType = 1
+	DuidEN   DuidType = 2
+	DuidLL   DuidType = 3
+	DuidUUID DuidType = 4
 )
 
 var DuidTypeToString = map[DuidType]string{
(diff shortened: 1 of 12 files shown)


insomniacslk:
I agree with you that this may not improve readability, rather just make the go linter happier.
In cases like DUID_LLT -> DuidLLT, it should probably be DUIDLLT because it's all acronyms, but it would be worse than the other two options. I wouldn't like DuidLlt either. However I don't have a strong opinion on this. @pmazzini what's your thought?

pmazzini:
I don't have a strong opinion about it either.

insomniacslk:
hey @get9 , sorry for making you wait. Given that this would be a breaking change (and in https://github.com/insomniacslk/dhcp/issues/123 I committed to doing no breaking changes before versioning - or, I'd add now, unless really necessary), and that there is no strong reasoning in either direction, I'd rather not do this change and delay it until later

dhcpv6/duid.go
```diff
@@ -12,17 +12,17 @@ type DuidType uint16
 
 // DUID types
 const (
-	DUID_LLT  DuidType = 1
```
hugelgupf: Given the Go standard, these should actually be `DUIDLLT` and so on.

PR #108, opened 2018-07-30
add OptIPAddressLeaseTime

pmazzini:


dhcpv4/option_ip_address_lease_time.go
```diff
@@ -0,0 +1,57 @@
...
+// OptIPAddressLeaseTime represents the IP Address Lease Time option.
+type OptIPAddressLeaseTime struct {
+	LeaseTime uint32
+}
+
+// ParseOptIPAddressLeaseTime constructs an OptIPAddressLeaseTime struct from a
+// sequence of bytes and returns it, or an error.
+func ParseOptIPAddressLeaseTime(data []byte) (*OptIPAddressLeaseTime, error) {
+	// Should at least have code, length, and lease time.
+	if len(data) < 4 {
```
insomniacslk: this should have been 6, not 4. Can we also add a test to catch short options where len(data) doesn't match the opt len?
e.g. data := []byte{51, 4, 168, 192}

pmazzini: Good catch, I opened #109

PR #111, opened 2018-07-31
add OptDomainSearch

pmazzini:


dhcpv4/option_rfc1035label.go
```diff
@@ -0,0 +1,53 @@
+package dhcpv4
```
insomniacslk: isn't this the same as dhcpv6/option_rfc1035label.go ? Shall we extract it instead of duplicating?

dhcpv4/option_domain_search.go
```diff
@@ -0,0 +1,61 @@
+package dhcpv4
+
+// This module defines the OptDomainSearch structure.
+// https://tools.ietf.org/html/rfc3397
```
insomniacslk: use the text link for consistency with other options, https://tools.ietf.org/rfc/rfc3397.txt

dhcpv4/option_rfc1035label.go
pmazzini: Yes, the same. Do you mean extract it to a directory inside `insomniacslk/dhcp`?
Sounds good. How should we call it?

dhcpv4/option_domain_search.go
pmazzini: I think all options are using https://tools.ietf.org/html/rfc2132

dnscompress/label.go
```diff
@@ -1,10 +1,14 @@
-package dhcpv6
+package dnscompress
```
insomniacslk: what about calling this simply `rfc1035label`?

insomniacslk:
There's also a small coverage decrease, can you look at that too?

dnscompress/label.go
pmazzini: Sounds good.

pmazzini:
Yes, I'll look at it.

PR #112, opened 2018-07-31
add OptHostName

pmazzini:


dhcpv4/option_host_name.go
```diff
@@ -0,0 +1,49 @@
+package dhcpv4
+
+import "fmt"
+
+// This option implements the host name option
+// https://tools.ietf.org/html/rfc2132
```
insomniacslk: use https://tools.ietf.org/rfc/rfc2132.txt for consistency with other options

pmazzini: I think all options are using https://tools.ietf.org/html/rfc2132

insomniacslk: ok - I was quite sure I stuck on txt format but let's proceed here, can be changed later

Issue #113, opened 2018-07-31
dhcpv4 user class parsing


pmazzini:
The user class parsing added in PR #90 does not follow the RFC:

> The value in UC_Len_i does not include the length field itself and MUST be non-zero.

pmazzini:
cc @insomniacslk, @mooneyow

pmazzini:
I may have opened this to soon.
Apparently some clients are sending a non rfc compliant user class option.

mooneyow:
It seems there is a conflicting Microsoft standard where `UC_Len_i` is not included and only one user class can be specified.
https://osqa-ask.wireshark.org/answer_link/55006/

https://msdn.microsoft.com/en-us/library/dd358019.aspx

mooneyow:
https://github.com/insomniacslk/dhcp/pull/114

pmazzini:
Closing this after PR117.

PR #114, opened 2018-08-01
Parse non-RFC compliant user class

mooneyow:
https://github.com/insomniacslk/dhcp/issues/113

### DIFF
--- a/dhcpv4/option_userclass.go
+++ b/dhcpv4/option_userclass.go
@@ -66,6 +66,23 @@ func ParseOptUserClass(data []byte) (*OptUserClass, error) {
 			totalLength, len(data))
 	}
 
+	// Check if option is Microsoft style instead of RFC compliant, issue #113
+
+	// User-class options are, according to RFC3004, supposed to contain a set
+	// of strings each with length UC_Len_i. Here we check that this is so,
+	// by seeing if all the UC_Len_i lengths are consistent with the overall
+	// option length. If the lengths don't add up, we assume that the option
+	// is a single string and non RFC3004 compliant
+	var counting int
+	for counting < totalLength {
+		// UC_Len_i does not include itself so add 1
+		counting += int(data[counting]) + 1
+	}
+	if counting != totalLength {
+		opt.UserClasses = append(opt.UserClasses, data[:totalLength])
+		return &opt, nil
+	}
+
 	for i := 0; i < totalLength; {
 		ucLen := int(data[i])
 		if ucLen == 0 {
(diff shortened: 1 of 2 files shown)


dhcpv4/option_userclass.go
```diff
@@ -66,6 +66,18 @@ func ParseOptUserClass(data []byte) (*OptUserClass, error) {
 			totalLength, len(data))
 	}
 
+	// Check if option is Microsoft style instead of RFC compliant
+	var counting int
+	for i := 0; i < totalLength; {
+		// UC_Len_i does not include itself so add 1
+		counting += int(data[i]) + 1
+		i += int(data[i]) + 1
```
pmazzini: `counting` and `i` look to be the same, can we have only one of them?

pmazzini:
I was initially reluctant to add this behaviour but after seeing that another client is also doing this changed my mind.

Waiting from @insomniacslk's review.

dhcpv4/option_userclass.go
mooneyow: My bad, fixed

dhcpv4/option_userclass.go
```diff
@@ -66,6 +66,17 @@ func ParseOptUserClass(data []byte) (*OptUserClass, error) {
...
 	}
 
+	// Check if option is Microsoft style instead of RFC compliant
+	var counting int
+	for counting < totalLength {
+		// UC_Len_i does not include itself so add 1
+		counting += int(data[counting]) + 1
+	}
+	if counting != totalLength {
+		opt.UserClasses = append(opt.UserClasses, data)
```
pmazzini: I think this should be `opt.UserClasses = append(opt.UserClasses, data[:totalLength])`
Can a `TestParseOptUserClassMicrosoftLongerThanLength` be added?
I think it will fail if that is not changed.

dhcpv4/option_userclass.go
```diff
@@ -66,6 +66,17 @@ func ParseOptUserClass(data []byte) (*OptUserClass, error) {
 			totalLength, len(data))
 	}
 
+	// Check if option is Microsoft style instead of RFC compliant
```
insomniacslk: can you please add a reference to the issue? Either a reference to https://github.com/insomniacslk/dhcp/issues/113 or to the links you posted there. Thanks!

insomniacslk: also mention the heuristic (similarly to how it's done in dnsmasq)

PR #115, opened 2018-08-01
add OptNTPServers

pmazzini:


dhcpv4/option_ntp_servers.go
```diff
@@ -0,0 +1,70 @@
...
+// String returns a human-readable string.
+func (o *OptNTPServers) String() string {
+	var ntpServers string
+	for idx, ntp := range o.NTPServers {
+		ntpServers += ntp.String()
+		if idx < len(o.NTPServers)-1 {
+			ntpServers += ", "
+		}
+	}
+	return fmt.Sprintf("NTP Servers -> %v", ntpServers)
```
insomniacslk: what about just `strings.Join(o.NTPServers, ", ")` ?

pmazzini: The problem is that o.NTPServers is not a list of strings.

insomniacslk: right - and not even bytes.Join as this should still need explicit casting

insomniacslk:
Can you look at the merge conflict?

pmazzini:
Rebased, I think it should be good now.

PR #117, opened 2018-08-02
fix OptUserClass

pmazzini:
The ToBytes was always writing a packet in a rfc3004 compliant way while the FromBytes was parsing a user class option either following the rfc3004 or the Microsoft standard.

This will make it more consistent with what we read and write.


dhcpv4/option_userclass.go
```diff
@@ -32,6 +36,9 @@ func (op *OptUserClass) ToBytes() []byte {
 // Length returns the option length
 func (op *OptUserClass) Length() int {
 	ret := 0
+	if !op.Rfc3004 && len(op.UserClasses) == 1 {
```
insomniacslk: the `String` method should be modified as well to only print the first user class only

pmazzini: If more than 1 user class is present, I am assuming rfc3004 is being used so printing all of them should be fine.

PR #119, opened 2018-08-02
add OptClientArchType

pmazzini:


dhcpv4/option_archtype.go
```diff
@@ -0,0 +1,109 @@
+package dhcpv4
+
+// This option implements the Client System Architecture Type option
+// https://tools.ietf.org/html/rfc4578
```
insomniacslk: the constants below seem a duplicate of what's in the dhcpv6 package, however it's probably not worth keeping them in a separate package, we can tolerate some duplication here as this stuff won't change. Can you just add a comment here to ask to keep this in sync with the dhcpv6 counterpart in case of changes?

pmazzini: Sounds good.

pmazzini: I will add it on a separate PR since this is already merged.

pmazzini: working on it in #125

PR #122, opened 2018-08-03
add IsOptionRequested

pmazzini:
* add IsOptionRequested
* move functions from utils.go to dhcv6.go


insomniacslk:
as discussed in person, let's merge these functions from `utils.go` in the relevant `dhcpv4.go` and `dhcpv6.go` , since they operate on those object (I'm really not a fan of names like "utils" or "tools" that collect unrelated stuff)

dhcpv4/utils.go
```diff
@@ -0,0 +1,14 @@
+package dhcpv4
+
+// IsRequested function takes a DHCPv4 message and an OptionCode, and returns
+// true if that option is within the requested options of the DHCPv4 message.
+func IsRequested(pkt *DHCPv4, requested OptionCode) bool {
```
insomniacslk: we should call this function with a more obvious (despite longer) name. What about `IsOptionRequested` ?

insomniacslk: also wondering if this shouldn't be a method of `DHCPv4`

dhcpv4/utils.go
```diff
@@ -0,0 +1,14 @@
...
+	for _, optprl := range pkt.GetOption(OptionParameterRequestList) {
```
insomniacslk: in general we don't validate if more than one option of the same type is found in a packet. Also most DHCP implementations stop looking after the first option of a given type. 
Shall we make this more opinionated and only look in the first option only? I.e. using `GetOneOption` instead of `GetOption`

dhcpv4/utils_test.go
```diff
@@ -0,0 +1,17 @@
+package dhcpv4
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsRequetsed(t *testing.T) {
```
insomniacslk: typo Requetsed -> Requested

dhcpv6/utils.go
```diff
@@ -11,16 +11,25 @@ import (
...
+	}
+	if optbf := msg.GetOneOption(OptionBootfileURL); optbf != nil {
+		return true
+	}
+	return false
+}
+
+// IsRequested function takes a DHCPv6 message and an OptionCode, and returns
+// true if that option is within the requested options of the DHCPv6 message.
+func IsRequested(msg DHCPv6, requested OptionCode) bool {
```
insomniacslk: same comments as above

dhcpv4/utils.go
pmazzini: Sounds good, I'll rename it `IsOptionRequested` and make it a method of `DHCPv4`.

dhcpv4/utils.go
pmazzini: For v6 I have seen clients sending more than 1 requested option, each one requesting different things. I'll rather check all of them.

insomniacslk: oh wow. OK. And the RFC doesn't make it easy either:

>   Unless otherwise noted, each option may appear only in the options
>   area of a DHCP message and may appear only once.  If an option does
>   appear multiple times, each instance is considered separate and the
>   data areas of the options MUST NOT be concatenated or otherwise
>   combined.

dhcpv4/dhcpv4.go
```diff
@@ -682,6 +682,20 @@ func (d *DHCPv4) ValidateOptions() {
 	}
 }
 
+// IsOptionRequested function takes a DHCPv4 message and an OptionCode, and
```
insomniacslk: the comment should be slightly updated - DHCPv4 is not anymore an input parameter, so just say "IsOptionRequested takes an OptionCode" etc, or even simpler, "IsOptionRequested returns whether the provided option code is found in the requested options of the DHCPv4 message"

dhcpv6/dhcpv6.go
```diff
@@ -199,3 +201,84 @@ func EncapsulateRelay(d DHCPv6, mType MessageType, linkAddr, peerAddr net.IP) (D
 	outer.AddOption(&orm)
 	return &outer, nil
 }
+
+// IsNetboot function takes a DHCPv6 message and returns true if the machine
+// is trying to netboot. It checks if "boot file" is one of the requested
+// options, which is useful for SOLICIT/REQUEST packet types, it also checks
+// if the "boot file" option is included in the packet, which is useful for
+// ADVERTISE/REPLY packet.
+func IsNetboot(msg DHCPv6) bool {
```
insomniacslk: similarly to the above, what about making `IsNetboot` a method of `DHCPv6` ?

dhcpv6/dhcpv6.go
```diff
@@ -199,3 +201,84 @@ func EncapsulateRelay(d DHCPv6, mType MessageType, linkAddr, peerAddr net.IP) (D
...
+	if optbf := msg.GetOneOption(OptionBootfileURL); optbf != nil {
+		return true
+	}
+	return false
+}
+
+// IsOptionRequested function takes a DHCPv6 message and an OptionCode, and
+// returns true if that option is within the requested options of the DHCPv6
+// message.
+func IsOptionRequested(msg DHCPv6, requested OptionCode) bool {
```
insomniacslk: same comment as above for the DHCPv4 version

dhcpv4/dhcpv4.go
pmazzini: Right. I am going to update it.

dhcpv6/dhcpv6.go
pmazzini: I am not sure. Should we add it to the `DHCPv6` interface? As a method of `DHCPv6Message`? Or leave it as a function.

insomniacslk:
>> func IsNetboot(msg DHCPv6) bool { 


> I am not sure. Should we add it to the DHCPv6 interface? As a method of DHCPv6Message ? Or leave it as a function. 
If anything, I think we should add it as a method of DHCPv6Message, because it only applies to it, not to a generic DHCPv6 message (e.g. it would fail on a relay msg)

PR #124, opened 2018-08-10
Added modifier for netbootv4

mooneyow:
Also added an rfc parameter for WithUserClass


dhcpv4/modifiers.go
```diff
@@ -1,11 +1,47 @@
 package dhcpv4
 
 // WithUserClass adds a user class option to the packet
-func WithUserClass(uc []byte) Modifier {
+func WithUserClass(uc []byte, rfc bool) Modifier {
```
insomniacslk: can you update the function comment to explain what the `rfc` bool is for?

PR #125, opened 2018-08-12
fix OptClientArchType

### LINKED ISSUE #120: dhcpv6 OptClientArchType not following RFC
rfc5970 specifies that multiple client arch types can be specified.

> The client MAY use this option to send a list of supported
> architecture types to the server

The library is assuming only one is specified.

pmazzini:
* support multiple arch types (fixes #120)
* move arch types to iana package


insomniacslk:
needs to resolve conflicts before merging

PR #126, opened 2018-08-13
Update CONTRIBUTORS.md

insomniacslk:
Added more contributors


insomniacslk:
The list is sorted by number of commits, using `git shortlog -sn`

PR #130, opened 2018-08-13
Adding module for DHCPv6 Vendor Options (Opt 17)

gorhamc:
Adding option code for DHVPv6 Option 17 based on https://tools.ietf.org/html/rfc3315#section-22.17

I'm not modifying the options.go ParseOptions function just yet. The vendor I need this for isn't currently following the RFC apparently. I will add this option into ParseOptions when they abide by RFC, hopefully very soon. For now this is reserved for future use.

This was originally attempted in PR#110 but I messed it up because I'm a github noob.


dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"encoding/binary"
+	"fmt"
+)
+
+// OptVendorOpts represents a DHCPv6 Status Code option
+type OptVendorOpts struct {
+	enterpriseNumber uint32
+	vendorOpts       []byte
```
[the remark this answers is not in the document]
gorhamc: I will do some testing and see if this works with what the vendor is sending us. I don't think they are sending the data as additional options in that field.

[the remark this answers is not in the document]
insomniacslk: feel free to ping me privately for sake of speed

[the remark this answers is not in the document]
pmazzini: > I've been testing with the GenericOption since the last comment here but now it seems to be grabbing too many bytes by default for what the vendor is now sending me.

Are all of those bytes part of the vendor options?

> Would you recommend I request the vendor to change the option 17 data they are sending me to better match the RFC

Yes, ideally they should be following the RFC.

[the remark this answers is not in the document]
pmazzini: > according to the RFC this is a list of DHCP option fields, so this can become []Option. This will also help enforce the correct format and size

I had a look at this with @gorhamc. As per RFC I think we should treat them as opaque data instead of `[]Option`. Same thing that we are doing for option 16.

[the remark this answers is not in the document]
pmazzini: I spoke to soon. They have the same format as an `Option` but the codes do not match the ones in iana so maybe we can treat all of them as generic options.

pmazzini:
There are commits that don't belong to this PR. Can those commits be removed? I think that with git rebase -i you can leave out the commits that you don't need.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+			+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+			.                                                               .
+			.                          option-data                          .
+			.                                                               .
+			+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+*/
+
+import (
+	"encoding/binary"
+  "errors"
```
pmazzini: indentation needs to be fixed in a few places

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// String returns a string representation of the VendorOpts data
+func (op *OptVendorOpts) String() string {
+	return fmt.Sprintf("OptVendorOpts{enterprisenum=%v, vendorOpts=%v}",
+		op.EnterpriseNumber, op.VendorOpts,
+	)
+}
+
+// Build an OptVendorOpts structure from a sequence of bytes.
```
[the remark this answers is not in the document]
pmazzini: Add a comment starting with `ParseOptVendorOpts` to make lint happy.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"encoding/binary"
+  "errors"
+	"fmt"
+)
+
+// OptVendorOpts represents a DHCPv6 Status Code option
+type OptVendorOpts struct {
+	EnterpriseNumber uint32
+	VendorOpts       []OptionGeneric
```
pmazzini: How about having `[]Option` instead? We can still append all references to `OptionGeneric`.
And may be able to define some specific vendor options in the future.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+
+				Option 17
+       0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      |      OPTION_VENDOR_OPTS       |           option-len          |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      |                       enterprise-number                       |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      .                                                               .
+      .                   option-data (sub-options)         					.
```
pmazzini: nit: there is a dot with bad indentation at the right

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+	var err error
+	opt.VendorOpts, err = VendorOptionsFromBytes(data[4:])
+	if err != nil {
+		return nil, err
+	}
+	return &opt, nil
+}
+
+// VendorOptionsFromBytes builds a slice of GenericOptions from a slice of bytes
+func VendorOptionsFromBytes(data []byte) ([]Option, error) {
```
pmazzini: Is it useful to export this? Otherwise I would make it private by starting the function with a lowercase letter.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+		idx += opt.Length() + 4 // 4 bytes for type + length
+	}
+	return options, nil
+}
+
+// VendParseOption builds a GenericOption from a slice of bytes
+// We cannot use the exisitng ParseOption function in options.go because the
+// sub-options include codes specific to each vendor. There are overlaps in these
+// codes with RFC standard codes.
+func VendParseOption(dataStart []byte) (Option, error) {
```
pmazzini: Is it useful to export this? Otherwise I would make it private by starting the function with a lowercase letter.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+	// Parse a sequence of bytes as a single DHCPv6 option.
+	// Returns the option structure, or an error if any.
+	opt := &OptionGeneric{}
+
+	if len(dataStart) < 4 {
+		return opt, fmt.Errorf("Invalid DHCPv6 option: less than 4 bytes")
```
pmazzini: I would specify this is a vendor option: "Invalid DHCPv6 vendor option"

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	code := OptionCode(binary.BigEndian.Uint16(dataStart[:2]))
+	length := int(binary.BigEndian.Uint16(dataStart[2:4]))
+	if len(dataStart) < length+4 {
+		return opt, fmt.Errorf("Invalid option length for option %v. Declared %v, actual %v",
```
pmazzini: I would specify this is a vendor option: "Invalid option length for vendor option"

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	opt.OptionData = data
+	opt.OptionCode = code
+
+	if len(opt.OptionData) < 1 {
+		return opt, errors.New("ParseOptVendorOpts: at least one vendor options data is required")
+	}
+
+	if length != opt.Length() {
+		return opt, fmt.Errorf("Error: declared length is different from actual length for option %d: %d != %d",
```
pmazzini: I would specify this is a vendor option: "Error: declared length is different from actual length for vendor option"

dhcpv6/options.go
```diff
@@ -106,6 +106,8 @@ func ParseOption(dataStart []byte) (Option, error) {
 		opt, err = ParseOptClientArchType(optData)
 	case OptionNII:
 		opt, err = ParseOptNetworkInterfaceId(optData)
+	case OptionVendorOpts:
```
pmazzini: I would sort this by option number instead of adding to the end. How about poting this after OptionVendorClass.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, err
+	}
+	return &opt, nil
+}
+
+// VendorOptionsFromBytes builds a slice of GenericOptions from a slice of bytes
+func VendorOptionsFromBytes(data []byte) ([]Option, error) {
+	// Parse a sequence of bytes until the end and build a list of options from
+	// it. Returns an error if any invalid option or length is found.
+	options := make([]Option, 0, 10)
```
insomniacslk: I imagine that 10 is an arbitrary capacity, maybe too high. Do you know from experience what's a common number of options in a vendor opt? We may tune around that instead, or it could be even 0 since this doesn't seem to be a performance offender

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+		// no options, no party
+		return options, nil
+	}
+	if len(data) < 4 {
+		// cannot be shorter than option code (2 bytes) + length (2 bytes)
+		return nil, fmt.Errorf("Invalid options: shorter than 4 bytes")
+	}
+	idx := 0
+	for {
+		if idx == len(data) {
```
insomniacslk: this should be `>=` because we blindly increase idx below with whatever opt.Length says

dhcpv6/option_vendor_opts.go
insomniacslk: Isn't this the same as `OptionsFromBytes`?

https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/options.go#L122

pmazzini: Almost. Vendor options don't follow the iana codes, so for example we can have a vendor option number 1 which doesn't follow the structure of the Client ID option. That is why we can't use `OptionsFromBytes`.

dhcpv6/option_vendor_opts.go
pmazzini: I think this is fine. If the length of what we processed and the length of the data matches we return no error. Otherwise we return an error.

pmazzini: Same logic as in https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/options.go#L136

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,161 @@
...
+
+	data := dataStart[4 : 4+length]
+	if len(data) < 2 {
+		return opt, errors.New("ParseOptVendorOpts: short data: missing length field")
+	}
+	opt.OptionData = data
+	opt.OptionCode = code
+
+	if len(opt.OptionData) < 1 {
+		return opt, errors.New("ParseOptVendorOpts: at least one vendor options data is required")
```
insomniacslk: The function name should be changed from the error message, it's not `ParseOptVendorOpts` (I guess a leftover from refactoring)

dhcpv6/option_vendor_opts.go
insomniacslk: So the difference would be only that `OptionsFromBytes` calls internally `ParseOption`, but the latter is not what we need?
If so, we can generalize it as follows:
* `OptionsFromBytes` is split into `OptionsFromBytesWithParser` (or better name) that takes a parser function,. and the original `OptionsFromBytes` just calls the new function passing `ParseOption` as function pointer
* Instead of calling `VendorParseOption`, it will call `OptionsFromBytesWithParser` with a custom anonymous function that simply assigns `opt.OptionCode` and `opt.OptionData`.
How does that sound?

pmazzini: Sounds good. We will need to define a new type for the parser function, won't we?

insomniacslk: Yes. Even if it sounds overcomplicated, I think it's still better than duplicating and testing that code again. We can see how it comes and decide

pmazzini: To add some light into it, if I am not mistaken, I think we should define something like:
```
type Parser func(dataStart []byte) (Option, error)
```

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,129 @@
...
+
+				Option 17
+       0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      |      OPTION_VENDOR_OPTS       |           option-len          |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      |                       enterprise-number                       |
+      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+      .                                                               .
+      .                   option-data (sub-options)         	       .
```
pmazzini: nit: the dot in the right doesn't look aligned

pmazzini:
Overall it looks good to me.

pmazzini:
Looks good to me. Waiting for @insomniacslk's review.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,129 @@
...
+// We cannot use the exisitng ParseOption function in options.go because the
+// sub-options include codes specific to each vendor. There are overlaps in these
+// codes with RFC standard codes.
+func vendParseOption(dataStart []byte) (Option, error) {
+	// Parse a sequence of bytes as a single DHCPv6 option.
+	// Returns the option structure, or an error if any.
+	opt := &OptionGeneric{}
+
+	if len(dataStart) < 4 {
+		return opt, fmt.Errorf("Invalid DHCPv6 vendor option: less than 4 bytes")
```
pmazzini: I think this should return nil instead of opt.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	code := OptionCode(binary.BigEndian.Uint16(dataStart[:2]))
+	length := int(binary.BigEndian.Uint16(dataStart[2:4]))
+	if len(dataStart) < length+4 {
+		return opt, fmt.Errorf("Invalid option length for vendor option %v. Declared %v, actual %v",
```
pmazzini: I think this should return nil instead of opt.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	opt.OptionData = data
+	opt.OptionCode = code
+
+	if len(opt.OptionData) < 1 {
+		return opt, errors.New("vendParseOption: at least one vendor options data is required")
+	}
+
+	if length != opt.Length() {
+		return opt, fmt.Errorf("Error: declared length is different from actual length for vendor option %d: %d != %d",
```
pmazzini: I think this should return nil instead of opt.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,129 @@
...
+
+	data := dataStart[4 : 4+length]
+	if len(data) < 2 {
+		return opt, errors.New("vendParseOption: short data: missing length field")
+	}
+	opt.OptionData = data
+	opt.OptionCode = code
+
+	if len(opt.OptionData) < 1 {
+		return opt, errors.New("vendParseOption: at least one vendor options data is required")
```
pmazzini: I think this should return nil instead of opt.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	if length != opt.Length() {
+		return opt, fmt.Errorf("Error: declared length is different from actual length for vendor option %d: %d != %d",
+			code, opt.Length(), length)
+	}
+
+	return opt, nil
```
pmazzini: How about doing `return &OptionGeneric{OptionCode: code, OptionData: optData}, nil` and removing lines 99, 116 and 117.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,126 @@
...
+	code := OptionCode(binary.BigEndian.Uint16(dataStart[:2]))
+	length := int(binary.BigEndian.Uint16(dataStart[2:4]))
+	if len(dataStart) < length+4 {
+		return nil, fmt.Errorf("Invalid option length for vendor option %v. Declared %v, actual %v",
+			code, length, len(dataStart)-4,
+		)
+	}
+
+	optData := dataStart[4 : 4+length]
+	if len(optData) < 2 {
```
pmazzini: Why are we checking for 2 bytes here? I think this is not needed.

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, errors.New("vendParseOption: short data: missing length field")
+	}
+
+	if len(optData) < 1 {
+		return nil, errors.New("vendParseOption: at least one vendor options data is required")
+	}
+
+	if length != len(optData) {
```
pmazzini: We do create the optData based on length. I don't think this check is needed, is it?

dhcpv6/option_vendor_opts.go
gorhamc: you're right. I don't think this is needed here.

dhcpv6/option_vendor_opts.go
gorhamc: you're right. Length is already checked at Line 105
`
if len(dataStart) < length+4 {
		return nil, fmt.Errorf("Invalid option length for vendor option %v. Declared %v, actual %v", code, length, len(dataStart)-4,
		)
	}`

insomniacslk:
still in a meeting, I'll be able to review in about 30 minutes

dhcpv6/option_vendor_opts.go
```diff
@@ -0,0 +1,117 @@
...
+		buf = append(buf, opt.ToBytes()...)
+	}
+	return buf
+}
+
+// Length returns the option length
+func (op *OptVendorOpts) Length() int {
+	l := 4 // 4 bytes for Enterprise Number
+	for _, opt := range op.VendorOpts {
+		l += 4 + opt.Length() // 4 bytes for Code and Length from Vendor
```
insomniacslk: the comment should be on a line on its own as per Go formatting rules. I'm surprised that Stickler didn't catch it!

insomniacslk:
duh, codecov is not too happy, let me check what's wrong

insomniacslk:
it's missing coverage for short packets in `ParseOptVendorOpts` and `vendParseOption` (all the code branches where those functions return an error). It would be nice to get those tested as well if you can

insomniacslk:
Thanks @gorhamc and @pmazzini !

PR #134, opened 2018-08-14
Async DHCPv4 client

### LINKED ISSUE #89: DHCPv4 async client
Similar to the dhcpv6 we may want to add the v4 asynchronous client

earlgreyz:
A prototype of DHCPv4 Async client. Fixes #89.

### DIFF
--- a/dhcpv4/async/client.go
+++ b/dhcpv4/async/client.go
@@ -1,0 +1,208 @@
+package async
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"sync"
+	"time"
+
+	"github.com/fanliao/go-promise"
+	"github.com/insomniacslk/dhcp/dhcpv4"
+)
+
+// Default ports
+const (
+	DefaultServerPort = 67
+	DefaultClientPort = 68
+)
+
+// Client implements an asynchronous DHCPv4 client
+// It doesn't use the broadcast socket! Which means it should be used only when
+// the network is already established.
+// https://github.com/insomniacslk/dhcp/issues/143
+type Client struct {
+	ReadTimeout  time.Duration
+	WriteTimeout time.Duration
+	LocalAddr    net.Addr
+	RemoteAddr   net.Addr
+	IgnoreErrors bool
+
+	connection   *net.UDPConn
+	cancel       context.CancelFunc
+	stopping     *sync.WaitGroup
+	receiveQueue chan *dhcpv4.DHCPv4
+	sendQueue    chan *dhcpv4.DHCPv4
+	packetsLock  sync.Mutex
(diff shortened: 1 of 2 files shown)


earlgreyz:
The implemented client is for purpose of unicast dhcp requests. Won't work with broadcast but in that case we probably don't need async client anyway.

dhcpv4/async/client.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+// Default ports
+const (
+	DefaultServerPort = 67
+	DefaultClientPort = 68
+)
+
+// Client implements an asynchronous DHCPv4 client
+// It doesn't use the broadcast socket! Which means it should be used only when
```
insomniacslk: can you open an issue for the broadcast implementation, and link it here? I think it's something we may still want in the future

dhcpv4/async/client.go
```diff
@@ -0,0 +1,206 @@
...
+
+// NewClient creates an asynchronous client
+func NewClient() *Client {
+	return &Client{
+		ReadTimeout:  dhcpv4.DefaultReadTimeout,
+		WriteTimeout: dhcpv4.DefaultWriteTimeout,
+	}
+}
+
+// Open starts the client
```
insomniacslk: can you add a very brief mention on the meaning of buffer size?

dhcpv4/async/client.go
```diff
@@ -0,0 +1,206 @@
...
+
+	close(c.sendQueue)
+	close(c.receiveQueue)
+	close(c.errors)
+
+	c.connection.Close()
+}
+
+// Errors returns a channel where runtime errors are posted
+func (c *Client) Errors() <-chan error {
```
insomniacslk: would it make sense to associate errors to transaction IDs? If so, fee free to make a separate PR or open an issue for that

dhcpv4/async/client.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (c *Client) receive(_ *dhcpv4.DHCPv4) {
+	var (
+		oobdata  = []byte{}
+		received *dhcpv4.DHCPv4
+	)
+
+	c.connection.SetReadDeadline(time.Now().Add(c.ReadTimeout))
+	for {
```
insomniacslk: probably worth deduplicating this code from the synchronous client

dhcpv4/async/client.go
earlgreyz: The errors associated with transaction IDs are not posted to this channel, they are `Rejected` in a promise for a given request. These are the errors we were unable to associate with any of the known promises.

PR #135, opened 2018-08-14
NewRelayReplFromRelayForw: add remoteid

pmazzini:
Add the remote id option when creating a relay reply.

According to https://tools.ietf.org/html/rfc4649#section-5 this shouldn't be needed:

> There is no requirement that a server return this option and its data
> in a RELAY-REPLY message.

However, there are some relay agents that expect this data so I am adding it to the reply.

cc @gorhamc

### DIFF
--- a/dhcpv6/dhcpv6relay.go
+++ b/dhcpv6/dhcpv6relay.go
@@ -170,14 +170,14 @@ func NewRelayReplFromRelayForw(relayForw, msg DHCPv6) (DHCPv6, error) {
 		optrid             []Option
 	)
 	if relayForw == nil {
-		return nil, errors.New("RELAY_FORW cannot be nil")
+		return nil, errors.New("MessageTypeRelayForward cannot be nil")
 	}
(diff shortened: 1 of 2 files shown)


pmazzini:
I'll increase the test coverage.

dhcpv6/dhcpv6relay.go
```diff
@@ -158,23 +158,26 @@ func (d *DHCPv6Relay) GetInnerMessage() (DHCPv6, error) {
...
 	var (
 		err                error
 		linkAddr, peerAddr []net.IP
-		optiids            []Option
+		optiid             []Option
+		optrid             []Option
 	)
 	if relayForw == nil {
-		return nil, errors.New("RELAY_FORW cannot be nil")
+		return nil, errors.New("MessageTypeRelayForward cannot be nil")
```
insomniacslk: to be more accurate, the message here could be "Relay message cannot be nil"

PR #140, opened 2018-08-19
Added String methods for types

insomniacslk:
I've added a String method to all the option/message types so we don't need to check the map manually every time, or to use a function (a method is more intuitive).

Only one exception: bsdp.OptionCode cannot be easily converted to this approach unless we define it as its own type


insomniacslk:
this should not break any interface. The only place that could have broken it is where I marked the function as deprecated. That can be removed as soon as we will have versioned modules and releases

PR #149, opened 2018-08-24
DHCPv4 relay modifier

earlgreyz:
Modifier which sets the parameters of DHCPv4 packet to relay.

### DIFF
--- a/dhcpv4/modifiers.go
+++ b/dhcpv4/modifiers.go
@@ -1,5 +1,9 @@
 package dhcpv4
 
+import (
+	"net"
+)
+
 // WithUserClass adds a user class option to the packet.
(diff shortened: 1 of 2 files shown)


dhcpv4/modifiers.go
```diff
@@ -63,3 +63,14 @@ func WithRequestedOptions(optionCodes ...OptionCode) Modifier {
 		return d
 	}
 }
+
+// WithRelay adds parameters required for DHCPv4 to be relayed by the relay
+// server with given ip
+func WithRelay(ip net.IP) Modifier {
```
insomniacslk: missing `net` import, TravisCI is failing

PR #155, opened 2018-09-20
Add NoPrefixAvail status code

davebarrau:
As per https://www.ietf.org/rfc/rfc3633.txt section 16, status code 6 has been reserved for "Delegating router has no prefixes available to assign to the IAPD(s)"


insomniacslk:
Thanks for your contribution! Looks good to me, I'll wait for tests to finish before merging

PR #161, opened 2018-10-01
BSDP: Add MessageTypeFromPacket helper

get9:
This is a helper function to extract the BSDP message type from a given
`dhcpv4.DHCPv4` packet structure. It's useful in helping to identify
what kind of BSDP packet you're dealing with (since the distinction is
only made in one of the vendor-specific options encapsulated in Option
43).

### DIFF
--- a/dhcpv4/bsdp/bsdp.go
+++ b/dhcpv4/bsdp/bsdp.go
@@ -51,6 +51,26 @@ func needsReplyPort(replyPort uint16) bool {
 	return replyPort != 0 && replyPort != dhcpv4.ClientPort
 }
 
+func MessageTypeFromPacket(packet *dhcpv4.DHCPv4) (MessageType, bool) {
+	var (
+		messageType MessageType
+		vendorOpts  *OptVendorSpecificInformation
+		err         error
+	)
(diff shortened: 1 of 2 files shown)


dhcpv4/bsdp/bsdp.go
```diff
@@ -51,6 +51,29 @@ func needsReplyPort(replyPort uint16) bool {
 	return replyPort != 0 && replyPort != dhcpv4.ClientPort
 }
 
+// MessageTypeFromPacket extracts the BSDP message type (LIST, SELECT) from the
+// vendor-specific options and returns it. If the message type option cannot be
+// found, returns false.
+func MessageTypeFromPacket(packet *dhcpv4.DHCPv4) (MessageType, bool) {
```
insomniacslk: I think this function could have returned just a `*MessageType` instead of `(MessageType, bool)`. If the return value is nil, it signals that the function could not find the message type.
Adding a second return value (the boolean) makes the function more verbose (e.g. the return value cannot be passed directly to another function but requires a check on the boolean), and the object is passed by copy in any case, either that it succeeds or fails

get9: Good point, fixed with #162

PR #163, opened 2018-10-03
dhcpv4: Add OptRootPath

get9:
Adds a root path option that supports DHCPv4 option 17

### DIFF
--- a/dhcpv4/option_root_path.go
+++ b/dhcpv4/option_root_path.go
@@ -1,0 +1,52 @@
+package dhcpv4
+
+import (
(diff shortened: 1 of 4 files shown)


pmazzini:
Should this be added to the ParseOption function in options.go?

insomniacslk:
> Should this be added to the ParseOption function in options.go?

Yes, otherwise the option will not be parsed automatically. I hate the fact that we have to register the decoder manually though. We have to stick with it for now but any idea to make this less manual is welcome

PR #164, opened 2018-10-04
DHCPv4: Add sendto address

get9:
Sometimes you may want to send a packet to an address other than e.g.
the gateway address stored inside the packet. In these instances, you
need another field in which to store the address. The `SendToIPAddr`
field in the DHCPv4 struct is nominally `nil`, but in these rare
occasions where you do want to send another address, consumer code can
set the address.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -32,6 +32,7 @@ type DHCPv4 struct {
 	yourIPAddr     net.IP
 	serverIPAddr   net.IP
 	gatewayIPAddr  net.IP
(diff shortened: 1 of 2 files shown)


pmazzini:
I wouldn't do this because that field is not specified in the RFC:
https://tools.ietf.org/html/rfc2131#section-2

I would only store in that struct the data that is sent over the wire.

Moreover, I think the ToBytes and FromBytes would differ which is not desirable.

insomniacslk:
I agree with @pmazzini . I already pinged @get9 privately to sort out what what would be the use case, but better to talk here

PR #165, opened 2018-10-04
[WIP] Added DHCPv4 server stub

insomniacslk:
Added code to run a DHCPv4 server.

NOTE: this code won't work until the PR to do unicast DHCPv4 is submitted. Which should happen soon.

### DIFF
--- a/dhcpv4/server.go
+++ b/dhcpv4/server.go
@@ -58,7 +58,7 @@ func main() {
 // valid DHCPv4 message is received
 type Handler func(conn net.PacketConn, peer net.Addr, m *DHCPv4)
 
-// Server represents a DHCPv6 server object
+// Server represents a DHCPv4 server object
 type Server struct {
 	conn       net.PacketConn
 	connMutex  sync.Mutex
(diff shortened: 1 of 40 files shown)


dhcpv4/server.go
```diff
@@ -0,0 +1,154 @@
...
+	defer s.connMutex.Unlock()
+	if s.conn == nil {
+		return nil
+	}
+	return s.conn.LocalAddr()
+}
+
+// ActivateAndServe starts the DHCPv4 server
+func (s *Server) ActivateAndServe() error {
+	s.connMutex.Lock()
```
get9: why not defer the unlock here?

insomniacslk: because I need it to be unlocked as soon as possible (see after the check on `err`), but also unlock it if something fails (see inside the `err` check)

insomniacslk:
Thanks! This is actually pending https://github.com/insomniacslk/dhcp/pull/172

insomniacslk:
I messed up, dropping this and recreating

PR #166, opened 2018-10-06
Add support for DHCPv6-PD sub-options

davebarrau:
Thoughts on this? I tried to make it look the same as OptionIANA from an options perspective

- Allow for parsed sub-options in OptIAForPrefixDelegation
- Add initial test

### DIFF
--- a/dhcpv6/option_prefixdelegation.go
+++ b/dhcpv6/option_prefixdelegation.go
@@ -12,7 +12,7 @@ type OptIAForPrefixDelegation struct {
 	iaId    [4]byte
 	t1      uint32
 	t2      uint32
-	options []byte
+	Options []Option
 }
 
 func (op *OptIAForPrefixDelegation) Code() OptionCode {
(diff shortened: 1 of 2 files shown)


pmazzini:
Thanks for your contribution! It looks good to me.

The only problem is that this will break router7 build cc @insomniacslk:
https://github.com/rtr7/router7/blob/master/internal/dhcp6/dhcp6.go#L285

pmazzini:
Nice, thanks for fixing it. I would also add a deprecated message to the Options() method so that it can be removed later. Similar to:
https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/types.go#L45

davebarrau:
> Nice, thanks for fixing it. I would also add a deprecated message to the Options() method so that it can be removed later. Similar to:
> https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/types.go#L45

Would Options() be removed, or be changed to returning []Option as part of the deprecation? If the function is removed, there's no easy way to get all options otherwise, unless a differently named function is added as part of this pull request.

pmazzini:
For consistency I think it would be removed and the public field can be used instead, similar to OptionIANA.

dhcpv6/option_prefixdelegation.go
```diff
@@ -12,75 +12,114 @@ type OptIAForPrefixDelegation struct {
 	iaId    [4]byte
 	t1      uint32
 	t2      uint32
-	options []byte
+	options []Option
```
pmazzini: I would use capital letter to make this field public similar to OptIANA.

davebarrau: Ah, like how I originally had it, of course. :)

insomniacslk:
sorry, I was on a long flight and I'm seeing this only now, I'm looking at the code now

PR #167, opened 2018-10-07
Dynamically determine loopback interface for server test

davebarrau:
"lo" isn't configured everywhere. In FreeBSD for example, TestServerActivateAndServe() fails as the loopback address is "lo0", not "lo".

Also, I don't believe the Zone is required for net.UDPAddr as ::1 is being used, not an fe80:: address. On FreeBSD, since the "lo" Zone doesn't exist, it can't be mapped to an interface index and seems to be silently dropped (I can put "asdf" and it does the same thing). The tests still work fine though.

### DIFF
--- a/dhcpv6/server_test.go
+++ b/dhcpv6/server_test.go
@@ -9,13 +9,29 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// utility function to return the loopback interface name
+func getLoopbackInterface() string {
+        var ifaces []net.Interface
+        var err error
+        if ifaces, err = net.Interfaces(); err != nil {
+                return ""
+        }
+
+        for _, iface := range ifaces {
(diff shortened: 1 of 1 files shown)


dhcpv6/server_test.go
```diff
@@ -9,21 +9,36 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// utility function to return the loopback interface name
+func getLoopbackInterface() string {
+        var ifaces []net.Interface
+        var err error
+        if ifaces, err = net.Interfaces(); err != nil {
```
insomniacslk: I prefer to return an error here, so it is obvious what happened, i.e. missing loopback interface? Or any other error?

davebarrau: Added. :)

PR #170, opened 2018-10-09
Avoid panic on malformed OptIAPrefix

davebarrau:
Noticed the length check on ParseOptIAPrefix() is wrong. The other DHCPv6 options seem okay.


insomniacslk:
Good catch, thank you!

PR #172, opened 2018-10-11
Custom client/server addresses/ports for DHCPv4 client

gvialetto:
The current DHCPv4 client does not have the ability to specify remote/local addresses/ports. This small patch adds that.

`BroadcastSendReceive()` and `MakeRawBroadcastPacket()` are not used anymore, I did not remove them (since they're public) for backward compatibility.

### DIFF
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -247,7 +247,7 @@ func (c *Client) sendReceive(sendFd, recvFd int, packet *DHCPv4, messageType Mes
 	if err != nil {
 		return nil, err
 	}
-	packetBytes, err := makeRawPacket(packet.ToBytes(), raddr, laddr)
+	packetBytes, err := MakeRawPacket(packet.ToBytes(), raddr, laddr)
 	if err != nil {
 		return nil, err
 	}
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -50,6 +50,7 @@ func NewClient() *Client {
 // MakeRawBroadcastPacket leverages MakeRawPacket to create a raw packet suitable
 // for UDP broadcast.
 func MakeRawBroadcastPacket(payload []byte) ([]byte, error) {
+	log.Printf("Warning: dhcpv4.MakeRawBroadcastPacket() is deprecated and will be removed.")
 	serverAddr := &net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
 	clientAddr := &net.UDPAddr{IP: net.IPv4zero, Port: ClientPort}
 	return MakeRawPacket(payload, serverAddr, clientAddr)
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -28,16 +28,6 @@ var (
 	DefaultWriteTimeout = 3 * time.Second
 )
 
-// ClientExchangeOptions specifies a set of options to be used for a particular
-// exchange.
-type ClientExchangeOptions struct {
-	// UnicastSocket is used to ask the library to use an unicast socket instead
-	// of a broadcast socket for the DHCPv4 exchange. Note that it's on the
-	// user of the library to set up the actual DHCP packet (ex: by specifying a
-	// server address) so this actually works as expected.
-	UnicastSocket bool
-}
-
 // Client is the object that actually performs the DHCP exchange. It currently
 // only has read and write timeout values, plus (optional) local and remote
 // addresses.
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"time"
+	"fmt"
 
 	"golang.org/x/net/ipv4"
 	"golang.org/x/sys/unix"
@@ -42,8 +43,8 @@ type ClientExchangeOptions struct {
 // addresses.
 type Client struct {
 	ReadTimeout, WriteTimeout time.Duration
-	RemoteAddr   *net.UDPAddr
-	LocalAddr    *net.UDPAddr
+	RemoteAddr   net.Addr
+	LocalAddr    net.Addr
 }
 
 // NewClient generates a new client to perform a DHCP exchange with, setting the
@@ -228,15 +229,27 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier
 // response up to some read timeout value. If the message type is not
 // MessageTypeNone, it will wait for a specific message type
 func (c *Client) SendReceive(sendFd, recvFd int, packet *DHCPv4, messageType MessageType) (*DHCPv4, error) {
-	serverAddr := c.RemoteAddr
-	if serverAddr == nil {
-		serverAddr = &net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	var laddr net.UDPAddr
+	if c.LocalAddr == nil {
+		laddr = net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	} else {
+		if addr, ok := c.LocalAddr.(*net.UDPAddr); ok {
+			laddr = *addr
+		} else {
+			return nil, fmt.Errorf("Invalid local address: not a net.UDPAddr: %v", c.LocalAddr)
+		}
 	}
-	clientAddr := c.LocalAddr
-	if clientAddr == nil {
-		clientAddr = &net.UDPAddr{IP: net.IPv4zero, Port: ClientPort}
+	var raddr net.UDPAddr
+	if c.RemoteAddr == nil {
+		raddr = net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	} else {
+		if addr, ok := c.RemoteAddr.(*net.UDPAddr); ok {
+			raddr = *addr
+		} else {
+			return nil, fmt.Errorf("Invalid remote address: not a net.UDPAddr: %v", c.RemoteAddr)
+		}
 	}
-	packetBytes, err := MakeRawPacket(packet.ToBytes(), serverAddr, clientAddr)
+	packetBytes, err := MakeRawPacket(packet.ToBytes(), &raddr, &laddr)
 	if err != nil {
 		return nil, err
 	}
@@ -247,8 +260,8 @@ func (c *Client) SendReceive(sendFd, recvFd int, packet *DHCPv4, messageType Mes
 		destination [4]byte
 		response    *DHCPv4
 	)
-	copy(destination[:], serverAddr.IP.To4())
-	remoteAddr := unix.SockaddrInet4{Port: clientAddr.Port, Addr: destination}
+	copy(destination[:], raddr.IP.To4())
+	remoteAddr := unix.SockaddrInet4{Port: laddr.Port, Addr: destination}
 	recvErrors := make(chan error, 1)
 	go func(errs chan<- error) {
 		conn, innerErr := net.FileConn(os.NewFile(uintptr(recvFd), ""))
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"time"
 	"fmt"
+	"reflect"
 
 	"golang.org/x/net/ipv4"
 	"golang.org/x/sys/unix"
@@ -153,6 +154,31 @@ func MakeListeningSocket(ifname string) (int, error) {
 	return fd, nil
 }
 
+func (c *Client) getAddresses() (*net.UDPAddr, *net.UDPAddr, error) {
+	var laddr *net.UDPAddr
+	if c.LocalAddr == nil {
+		laddr = &net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	} else {
+		if addr, ok := c.LocalAddr.(*net.UDPAddr); ok {
+			laddr = addr
+		} else {
+			return nil, nil, fmt.Errorf("Invalid local address: only net.UDPAddr is supported, got %v instead", reflect.TypeOf(c.LocalAddr))
+		}
+	}
+	var raddr *net.UDPAddr
+	if c.RemoteAddr == nil {
+		raddr = &net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	} else {
+		if addr, ok := c.RemoteAddr.(*net.UDPAddr); ok {
+			raddr = addr
+		} else {
+			return nil, nil, fmt.Errorf("Invalid remote address: only net.UDPAddr is supported, got %v instead", reflect.TypeOf(c.RemoteAddr))
+		}
+	}
+
+	return raddr, laddr, nil
+}
+
 // Exchange runs a full DORA transaction: Discover, Offer, Request, Acknowledge,
 // over UDP. Does not retry in case of failures. Returns a list of DHCPv4
 // structures representing the exchange. It can contain up to four elements,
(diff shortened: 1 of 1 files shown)


dhcpv4/client.go
```diff
@@ -27,10 +27,23 @@ var (
 	DefaultWriteTimeout = 3 * time.Second
 )
 
+// ClientExchangeOptions specifies a set of options to be used for a particular
+// exchange.
+type ClientExchangeOptions struct {
```
pmazzini: How about removing this and using unicast when the RemoteAddr is != nil?

dhcpv4/client.go
```diff
@@ -27,10 +27,23 @@ var (
...
+	UnicastSocket bool
+}
+
 // Client is the object that actually performs the DHCP exchange. It currently
-// only has read and write timeout values.
+// only has read and write timeout values, plus (optional) local and remote
+// addresses.
 type Client struct {
 	ReadTimeout, WriteTimeout time.Duration
+	RemoteAddr   *net.UDPAddr
```
pmazzini: I would use net.Addr in order to be more consistent with the v6 client.

dhcpv4/client.go
```diff
@@ -118,18 +162,30 @@ func MakeListeningSocket(ifname string) (int, error) {
...
+
+
+// ExchangeWithOptions runs a full DORA transaction: Discover, Offer, Request, Acknowledge,
 // over UDP. Does not retry in case of failures. Returns a list of DHCPv4
 // structures representing the exchange. It can contain up to four elements,
 // ordered as Discovery, Offer, Request and Acknowledge. In case of errors, an
 // error is returned, and the list of DHCPv4 objects will be shorted than 4,
 // containing all the sent and received DHCPv4 messages.
-func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier) ([]*DHCPv4, error) {
+func (c *Client) ExchangeWithOptions(ifname string, discover *DHCPv4, options *ClientExchangeOptions, modifiers ...Modifier) ([]*DHCPv4, error) {
```
pmazzini: I would rename it back to Exchange.

gvialetto: I split this in another function to keep backward compatibility (so the signature of Exchange() does not change). I have no problem to actually add the parameter directly to the function it if we don't care much.

pmazzini: Do we actually need the parameter? I think we can check if we need to use broadcast or unicast by checking if the RemoteAddr is set.

dhcpv4/client.go
gvialetto: Mainly to allow people to actually play with the option - nothing stops you from using a broadcast socket and specify a remote address. It *may* break. Probably we should warn out when we possibly have broken options combined though.

pmazzini: I would keep it simple and infer if we need to use broadcast or unicast based on the RemoteAddr. Let's see what @insomniacslk thinks about it.

insomniacslk: I initially suggested to @gvialetto to leave the freedom to specify what kind of socket they want, since this library is inspired by scapy's principle of letting the user fiddle with the protocol (yet having sane defaults).
On the other side I appreciate simplicity. We could start simple - and on a second thought this is probably a good idea.

However checking for `RemoteAddr != nil` is not enough: to be done properly we have to check if the destination is a broadcast address, and to do that we need the subnet mask

dhcpv4/client.go
insomniacslk: my fault here! I suggested @gvialetto to use UDPAddr because on top of my mind I thought we were using UDPAddr instead of Addr

dhcpv4/client.go
gvialetto: It depends how much freedom we want to give library users. I have no hard feelings either way.

insomniacslk: fair point - I replied already on the other comment thread, basically I'd say: let's start simple as Pablo suggests (and ignore my initial request to create a struct), then we can expand it in a separate PR, to keep things simple. Would that make sense?

gvialetto: It does, I'll update the code 👍

dhcpv4/client.go
pmazzini: Another option is moving this bool to the Client struct. We already setup some configurations options in there such as read and write timeout.

dhcpv4/client.go
```diff
@@ -165,14 +207,110 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier
...
+	case <-time.After(c.ReadTimeout):
+		return nil, errors.New("timed out while listening for replies")
+	}
+
+	return response, nil
+}
+
 // BroadcastSendReceive broadcasts packet (with some write timeout) and waits for a
 // response up to some read timeout value. If the message type is not
 // MessageTypeNone, it will wait for a specific message type
```
pmazzini: Is BroadcastSendReceive being used? If not I would remove it.

dhcpv4/client.go
```diff
@@ -42,12 +46,20 @@ func NewClient() *Client {
 	}
 }
 
-// MakeRawBroadcastPacket converts payload (a serialized DHCPv4 packet) into a
-// raw packet suitable for UDP broadcast.
+// MakeRawBroadcastPacket leverages MakeRawPacket to create a raw packet suitable
+// for UDP broadcast.
 func MakeRawBroadcastPacket(payload []byte) ([]byte, error) {
```
pmazzini: Is this being used? If not I would remove it.

dhcpv4/client.go
```diff
@@ -165,14 +207,110 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier
...
 		return conversation, err
 	}
 	conversation = append(conversation, ack)
 	return conversation, nil
 }
 
+// SendReceive sends a packet (with some write timeout) and waits for a
+// response up to some read timeout value. If the message type is not
+// MessageTypeNone, it will wait for a specific message type
+func (c *Client) SendReceive(sendFd, recvFd int, packet *DHCPv4, messageType MessageType) (*DHCPv4, error) {
```
pmazzini: Maybe it is a good idea to unexport this as in the v6 client so that only Exchange is used.

dhcpv4/client.go
```diff
@@ -165,14 +207,110 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier
...
+	var raddr net.UDPAddr
+	if c.RemoteAddr == nil {
+		raddr = net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	} else {
+		if addr, ok := c.RemoteAddr.(*net.UDPAddr); ok {
+			raddr = *addr
+		} else {
+			return nil, fmt.Errorf("Invalid remote address: not a net.UDPAddr: %v", c.RemoteAddr)
+		}
+	}
```
pmazzini: How about simplifying this with:
`packetBytes, err := MakeRawPacket(packet.ToBytes(), c.RemoteAddr, c.LocalAddr)`

dhcpv4/client.go
insomniacslk: since it's part of the public interface, before unexporting this should have a deprecation warning for a while. We can open an issue to prune deprecated public functions

dhcpv4/client.go
```diff
@@ -165,14 +207,110 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier
...
+// MessageTypeNone, it will wait for a specific message type
+func (c *Client) SendReceive(sendFd, recvFd int, packet *DHCPv4, messageType MessageType) (*DHCPv4, error) {
+	var laddr net.UDPAddr
+	if c.LocalAddr == nil {
+		laddr = net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	} else {
+		if addr, ok := c.LocalAddr.(*net.UDPAddr); ok {
+			laddr = *addr
+		} else {
+			return nil, fmt.Errorf("Invalid local address: not a net.UDPAddr: %v", c.LocalAddr)
```
insomniacslk: DHCPv4 also supports TCP, but we don't (yet?), a more accurate error message would be 
```
Invalid local address, only net.UDPAddr is supported, got %v instead", reflect.TypeOf(c.LocalAddr))
```

dhcpv4/client.go
```diff
@@ -165,14 +207,110 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier
...
+		}
+	}
+	var raddr net.UDPAddr
+	if c.RemoteAddr == nil {
+		raddr = net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	} else {
+		if addr, ok := c.RemoteAddr.(*net.UDPAddr); ok {
+			raddr = *addr
+		} else {
+			return nil, fmt.Errorf("Invalid remote address: not a net.UDPAddr: %v", c.RemoteAddr)
```
insomniacslk: same as above

dhcpv4/client.go
insomniacslk: before removing it, this should be deprecated - grep "deprecated" in the tree to see other similar cases

dhcpv4/client.go
```diff
@@ -128,8 +163,15 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier
 	conversation := make([]*DHCPv4, 0)
 	var err error
 
-	// Get our file descriptor for the broadcast socket.
-	sfd, err := MakeBroadcastSocket(ifname)
+	// Get our file descriptor for the raw socket we need.
+	var sfd int
+	// TODO: we should check the IP is actually not a broadcast address for the
```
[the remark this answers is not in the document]
insomniacslk: handling broadcast addresses properly could be challenging. At this point we can restrict the usage to what's expected by the RFC, and simply use a broadcast socket when the address is `net.IPv4bcast`. Simple, and covering nearly every use case

dhcpv4/client.go
gvialetto: Two reasons:
- we cannot pass the values directly (`net.Addr` vs `*net.UDPAddr`). We need the port in MakeRawPacket, and I'm kinda wary of moving the typecast there.
- default values are set up here if the struct has `nil` values for `RemoteAddr` and `LocalAddr` and we need those default values later in the same function (lines 253-254) anyway.

dhcpv4/client.go
gvialetto: Will do.

dhcpv4/client.go
gvialetto: Will do.

dhcpv4/client.go
```diff
@@ -42,12 +46,20 @@ func NewClient() *Client {
...
-// raw packet suitable for UDP broadcast.
+// MakeRawBroadcastPacket leverages MakeRawPacket to create a raw packet suitable
+// for UDP broadcast.
 func MakeRawBroadcastPacket(payload []byte) ([]byte, error) {
+	serverAddr := &net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	clientAddr := &net.UDPAddr{IP: net.IPv4zero, Port: ClientPort}
+	return MakeRawPacket(payload, serverAddr, clientAddr)
+}
+
+// MakeRawPacket converts a payload (a serialized DHCPv4 packet) into a
```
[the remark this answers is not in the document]
gvialetto: 👍

dhcpv4/client.go
pmazzini: I think this was mostly being used internally but a "deprecated" message sounds ok.

dhcpv4/client.go
pmazzini: How about passing a net.Addr to the MakeRawPacket function. In that way we won't be needing a type assertion and the default ports can be set while creating the raw packet as we are currently doing it.

gvialetto: We'll still need to parse the IP address and convert it to a `net.IP` type to build the `ipv4.Header` instance is used there. 

I need to isolate the type checking in another function (to do the check on the broadcast address @insomniacslk asked below anyway) but I'd rather make `MakeRawPacket` private, so we don't have mismatch in the types exposed in the public interface. 

We could make the function take `srcIP, dstIP: net.IP, srcPort, dstPort int` instead of the current signature. I would rather have that function not "think" about what's in the address (I'm thinking right now that a `net.UnixAddr` can also IIRC be a `net.Addr`) and explicitly pass the parameters in.

dhcpv4/client.go
gvialetto: I will add a log to mark it as deprecated, something like 

```
Warning: dhcpv4.BroadcastSendAndReceive() is deprecated and will be removed. You should use dhcpv4.client.Exchange() instead.
```

dhcpv4/client.go
```diff
@@ -42,12 +46,20 @@ func NewClient() *Client {
...
+// raw packet for the specified serverAddr from the specified clientAddr.
+func MakeRawPacket(payload []byte, serverAddr, clientAddr *net.UDPAddr) ([]byte, error) {
 	udp := make([]byte, 8)
-	binary.BigEndian.PutUint16(udp[:2], ClientPort)
-	binary.BigEndian.PutUint16(udp[2:4], ServerPort)
+	binary.BigEndian.PutUint16(udp[:2], uint16(clientAddr.Port))
```
insomniacslk: this line and the one below will cause a panic if `clientAddr` or `serverAddr` are `nil`.

At the same time I'm not convinced we should use a pointer to `net.UDPAddr`, it feels redundant, since client and server addresses are required fields.

Another option could be to use `net.Addr` here instead (so the whole higher-level code always uses `net.Addr` objects, which is easier to follow) and let the bottom functions, like this one, deal with the details. Being an interface, it would still require a check against `nil`, and the type assertion should be moved here.

dhcpv4/client.go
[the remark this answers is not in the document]
insomniacslk: also please expand this comment to explain why we are keeping it simple as discussed above

dhcpv4/client.go
insomniacslk: I'm for the `net.Addr` approach for the reasons that I've explained above (see comments around `MakeRawPacket`).
I generally agree with not making functions too smart, but in this case *some* function has to be smart. I think the code would be easier to follow if only the bottom functions do this, while the upper level functions maintain a higher level interface. The `net.UnixAddr` issue would be circumvented by the type assertion in the callee.
No strong opinions anyway, if you see problems with this approach let's discuss them here

insomniacslk: > and probably the function should renamed appropriately to be fair

The function deserves renaming - I didn't mention earlier to avoid getting into yak shaving, but since it came up let's do it. `MakeRawUDPPacket`?

> We are still going to get port and address in sendReceive

This is probably the strongest argument we have to do the type assertion here instead of `MakeRawPacket`, let's go for it
 
In summary:
* type assertion here, error if not `net.UDPAddr`
* pass source and destination by copy, not by pointer, as it doesn't impact performances (theoretically it should be even faster because we remove one level of indirection and use short jumps instead of long jumps, but really doesn't matter). Also no need to check for `nil`.

dhcpv4/client.go
```diff
@@ -118,6 +138,31 @@ func MakeListeningSocket(ifname string) (int, error) {
 	return fd, nil
 }
 
+func (c *Client) getUDPAddresses() (*net.UDPAddr, *net.UDPAddr, error) {
```
insomniacslk: nit: this function is large enough to deserve a comment, even if private

dhcpv4/client.go
```diff
@@ -165,18 +221,99 @@ func (c *Client) Exchange(ifname string, discover *DHCPv4, modifiers ...Modifier
...
-	ack, err := BroadcastSendReceive(sfd, rfd, request, c.ReadTimeout, c.WriteTimeout, MessageTypeAck)
+	ack, err := c.sendReceive(sfd, rfd, request, MessageTypeAck)
 	if err != nil {
 		return conversation, err
 	}
 	conversation = append(conversation, ack)
 	return conversation, nil
 }
 
+// SendReceive sends a packet (with some write timeout) and waits for a
```
insomniacslk: the `S` is now lower-case
```suggestion
// sendReceive sends a packet (with some write timeout) and waits for a
```

dhcpv4/client.go
```diff
@@ -118,6 +138,31 @@ func MakeListeningSocket(ifname string) (int, error) {
 	return fd, nil
 }
 
+func (c *Client) getUDPAddresses() (*net.UDPAddr, *net.UDPAddr, error) {
+	var laddr *net.UDPAddr
+	if c.LocalAddr == nil {
+		laddr = &net.UDPAddr{IP: net.IPv4bcast, Port: ServerPort}
+	} else {
```
insomniacslk: I think this function should also check that the IP addresses are a valid IPv4, using `addr.To4()` and checking if not `nil`. This would avoid any risk in the functions that use it.

nit: to avoid duplication it may even be worth to call this function `toUDPAddr(addr net.Addr) (*net.UDPAddr, error)` and call it once for LocalAddr and once for RemoteAddr, but feel free to leave it as it is

gvialetto: Changing it to `toUDPAddr` also makes confusing the returned values impossible, which I like a lot. I'll change this.

gvialetto: I actually made the signature `toUDPAddr(addr net.Addr, defaultAddr *net.UDPAddr)  (*net.UDPAddr, error)` so we can pass a default value to use if `addr` is `nil` (instead of returning `nil` and then having to check for that in the caller).

PR #175, opened 2018-10-15
Added String methods for types

pmazzini:
Added string method for:
* DuidType
* ArchType
* HwTypeType

This makes it more consistent with PR #140.

Also remove deprecated method MessageType() and function MessageTypeToString().

### DIFF
--- a/dhcpv6/duid.go
+++ b/dhcpv6/duid.go
@@ -26,11 +26,10 @@ var DuidTypeToString = map[DuidType]string{
 }
 
 func (d DuidType) String() string {
-	dtype := DuidTypeToString[d]
-	if dtype == "" {
-		dtype = "Unknown"
+	if dtype, ok := DuidTypeToString[d]; ok {
+		return dtype
 	}
-	return dtype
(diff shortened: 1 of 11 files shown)


dhcpv6/duid.go
```diff
@@ -25,6 +25,14 @@ var DuidTypeToString = map[DuidType]string{
 	DUID_UUID: "DUID-UUID",
 }
 
+func (d DuidType) String() string {
+	dtype := DuidTypeToString[d]
```
insomniacslk: this function could be even simpler:
```
if dtype, ok := DuidTypeToString[d]; ok {
    return dtype
}
return "Unknown"
```

However I'm not sure if the default return value here should be "Unknown"

pmazzini: That is what we are currently returning in https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/duid.go#L84

insomniacslk: as discussed in person, let's leave it as it is :)

PR #176, opened 2018-10-15
OptIAForPrefixDelegation: make fields public

pmazzini:
Make the fields of OptIAForPrefixDelegation public. This will make it more consistent with what was done for OptIANA in: https://github.com/insomniacslk/dhcp/commit/bdb0bf25259985b962e230fbac5265a7623afd05

It was not possible to do it in PR #166 because at that point it was going to break router7.

However, this was fixed in https://github.com/rtr7/router7/pull/9.

cc @davebarrau

### DIFF
--- a/dhcpv6/option_prefixdelegation.go
+++ b/dhcpv6/option_prefixdelegation.go
@@ -38,6 +38,7 @@ func (op *OptIAForPrefixDelegation) ToBytes() []byte {
 func (op *OptIAForPrefixDelegation) T1() uint32 {
 	return op.t1
 }
+
 // SetT1 sets the T1 timer for this option
 func (op *OptIAForPrefixDelegation) SetT1(t1 uint32) {
 	op.t1 = t1
@@ -47,6 +48,7 @@ func (op *OptIAForPrefixDelegation) SetT1(t1 uint32) {
 func (op *OptIAForPrefixDelegation) T2() uint32 {
 	return op.t2
 }
+
 // SetT2 sets the T2 timer for this option
 func (op *OptIAForPrefixDelegation) SetT2(t2 uint32) {
 	op.t2 = t2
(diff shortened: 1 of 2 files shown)


pmazzini:
It is still failing router7 build. However this time is because router7 used the T1() method.

insomniacslk:
cc @stapelberg

insomniacslk:
hey @stapelberg , we are making a few changes to the library that would break router7's builds. To avoid breaking your project we can wait for you to update your code on top of our changes, could you please have a look?

pmazzini:
I can work around it by leaving the T1() method until we start using versioning.

insomniacslk:
I guess it's fine to make T1 public too, rather than doing half the work and waiting for versioning (which, by the way, can now be done).

If Michael can make the change on his side in a compatible time-frame, let's just wait, otherwise we can decide for a different next step.

stapelberg:
Thanks for the heads-up. We can certainly make a lockstep change. It’s not entirely clear to me what I’d need to change. Maybe just sending a PR to router7 would be easiest? :)

insomniacslk:
> Maybe just sending a PR to router7 would be easiest? :)

Probably it is, but it will fail initially until the changes aren't merged in Router7. I guess the easiest is to merge this PR and then make another PR to router7.

@pmazzini can you revert the T1/T2 change so they are public? I will make a PR to router7 just after. Thanks!

pmazzini:
> can you revert the T1/T2 change so they are public? I will make a PR to router7 just after. Thanks!

Sure!

insomniacslk:
Merging this PR. As discussed, the tests fail on the router7 checks, that will be fixed in a separate PR

davebarrau:
> Make the fields of OptIAForPrefixDelegation public. This will make it more consistent with what was done for OptIANA in: [bdb0bf2](https://github.com/insomniacslk/dhcp/commit/bdb0bf25259985b962e230fbac5265a7623afd05)
> 
> It was not possible to do it in PR #166 because at that point it was going to break router7.
> 
> However, this was fixed in [rtr7/router7#9](https://github.com/rtr7/router7/pull/9).
> 
> cc @davebarrau

Nice, thanks. :)

PR #178, opened 2018-10-17
[WIP] Added DHCPv4 server

insomniacslk:
Added DHCPv4 server code and tests.

Still WIP. Tests will fail: requires root access; and port 0 and raw sockets.

### DIFF
--- a/dhcpv4/server_test.go
+++ b/dhcpv4/server_test.go
@@ -128,11 +128,19 @@ func TestServerActivateAndServe(t *testing.T) {
 	lo, err := getLoopbackInterface()
 	require.NoError(t, err)
 
+	xid := uint32(0xaabbccdd)
+	hwaddr := [16]byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 0xa, 0xb, 0xc, 0xd, 0xe, 0xf}
+
 	modifiers := []Modifier{
-		WithTransactionID(0xaabbccdd),
-		WithHwAddr([]byte{0, 1, 2, 3, 4, 5, 6}),
+		WithTransactionID(xid),
+		WithHwAddr(hwaddr[:]),
 	}
 
-	_, err = c.Exchange(lo, nil, modifiers...)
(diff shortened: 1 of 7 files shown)


insomniacslk:
To simplify this PR I need to implement `NewOfferFromDiscovery` and `NewAckFromRequest`

insomniacslk:
The server tests must go in their own integration tests package

dhcpv4/server_test.go
```diff
@@ -0,0 +1,145 @@
...
+	if m.Opcode() != OpcodeBootRequest {
+		log.Printf("Not a BootRequest!")
+		return
+	}
+	response, err := New()
+	if err != nil {
+		log.Printf("New DHCPv4 packet creation failed: %v", err)
+		return
+	}
+	// TODO implement NewOfferFromDiscovery and NewAckFromRequest,
```
pmazzini: How about using `NewReplyFromRequest` here?

insomniacslk: you're damn right, sir!

dhcpv4/server_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+func TestServerActivateAndServe(t *testing.T) {
+	c, s := setUpClientAndServer(DORAHandler)
+	defer s.Close()
+
+	lo, err := getLoopbackInterface()
+	require.NoError(t, err)
+
+	hwaddr := []byte{0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06}
+	discover, err := NewDiscovery(hwaddr)
```
pmazzini: Do we need to create a discover here? How about passing nil to the `Exchange`?

insomniacslk: good point. My initial diff contained a refactor of `Exchange` into `Discover` and `Request`, but I removed it for a future diff. I still need to set the transaction ID and some other stuff, but I can do that via modifiers

insomniacslk:
> I didn't find anything to comment on. Approving to unblock you.

Thanks! I'll try to get coverage to be green before merging :)

PR #179, opened 2018-10-18
Make listening port for dhcpv4 client configurable

gvialetto:
When previously implementing the changes to the dhcpv4 client I allowed the user to set a custom `LocalAddr` - but unfortunately I did not adapt `MakeListeningSocket()` to use that value, so if we set a custom port and the server send us packet back to that port we will not see them because we're constantly listening to port 68.

This PR fixes that issue.

Note I did not change the `MakeListeningSocket()` signature since it's public and we'll break backward compatibility. There are good reasons to make `makeListeningSocketWithCustomPort()` public too, so I'm not against doing that if we want to.

### DIFF
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -118,6 +118,10 @@ func MakeBroadcastSocket(ifname string) (int, error) {
 // MakeListeningSocket creates a listening socket on 0.0.0.0 for the DHCP client
 // port and returns it.
 func MakeListeningSocket(ifname string) (int, error) {
+	return makeListeningSocketWithCustomPort(ifname, ClientPort)
+}
+
+func makeListeningSocketWithCustomPort(ifname string, port int) (int, error) {
 	fd, err := unix.Socket(unix.AF_INET, unix.SOCK_DGRAM, unix.IPPROTO_UDP)
 	if err != nil {
 		return fd, err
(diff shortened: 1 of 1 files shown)


dhcpv4/client.go
```diff
@@ -118,6 +118,10 @@ func MakeBroadcastSocket(ifname string) (int, error) {
 // MakeListeningSocket creates a listening socket on 0.0.0.0 for the DHCP client
 // port and returns it.
 func MakeListeningSocket(ifname string) (int, error) {
+	return makeListeningSocketWithCustomPort(ifname, ClientPort)
+}
+
+func makeListeningSocketWithCustomPort(ifname string, port int) (int, error) {
```
insomniacslk: do we need this? Can't we just call `c.getLocalUDPAddr()` inside MakeListeningSocket` and use that port? The port will be already set accordingly

gvialetto: We don't currently have access to the client inside `MakeListeningSocket()`, so we would have to change that, which I can do without problems (and leave the current function untouched?)

insomniacslk: no, you're right, I didn't realize that. Better to leave the interface as it is. I am not fond of adding new functions, but this one is OK

PR #181, opened 2018-10-28
simplify client interface

pmazzini:
All calls to Exchange(), Solicit() and Request() that I have seen are passing a nil packet.

I think it would be a good idea to simplify this by not passing the packet we want to send and always creating a new one.

The modifiers can completely modify the created packet if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ func main() {
 	// A default Solicit packet will be used during the "conversation",
 	// which can be manipulated by using modifiers.
 	conversation, err := client.Exchange("eth0")
-	
+
 	// Summary() prints a verbose representation of the exchanged packets.
 	for _, packet := range conversation {
 		log.Print(packet.Summary())
--- a/README.md
+++ b/README.md
@@ -59,10 +59,8 @@ func main() {
 	// return a non-empty packet list even if there is an error. This is
 	// intended, because the transaction may fail at any point, and we
 	// still want to know what packets were exchanged until then.
-	// The `nil` argument indicates that we want to use a default Solicit
-	// packet, instead of specifying a custom one ourselves.
-	conversation, err := client.Exchange("eth0", nil)
-
+	conversation, err := client.Exchange("eth0")
+	
 	// Summary() prints a verbose representation of the exchanged packets.
 	for _, packet := range conversation {
 		log.Print(packet.Summary())
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ func main() {
 	// return a non-empty packet list even if there is an error. This is
 	// intended, because the transaction may fail at any point, and we
 	// still want to know what packets were exchanged until then.
+	// A default Solicit packet will be used during the "conversation",
+	// which can be manipulated by using modifiers.
 	conversation, err := client.Exchange("eth0")
 	
 	// Summary() prints a verbose representation of the exchanged packets.
--- a/dhcpv6/client.go
+++ b/dhcpv6/client.go
@@ -38,21 +38,13 @@ func NewClient() *Client {
 }
 
 // Exchange executes a 4-way DHCPv6 request (SOLICIT, ADVERTISE, REQUEST,
-// REPLY). If the SOLICIT packet is nil, defaults are used. The modifiers will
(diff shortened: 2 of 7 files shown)


insomniacslk:
This would not allow the caller anymore to send a custom packet if they want to, and the fact that we don't use it doesn't mean somebody else doesn't. In general I think it's a desirable feature, and after this change this would be simply not possible anymore.
I would also argue that the simplification is not worth the downsides, as I think that passing `nil` does not over-complicate the interface and is still Go idiomatic.

To summarize pros and cons, in case I'm missing something:

Pros:
* simpler interface

Cons:
* significant breaking change
* removes the ability to send custom packets

pmazzini:
The main reasoning behind this is that all the calls to those functions I have seen are simply passing a nil packet.

> significant breaking change

True but I think non of the open source users from the readme file will break with this change.

> removes the ability to send custom packets

We still can do that with modifiers, a modifier can customize the whole packet so it feels redundant.

pmazzini:
I am happy to drop this if it doesn't convince you.

insomniacslk:
> We still can do that with modifiers, a modifier can customize the whole packet so it feels redundant.

This is very interesting, and it's making me change my mind. The modifiers interface already requires the user to know what they're doing (e.g. they have to keep in mind the order of the modifiers). Let's validate it and come up with good documentation around this use case, and this is good to go for me.

On making a breaking change - I'm still a bit uncomfortable. `Exchange` is the primary interface, and I'm sure it would break something out there (systemboot, bender/fbender?), can be fixed of course, but to keep in mind.
On the other side, we haven't stabilized the interface yet, so I may be unnecessarily concerned here.

> I am happy to drop this if it doesn't convince you.

it doesn't have to convince me, it has to convince us :)

pmazzini:
> Let's validate it and come up with good documentation around this use case

Ok, where do you suggest to document it?

> and I'm sure it would break something out there (systemboot, bender/fbender?)

Systemboot calls the netboot package which has been updated with this PR.
Bender and fbender use the async client. I don't think we will break any known OS project.

insomniacslk:
> Ok, where do you suggest to document it?

I guess either the top-level doc in `client.go`, or the functions' doc (or both)

> Systemboot calls the netboot package which has been updated with this PR.
> Bender and fbender use the async client. I don't think we will break any known OS project.

I also checked the other projects that use this library (that I know of, on github), and this change should be safe for all of them.

pmazzini:
Abandoning this for now.

pmazzini:
I though you were not very convinced about the api simplification and I don't have a strong opinion about it, we can keep the ability to pass a discover/solicit.

insomniacslk:
Sorry for the misunderstanding - I still think this is useful

README.md
```diff
@@ -56,9 +56,7 @@ func main() {
 	// return a non-empty packet list even if there is an error. This is
 	// intended, because the transaction may fail at any point, and we
 	// still want to know what packets were exchanged until then.
-	// The `nil` argument indicates that we want to use a default Solicit
-	// packet, instead of specifying a custom one ourselves.
-	conversation, err := client.Exchange("eth0", nil)
+	conversation, err := client.Exchange("eth0")
```
insomniacslk: Can you explain here that it will build a "default" solicit message, and that the user can override it via modifiers?

dhcpv6/client.go
```diff
@@ -189,15 +185,13 @@ func (c *Client) sendReceive(ifname string, packet DHCPv6, expectedType MessageT
...
-// Solicit sends a SOLICIT, return the solicit, an ADVERTISE (if not nil), and
-// an error if any
-func (c *Client) Solicit(ifname string, solicit DHCPv6, modifiers ...Modifier) (DHCPv6, DHCPv6, error) {
-	var err error
-	if solicit == nil {
-		solicit, err = NewSolicitForInterface(ifname)
-		if err != nil {
-			return nil, nil, err
-		}
+// Solicit sends a Solicit, return the Solicit, an Advertise (if not nil), and
```
insomniacslk: can you change "return" to "returns"?

Issue #184, opened 2018-11-01
Problematic DHCPv4 on unconfigured interface 


marcoguerri:
I have been working on implementing `RequestNetbootv4` and stumbled across some issues with the way sockets are used in the v4 part of the library.

If one starts from an interface that does not have any configuration (i.e. no IP, no routes), the attempt to broadcast the DISCOVER fails with ENETUNREACH. The client socket is an `AF_INET`, `SOCK_RAW`, `IPPROTO_RAW` socket, and without any route, even raw traffic gets dropped at layer 3. One possible way around this is to set `SO_BINDTODEVICE` on the device which is being configured, but this solves only the send part of the communication.

The receive socket, without any default route with link scope, does not see the incoming OFFER. I definitely see the OFFER coming up at layer 2 (tcpdump does see it), but it never gets to the socket. 

Adding a default route with link scope before initiating the transaction fixes the problem.

I had a quick look at dhclient from ISC and from what I can see it uses PF_PACKET sockets, so it works at layer 2. I am sending a pull request with a `RequestNetbootv4` implementation, but it's certainly not great as if a default route is not added beforehand on an unconfigured interface, the transaction fails with ENETUNREACH. As I mentioned above, even when binding the sender socket to the interface, the receiving end does not see the broadcasted OFFER. Possible ways to fix this:

* The library might try to be smart and check if a route is present, and if not add it (my `RequestNetbootv4` implementation does not do it).
* Move to layer 2 sockets

PR #185, opened 2018-11-01
Add netboot/netconf support for DHCPv4

marcoguerri:
Adds support for `RequestNetbootv4`. The main problem with this implementation is explained in issue #184. If a default route is not added before calling `RequestNetbootv4`, the transaction fails.

### DIFF
--- a/netboot/netboot.go
+++ b/netboot/netboot.go
@@ -10,6 +10,10 @@ import (
 	"github.com/insomniacslk/dhcp/dhcpv6"
 )
 
+var sleeper = func(d time.Duration) {
+	time.Sleep(d)
+}
+
 // RequestNetbootv6 sends a netboot request via DHCPv6 and returns the exchanged packets. Additional modifiers
 // can be passed to manipulate both solicit and advertise packets.
 func RequestNetbootv6(ifname string, timeout time.Duration, retries int, modifiers ...dhcpv6.Modifier) ([]dhcpv6.DHCPv6, error) {
@@ -37,7 +41,7 @@ func RequestNetbootv6(ifname string, timeout time.Duration, retries int, modifie
 				// don't wait at the end of the last attempt
 				break
 			}
-			time.Sleep(delay)
+			sleeper(delay)
 			// TODO add random splay
 			delay = delay * 2
 			continue
@@ -68,7 +72,7 @@ func RequestNetbootv4(ifname string, timeout time.Duration, retries int, modifie
 				// don't wait at the end of the last attempt
 				break
 			}
-			time.Sleep(delay)
+			sleeper(delay)
 			// TODO add random splay
 			delay = delay * 2
 			continue
--- a/netboot/netboot.go
+++ b/netboot/netboot.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/insomniacslk/dhcp/dhcpv4"
 	"github.com/insomniacslk/dhcp/dhcpv6"
 )
 
@@ -29,6 +30,37 @@ func RequestNetbootv6(ifname string, timeout time.Duration, retries int, modifie
 		// here and one in the above call to NewSolicitForInterface)
 		modifiers = append(modifiers, dhcpv6.WithNetboot)
 		conversation, err = client.Exchange(ifname, solicit, modifiers...)
+		if err != nil {
+			log.Printf("Client.Exchange failed: %v", err)
+			log.Printf("sleeping %v before retrying", delay)
+			if i >= retries {
+				// don't wait at the end of the last attempt
+				break
+			}
+			time.Sleep(delay)
+			// TODO add random splay
+			delay = delay * 2
+			continue
+		}
+		break
+	}
+	return conversation, nil
+}
+
+// RequestNetbootv4 sends a netboot request via DHCPv4 and returns the exchanged packets. Additional modifiers
+// can be passed to manipulate both the discover and offer packets.
+func RequestNetbootv4(ifname string, timeout time.Duration, retries int, modifiers ...dhcpv4.Modifier) ([]*dhcpv4.DHCPv4, error) {
+	var (
+		conversation []*dhcpv4.DHCPv4
+		err          error
+	)
+	delay := 2 * time.Second
+	for i := 0; i <= retries; i++ {
+		log.Printf("sending request, attempt #%d", i+1)
+		modifiers = append(modifiers, dhcpv4.WithNetboot)
+		client := dhcpv4.NewClient()
+		client.ReadTimeout = timeout
+		conversation, err = client.Exchange(ifname, nil, modifiers...)
 		if err != nil {
 			log.Printf("Client.Exchange failed: %v", err)
 			log.Printf("sleeping %v before retrying", delay)
@@ -94,4 +126,29 @@ func ConversationToNetconf(conversation []dhcpv6.DHCPv6) (*NetConf, string, erro
 		bootfile = string(obf.BootFileURL)
 	}
 	return netconf, bootfile, nil
+}
+
+// ConversationToNetconfv4 extracts network configuration and boot file URL from a
+// DHCPv4 4-way conversation and returns them, or an error if any.
+func ConversationToNetconfv4(conversation []*dhcpv4.DHCPv4) (*NetConf, string, error) {
+	var reply *dhcpv4.DHCPv4
+	var bootFileUrl string
+	for _, m := range conversation {
+		// look for a BootReply packet of type Offer containing the bootfile URL.
+		// Normally both packets with Message Type OFFER or ACK do contain
+		// the bootfile URL.
+		if m.Opcode() == dhcpv4.OpcodeBootReply && *m.MessageType() == dhcpv4.MessageTypeOffer {
+			bootFileUrl = m.BootFileNameToString()
+			reply = m
+			break
+		}
+	}
+	if reply == nil {
+		return nil, "", errors.New("no OFFER with valid bootfile URL received")
+	}
+	netconf, err := GetNetConfFromPacketv4(reply)
+	if err != nil {
+		return nil, "", fmt.Errorf("could not get netconf: %v", err)
+	}
+	return netconf, bootFileUrl, nil
 }
--- a/netboot/netboot.go
+++ b/netboot/netboot.go
@@ -55,9 +55,9 @@ func RequestNetbootv4(ifname string, timeout time.Duration, retries int, modifie
 		err          error
 	)
 	delay := 2 * time.Second
+	modifiers = append(modifiers, dhcpv4.WithNetboot)
 	for i := 0; i <= retries; i++ {
 		log.Printf("sending request, attempt #%d", i+1)
-		modifiers = append(modifiers, dhcpv4.WithNetboot)
 		client := dhcpv4.NewClient()
 		client.ReadTimeout = timeout
 		conversation, err = client.Exchange(ifname, nil, modifiers...)
--- a/netboot/netconf.go
+++ b/netboot/netconf.go
@@ -1,6 +1,8 @@
 package netboot
 
 import (
+	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -9,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/insomniacslk/dhcp/dhcpv4"
 	"github.com/insomniacslk/dhcp/dhcpv6"
 	"github.com/vishvananda/netlink"
 )
@@ -25,6 +28,7 @@ type NetConf struct {
 	Addresses     []AddrConf
 	DNSServers    []net.IP
 	DNSSearchList []string
+	Routers       []net.IP
 }
 
 // GetNetConfFromPacketv6 extracts network configuration information from a DHCPv6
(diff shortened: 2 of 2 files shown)


netboot/netboot.go
```diff
@@ -46,6 +47,39 @@ func RequestNetbootv6(ifname string, timeout time.Duration, retries int, modifie
...
 }
 
+func RequestNetbootv4(ifname string, timeout time.Duration, retries int, modifiers ...dhcpv4.Modifier) ([]*dhcpv4.DHCPv4, error) {
+	var (
+		conversation []*dhcpv4.DHCPv4
+	)
+	delay := 2 * time.Second
+	for i := 0; i <= retries; i++ {
+		log.Printf("sending request, attempt #%d", i+1)
+		discover, err := dhcpv4.NewDiscoveryForInterface(ifname)
```
pmazzini: Although this is consistent with what is being done in v6, I think it can be simplified by not creating the discover packet and passing nil to the `Exchange` call.

netboot/netconf.go
```diff
@@ -1,16 +1,18 @@
 package netboot
 
 import (
+	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/insomniacslk/dhcp/dhcpv4"
```
pmazzini: I think we want to keep the github imports in a different group:
https://github.com/golang/go/wiki/CodeReviewComments#imports

netboot/netconf.go
```diff
@@ -74,6 +77,93 @@ func GetNetConfFromPacketv6(d *dhcpv6.DHCPv6Message) (*NetConf, error) {
...
+	// netconf struct requires a valid lifetime to be specified. ValidLifetime is a dhcpv6
+	// concept, the closest mapping in dhcpv4 world is "IP Address Lease Time". If the lease
+	// time option is nil, we set it to 0
+	leaseTimeOption := d.GetOneOption(dhcpv4.OptionIPAddressLeaseTime)
+	leaseTime := uint32(0)
+	if leaseTimeOption != nil {
+		leaseTime = leaseTimeOption.(*dhcpv4.OptIPAddressLeaseTime).LeaseTime
+	}
+
+	// TODO: The lease time extracted from the packet does not seem to match what tcpdump
```
pmazzini: As discussed in person, I would like to double check this.

netboot/netconf.go
```diff
@@ -127,5 +217,36 @@ func ConfigureInterface(ifname string, netconf *NetConf) error {
...
+	iface, err = netlink.LinkByName(ifname)
+	if err != nil {
+		return fmt.Errorf("could not obtain interface when adding default route: %v", err)
+	}
+
+	dst := &net.IPNet{
+		IP:   net.IPv4(0, 0, 0, 0),
+		Mask: net.CIDRMask(0, 32),
+	}
+
```
pmazzini: I wonder about the implications of doing this for a v6 client.

netboot/netconf.go
marcoguerri: Yeah you are totally right, I was looking into the ACK's lease time. The OFFER lease time does match. I was chasing ghosts.

netboot/netconf.go
marcoguerri: Got it, will amend.

netboot/netconf.go
marcoguerri: As discussed, this can be indeed moved a bit further down where I check if I got a default route from the packet. If that's the case, then I don't have any other option but removing the current default route. If we want to support IPv6 routes via DHCPv6, then the `Netconf ` struct will have to be extended to hold both v4 and v6 routes. I can straight away rename `Routers` to `Routersv4`, if you think there is good chance that the library will get IPv6 routes support.

netboot/netboot.go
marcoguerri: Wait a sec, I might not be able to do that, as the discover I pass to `Exchange` has been modified with a `dhcpv4.WithNetboot` modifier.

netboot/netconf.go
pmazzini: Moving a bit further look good to me.

marcoguerri:
Addressed the various comments. One problem that I realized exists with this implementation but forgot to mention is that resolv.conf gets overwritten. It seems to be the same behavior as `dhclient`.

netboot/netboot.go
```diff
@@ -46,6 +47,37 @@ func RequestNetbootv6(ifname string, timeout time.Duration, retries int, modifie
...
+// can be passed to manipulate both the discover and offer packets.
+func RequestNetbootv4(ifname string, timeout time.Duration, retries int, modifiers ...dhcpv4.Modifier) ([]*dhcpv4.DHCPv4, error) {
+	var (
+		conversation []*dhcpv4.DHCPv4
+		err          error
+	)
+	delay := 2 * time.Second
+	for i := 0; i <= retries; i++ {
+		log.Printf("sending request, attempt #%d", i+1)
+		modifiers = append(modifiers, dhcpv4.WithNetboot)
```
pmazzini: nit: I would move this outside of the for to avoid appending the same modifier in the retries.

Apart from this, it looks good to me.

netboot/netboot.go
```diff
@@ -46,6 +47,37 @@ func RequestNetbootv6(ifname string, timeout time.Duration, retries int, modifie
...
+		client.ReadTimeout = timeout
+		conversation, err = client.Exchange(ifname, nil, modifiers...)
+		if err != nil {
+			log.Printf("Client.Exchange failed: %v", err)
+			log.Printf("sleeping %v before retrying", delay)
+			if i >= retries {
+				// don't wait at the end of the last attempt
+				break
+			}
+			time.Sleep(delay)
```
insomniacslk: calls that sleep are not test-friendly - what about making the initial sleep time configurable, or the sleep function configurable? They could be top-level variables so that the caller can override them (e.g. `netboot.Sleeper` and `netboot.InitialRetryInterval`). Another option for the interval is to become a function argument, but I'd rather go for top-level var to conform with the v6 version of this function.

netboot/netboot.go
```diff
@@ -95,3 +127,28 @@ func ConversationToNetconf(conversation []dhcpv6.DHCPv6) (*NetConf, string, erro
...
+		// Normally both packets with Message Type OFFER or ACK do contain
+		// the bootfile URL.
+		if m.Opcode() == dhcpv4.OpcodeBootReply && *m.MessageType() == dhcpv4.MessageTypeOffer {
+			bootFileUrl = m.BootFileNameToString()
+			reply = m
+			break
+		}
+	}
+	if reply == nil {
+		return nil, "", errors.New("no OFFER with valid bootfile URL received")
```
insomniacslk: what about mentioning ACK too?

```
errors.New("no OFFER or ACK with valid bootfile URL received")
```

netboot/netconf.go
```diff
@@ -74,6 +78,91 @@ func GetNetConfFromPacketv6(d *dhcpv6.DHCPv6Message) (*NetConf, error) {
...
+	netconf.Addresses = append(netconf.Addresses, AddrConf{
+		IPNet: net.IPNet{
+			IP:   ipAddr,
+			Mask: netmask,
+		},
+		PreferredLifetime: 0,
+		ValidLifetime:     int(leaseTime),
+	})
+	if bytes.Equal(netmask, []byte{0, 0, 0, 0}) {
```
insomniacslk: didn't you check this already at line 98?

netboot/netboot.go
marcoguerri: Not sure what you mean with "conform with the v6 version of this function", as they are making identical use of sleep. I will amend both accordingly. Change in function arguments breaks the API if I want to align `RequestNetbootv6`, so I don't think it's an option.

netboot/netconf.go
```diff
@@ -74,6 +78,91 @@ func GetNetConfFromPacketv6(d *dhcpv6.DHCPv6Message) (*NetConf, error) {
...
+	netconf := NetConf{}
+
+	// get the subnet mask from OptionSubnetMask. If the netmask is not defined
+	// in the packet, an error is returned
+	netmaskOption := d.GetOneOption(dhcpv4.OptionSubnetMask)
+	if netmaskOption == nil {
+		return nil, errors.New("no netmask option in response packet")
+	}
+	netmask := netmaskOption.(*dhcpv4.OptSubnetMask).SubnetMask
+	if binary.LittleEndian.Uint32(netmask) == 0 {
```
insomniacslk: a stronger check (i.e. with less surprises, given how Go handles IP data) could be
```
ones, bits := netmask.Size()
if bits != 32 {
    return ...
}
if ones == 0 {
    return ...
}
```

You can probably omit the `bits` check though

netboot/netboot.go
marcoguerri: I'll then need to amend the logic which looks for bootfileurl, as now I am looking only into OFFERs:
```
m.Opcode() == dhcpv4.OpcodeBootReply && *m.MessageType() == dhcpv4.MessageTypeOffer 
```

netboot/netconf.go
```diff
@@ -74,6 +78,91 @@ func GetNetConfFromPacketv6(d *dhcpv6.DHCPv6Message) (*NetConf, error) {
...
+	}
+
+	// get DNS configuration
+	dnsServersOption := d.GetOneOption(dhcpv4.OptionDomainNameServer)
+	if dnsServersOption == nil {
+		return nil, errors.New("no dns server option in response packet")
+	}
+	dnsServers := dnsServersOption.(*dhcpv4.OptDomainNameServer).NameServers
+	if len(dnsServers) == 0 {
+		return nil, errors.New("no dns servers options in response packet")
```
insomniacslk: nit: this message may be confusing with the one above, maybe "name servers list is empty" or similar?

netboot/netconf.go
```diff
@@ -74,6 +78,91 @@ func GetNetConfFromPacketv6(d *dhcpv6.DHCPv6Message) (*NetConf, error) {
...
+	dnsSearchList := dnsDomainSearchListOption.(*dhcpv4.OptDomainSearch).DomainSearch
+	if len(dnsSearchList) == 0 {
+		return nil, errors.New("dns search list is empty")
+	}
+	netconf.DNSSearchList = dnsSearchList
+
+	// get default gateway
+	routerOption := d.GetOneOption(dhcpv4.OptionRouter)
+	if routerOption == nil {
+		return nil, errors.New("no router option specified in reponse packet")
```
insomniacslk: typo: reponse -> response

netboot/netboot.go
insomniacslk: On the interface, I was talking about the interface, not the content of the function.
Thanks for fixing the v6 version too!

netboot/netboot.go
insomniacslk: ignore this, I misread the code

netboot/netconf.go
```diff
@@ -127,5 +216,34 @@ func ConfigureInterface(ifname string, netconf *NetConf) error {
...
+	if err = ioutil.WriteFile("/etc/resolv.conf", []byte(resolvconf), 0644); err != nil {
+		return fmt.Errorf("could not write resolv.conf file %v", err)
+	}
+
+	// add default route information for v4 space. only one default route is allowed
+	// so ignore the others if there are multiple ones
+	if len(netconf.Routers) > 0 {
+		iface, err = netlink.LinkByName(ifname)
+		if err != nil {
+			return fmt.Errorf("could not obtain interface when adding default route: %v", err)
```
insomniacslk: nit: mention the interface in the error too?

netboot/netconf.go
```diff
@@ -127,5 +216,34 @@ func ConfigureInterface(ifname string, netconf *NetConf) error {
...
+		}
+		// if there is a default v4 route, remove it, as we want to add the one we just got during
+		// the dhcp transaction. if the route is not present, which is the final state we want,
+		// an error is returned so ignore it
+		dst := &net.IPNet{
+			IP:   net.IPv4(0, 0, 0, 0),
+			Mask: net.CIDRMask(0, 32),
+		}
+		route := netlink.Route{LinkIndex: iface.Attrs().Index, Dst: dst, Src: net.IPv4(0, 0, 0, 0)}
```
insomniacslk: question: does this work also when there are multiple default routes?

netboot/netconf.go
```diff
@@ -127,5 +216,34 @@ func ConfigureInterface(ifname string, netconf *NetConf) error {
...
+		netlink.RouteDel(&route)
+
+		src := netconf.Addresses[0].IPNet.IP
+		route = netlink.Route{LinkIndex: iface.Attrs().Index, Dst: dst, Src: src, Gw: netconf.Routers[0]}
+		err = netlink.RouteAdd(&route)
+		if err != nil {
+			return fmt.Errorf("could not add default route: %v", err)
```
insomniacslk: nit: also print `route`? Using `%+v` so we have all the information there

netboot/netconf.go
marcoguerri: Yes, it must have slipped through the cracks from previous versions. I will leave only the one above.

netboot/netconf.go
marcoguerri: I haven't honestly tested with multiple default routes. I guess this will delete all default source based and non source based routes. I can test that scenario as well, but in a scenario with multiple routes, this is very intrusive. At the same time, I wouldn't have, off the top of my head, an alternative solution.

insomniacslk: I don't know the exact solution either, I thought that a loop + check could be enough, but it's probably overcomplicating things. At this point let's leave this as it is, and maybe add a comment about "no idea what happens with multiple default routes". Would that make sense?

marcoguerri: I'll see at least what happens with multiple default routes and leave a comment behind.

marcoguerri: I have experimented a bit and the control path above actually removes only the default route (dst 0.0.0.0) to the L2 domain (gw: 0.0.0.0), which is what one would want to add to an unconfigured interface to make DHCP transaction succeed. Also, if this default route has a specific metric assigned, it doesn't get removed. The code doesn't remove any other default route (i.e. gw != 0.0.0.0). The code then adds a default route via the gateway specified in the OFFER for only the SRC address specified in the OFFER. Looks like a sane behavior.

marcoguerri: I left the above as a comment in the code.

netboot/netboot.go
```diff
@@ -6,16 +6,23 @@ import (
 	"log"
 	"time"
 
+	"github.com/insomniacslk/dhcp/dhcpv4"
 	"github.com/insomniacslk/dhcp/dhcpv6"
 )
 
+var retryDelay = 2 * time.Second
+
+var Sleeper = func(multiplier int) {
```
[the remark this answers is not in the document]
insomniacslk: I believe this should have the same signature as `time.Sleep`, https://github.com/golang/go/blob/master/src/time/sleep.go#L9

i.e.
```
var Sleeper = func(d Duration) {
    time.Sleep(d)
}
```

PR #187, opened 2018-11-06
[ztpv4] add Opt60 (vc) parsing

pmazzini:
How about moving the parsing to extract vendor, model and serial from https://github.com/facebookincubator/dhcplb/blob/master/lib/handler.go#L418 to the library.

### DIFF
--- a/dhcpv4/ztp/ztp.go
+++ b/dhcpv4/ztp/ztp.go
@@ -20,14 +20,6 @@ var errVendorOptionMalformed = errors.New("malformed vendor option")
 // ParseVendorData will try to parse dhcp4 options looking for more
 // specific vendor data (like model, serial number, etc).
 func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
-	return parseV4VendorClass(packet);
-}
-
-// parseV4Opt60 will attempt to look at the Vendor Class option (Option 60) on
-// DHCPv4.  The option is formatted as a string with the content being specific
-// for the vendor, usually using a delimitator to separate the values.
-// See: https://tools.ietf.org/html/rfc1533#section-9.11
-func parseV4VendorClass(packet *dhcpv4.DHCPv4) (*VendorData, error) {
 	opt := packet.GetOneOption(dhcpv4.OptionClassIdentifier)
 	if opt == nil {
 		return nil, nil
@@ -66,27 +58,20 @@ func parseV4VendorClass(packet *dhcpv4.DHCPv4) (*VendorData, error) {
 	//    Juniper-qfx10008           <vendor>-<model> (serial in hostname option)
 	//    Juniper-qfx10002-361-DN817 <vendor>-<model>-<serial> (model has a dash in it!)
 	case strings.HasPrefix(vc, "Juniper-"):
-		// strip of the prefix
-		vc := vc[len("Juniper-"):]
-		vd.VendorName = "Juniper"
-
-		sepIdx := strings.LastIndex(vc, "-")
-		if sepIdx == -1 {
-			// No separator was found. Attempt serial number from the hostname
+		p := strings.Split(vc, "-")
+		if len(p) < 3 {
+			vd.Model = p[1]
 			if opt := packet.GetOneOption(dhcpv4.OptionHostName); opt != nil {
 				vd.Serial = opt.(*dhcpv4.OptHostName).HostName
 			} else {
-				return nil, errVendorOptionMalformed
+				return nil, errors.New("host name option is missing")
 			}
 		} else {
-			if len(vc) == sepIdx+1 {
-				return nil, errVendorOptionMalformed
-			}
-			vd.Serial = vc[sepIdx+1:]
-			vc = vc[:sepIdx]
+			vd.Model = strings.Join(p[1:len(p)-1], "-")
+			vd.Serial = p[len(p)-1]
 		}
-		vd.Model = vc
 
+		vd.VendorName = p[0]
 		return vd, nil
 	}
 
--- a/dhcpv4/ztp/ztp.go
+++ b/dhcpv4/ztp/ztp.go
@@ -1,0 +1,139 @@
+package ztpv4
+
+import (
+	"bytes"
+	"errors"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/insomniacslk/dhcp/dhcpv4"
+)
+
+// VendorData is optional data a particular vendor may or may not include
+// in the Vendor Class options.  All values are optional and will be zero
+// values if not found.
+type VendorData struct {
+	VendorName string
+	Model      string
+	Serial     string
+}
+
+var errVendorOptionMalformed = errors.New("malformed vendor option")
+
+// VendorDataV4 will try to parse dhcp4 options data looking for more specific
+// vendor data (like model, serial number, etc).  If the options are missing
+func VendorDataV4(packet *dhcpv4.DHCPv4) VendorData {
+	vd := VendorData{}
+
+	if err := parseV4VendorClass(&vd, packet); err != nil {
+		glog.Errorf("failed to parse vendor data from vendor class: %v", err)
+	}
+
+	if err := parseV4VIVC(&vd, packet); err != nil {
+		glog.Errorf("failed to parse vendor data from vendor-idenitifying vendor class: %v", err)
+	}
+
+	return vd
+}
+
+// parseV4Opt60 will attempt to look at the Vendor Class option (Option 60) on
+// DHCPv4.  The option is formatted as a string with the content being specific
+// for the vendor, usually using a deliminator to separate the values.
+// See: https://tools.ietf.org/html/rfc1533#section-9.11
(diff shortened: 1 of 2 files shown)


insomniacslk:
sounds good to me :) but I need time to review it. I know it works, I just want to make sure it's aligned with the conventions used throughout this library

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+	VendorName string
+	Model      string
+	Serial     string
+}
+
+var errVendorOptionMalformed = errors.New("malformed vendor option")
+
+// VendorDataV4 will try to parse dhcp4 options data looking for more specific
+// vendor data (like model, serial number, etc).  If the options are missing
+func VendorDataV4(packet *dhcpv4.DHCPv4) VendorData {
```
insomniacslk: this function should not have the redundant "v4" at the end as it's part of the dhcpv4 package already. A more appropriate name would be `ParseVendorData`, because that's what the function does

insomniacslk: Also, why returning by value? I understand that it's a small structure, but this function's signature should be different, and also return an error:
```
func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error)
```

And should use the regular `log` module, not `glog`

dhcpv4/ztp/ztp_test.go
```diff
@@ -0,0 +1,134 @@
+package ztpv4
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
```
insomniacslk: similar thing here - let's reuse `stretchr/testify` for tests

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+	if err := parseV4VIVC(&vd, packet); err != nil {
+		glog.Errorf("failed to parse vendor data from vendor-idenitifying vendor class: %v", err)
+	}
+
+	return vd
+}
+
+// parseV4Opt60 will attempt to look at the Vendor Class option (Option 60) on
+// DHCPv4.  The option is formatted as a string with the content being specific
+// for the vendor, usually using a deliminator to separate the values.
```
insomniacslk: typo: deliminator -> delimitator

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	if err := parseV4VIVC(&vd, packet); err != nil {
+		glog.Errorf("failed to parse vendor data from vendor-idenitifying vendor class: %v", err)
+	}
+
+	return vd
+}
+
+// parseV4Opt60 will attempt to look at the Vendor Class option (Option 60) on
```
insomniacslk: This comment omits an important thing about what the function does. This function modifies *in-place* the input object. I'm not a fan of this approach because it is less obvious to understand what it does, by its name. Wondering if we should instead return a copy with the modified fields along with an error

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+			return errVendorOptionMalformed
+		}
+
+		vd.VendorName = p[0]
+		vd.Model = p[1]
+		vd.Serial = p[2]
+		return nil
+
+	// Juniper option 60 parsing is a bit more nuanced.  The following are all
+	// "valid" indetifing stings for Juniper:
```
insomniacslk: typo: indetifing -> identifying

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+		vd.Model = vc
+
+		return nil
+	}
+
+	// We didn't match anything, just return an empty vendor data.
+	return nil
+}
+
+const entIDCiscoSystems = 0x9
```
insomniacslk: should this be a public, top-level variable? Any chance it could be reused outside of this module?

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+func parseV4VendorClass(vd *VendorData, packet *dhcpv4.DHCPv4) error {
+	opt := packet.GetOneOption(dhcpv4.OptionClassIdentifier)
+	if opt == nil {
+		return nil
+	}
+	vc := opt.(*dhcpv4.OptClassIdentifier).Identifier
+
+	switch {
+	// Arista;DCS-7050S-64;01.23;JPE12221671
+	case strings.HasPrefix(vc, "Arista;"):
```
insomniacslk: any chance that we may have to do case-insensitive matching?

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil
+
+	// Juniper option 60 parsing is a bit more nuanced.  The following are all
+	// "valid" indetifing stings for Juniper:
+	//    Juniper-ptx1000-DD576      <vendor>-<model>-<serial
+	//    Juniper-qfx10008           <vendor>-<model> (serial in hostname option)
+	//    Juniper-qfx10002-361-DN817 <vendor>-<model>-<serial> (model has a dash in it!)
+	case strings.HasPrefix(vc, "Juniper-"):
+		// strip of the prefix
+		vc := vc[len("Juniper-"):]
```
insomniacslk: why not splitting here too? It seems that all the fields are dash-separated in every case, and it would simplify the following code

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+		vd.VendorName = "Juniper"
+
+		sepIdx := strings.LastIndex(vc, "-")
+		if sepIdx == -1 {
+			// No separator was found. Attempt serial number from the hostname
+			if opt := packet.GetOneOption(dhcpv4.OptionHostName); opt != nil {
```
insomniacslk: what if no hostname is found? Should leave it empty, or return an error?

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// We didn't match anything, just return an empty vendor data.
+	return nil
+}
+
+const entIDCiscoSystems = 0x9
+
+// parseV4Opt124 will attempt to read the Vendor-Identifying Vendor Class
+// (Option 124) on a DHCPv4 packet.  The data is represented in a length/value
+// format with an indentifying enterprise number.
```
insomniacslk: typo: indentifying -> identifying

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+//
+// See: https://tools.ietf.org/html/rfc3925
+func parseV4VIVC(vd *VendorData, packet *dhcpv4.DHCPv4) error {
```
insomniacslk: same thoughts for `parseV4VendorClass` apply here for the return value and comments

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,139 @@
...
+	opt := packet.GetOneOption(dhcpv4.OptionVendorIdentifyingVendorClass)
+	if opt == nil {
+		return nil
+	}
+	ids := opt.(*dhcpv4.OptVIVC).Identifiers
+
+	for _, id := range ids {
```
insomniacslk: the paranoid in me would also check that `ids != nil`

dhcpv4/ztp/ztp.go
pmazzini: Not that I am aware of, I think it should always match it like that.

dhcpv4/ztp/ztp.go
pmazzini: I think this was done this way because the model may have a dash in it, example: Juniper-qfx10002-361-DN817.

pmazzini:
I removed the Opt124 parsing since we do not need it right now.

dhcpv4/ztp/ztp.go
insomniacslk: I see - what I meant is, why don't we just split by dashes, and use the number of fields to know if the model has a dash in it or not? E.g. if len is 4, the two middle fields are the parts of the model name. This means you have to rejoin them, but it simplifies all the rest of the code block

dhcpv4/ztp/ztp.go
```diff
@@ -0,0 +1,95 @@
...
+		vc := vc[len("Juniper-"):]
+		vd.VendorName = "Juniper"
+
+		sepIdx := strings.LastIndex(vc, "-")
+		if sepIdx == -1 {
+			// No separator was found. Attempt serial number from the hostname
+			if opt := packet.GetOneOption(dhcpv4.OptionHostName); opt != nil {
+				vd.Serial = opt.(*dhcpv4.OptHostName).HostName
+			} else {
+				return nil, errVendorOptionMalformed
```
insomniacslk: Technically the vendor option is not malformed, it's just missing required information from another option. This could be its own error, like:
```
return nil, errors.New("host name is missing from both vendor option and host name option")
```

or similar

PR #182, opened 2018-10-28
remove deprecated functions

pmazzini:

PR #189, opened 2018-11-07
bsdp: Fix client to not use deprecated functions

get9:
Fixes the bsdp client to not use deprecated functionality; instead,
reuse parts of dhcpv4.Client and implement only the BDSP transaction
logic (effectively overloading the client.Exchange method).

### DIFF
--- a/dhcpv4/bsdp/client.go
+++ b/dhcpv4/bsdp/client.go
@@ -2,21 +2,26 @@ package bsdp
 
 import (
 	"errors"
+	"time"
 
 	"github.com/insomniacslk/dhcp/dhcpv4"
 )
 
 // Client represents a BSDP client that can perform BSDP exchanges via the
 // broadcast address.
-type Client dhcpv4.Client
+type Client struct {
+	ReadTimeout, WriteTimeout time.Duration
+	dhcpv4.Client
+}
 
 // NewClient constructs a new client with default read and write timeouts from
 // dhcpv4.Client.
-func NewClient() *Client {
-	c := dhcpv4.NewClient()
+func NewClient(readTimeout, writeTimeout time.Duration) *Client {
 	return &Client{
-		ReadTimeout:  c.ReadTimeout,
-		WriteTimeout: c.WriteTimeout,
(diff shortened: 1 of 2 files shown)


pmazzini:
This is similar to what I started doing in PR #182.

dhcpv4/bsdp/client.go
```diff
@@ -2,21 +2,26 @@ package bsdp
...
+type Client struct {
+	ReadTimeout, WriteTimeout time.Duration
+	dhcpv4.Client
+}
 
 // NewClient constructs a new client with default read and write timeouts from
 // dhcpv4.Client.
-func NewClient() *Client {
-	c := dhcpv4.NewClient()
+func NewClient(readTimeout, writeTimeout time.Duration) *Client {
```
pmazzini: We were previously using the default here when creating a new client. Is there any reason to change that?

get9:
Ah yeah, that's what I get for not checking open PR's...

dhcpv4/bsdp/client.go
get9: I couldn't come up with a good way to try and share code from `dhcpv4.Client` with `bsdp.Client`. Effectively all I want to do is overwrite the `Client.Exchange` method, but Go doesn't really let me do that in an easy way. The cleanest way I figured was type-embedding a `dhcpv4.Client` inside a `bsdp.Client`, but I'd still like to pass the read/write timeouts directly to the client. A constructor is the only easy way of doing so.

insomniacslk: I agree with @pmazzini, I feel more beneficial to make Client public here. The default constructor in my opinion should only offer arguments for things that are necessary to the transaction and for which there are no sane defaults, which is not the case for timeouts

pmazzini:
It would be nice to get this merged to remove the deprecated functions. @get9 Can the default client constructor be changed?

pmazzini:
I took care of it in #182, which also removes the deprecated functions.

pmazzini:
I think this can be closed.

PR #196, opened 2018-11-13
Consolidated v6 and v4 server handlers

insomniacslk:
Code for v6 and v4 servers is now consolidated. There are also various bug fixes.

### DIFF
--- a/dhcpv4/server.go
+++ b/dhcpv4/server.go
@@ -82,17 +82,17 @@ func (s *Server) LocalAddr() net.Addr {
 // background, and can be interrupted with `Server.Close`.
 func (s *Server) ActivateAndServe() error {
 	s.connMutex.Lock()
-	// FIXME check if conn != nil instead, and close it if so. It may
-	//       panic if it's in an invalid state, but better than assuming
-	//       that the already open connection is usable.
-	if s.conn == nil {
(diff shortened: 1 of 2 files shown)


dhcpv4/server.go
```diff
@@ -90,49 +97,52 @@ func (s *Server) ActivateAndServe() error {
...
-		}
-		pc.SetReadDeadline(time.Now().Add(time.Second))
-		rbuf := make([]byte, 4096) // FIXME this is bad
-		n, peer, err := pc.ReadFrom(rbuf)
-		if err != nil {
-			switch err.(type) {
-			case net.Error:
-				if err.(net.Error).Timeout() {
-					return err
+	go func() {
```
pmazzini: Do we need a goroutine here? We are already using 1 goroutine per response handler.

insomniacslk: ...I wanted to say a channel, sorry. Fixing

insomniacslk:
@pmazzini code updated to be bugfix and consolidation only

dhcpv4/server.go
insomniacslk: discussed offline - as you suggest it's not necessary to make the function non-blocking here. It can be easily achieved by the caller, and error handling is easier.
I have reverted the goroutine part and left only the bug fixes and code consolidation

PR #204, opened 2018-11-20
rfc1035label: support compression pointers

### LINKED ISSUE #203: Implement DNS compression in rfc1035label package
Label compression is not implemented, I have hit a DHCP server that responds with compressed labels and the library failed to parse the response

insomniacslk:
The `rfc1035label` module did not support compressed labels, now it does.

Fixes #203

### DIFF
--- a/rfc1035label/label.go
+++ b/rfc1035label/label.go
@@ -1,39 +1,61 @@
 package rfc1035label
 
 import (
-	"fmt"
+	"errors"
 	"strings"
 )
 
-// This implements the compression from RFC 1035, section 4.1.4
-// https://tools.ietf.org/html/rfc1035
+// This implements RFC 1035 labels, including compression.
+// https://tools.ietf.org/html/rfc1035#section-4.1.4
 
 // LabelsFromBytes decodes a serialized stream and returns a list of labels
 func LabelsFromBytes(buf []byte) ([]string, error) {
 	var (
-		pos     = 0
-		domains = make([]string, 0)
-		label   = ""
+		pos, oldPos     int
+		labels          = make([]string, 0)
+		label           string
+		handlingPointer bool
 	)
+
 	for {
 		if pos >= len(buf) {
-			return domains, nil
+			break
 		}
 		length := int(buf[pos])
 		pos++
+		var chunk string
 		if length == 0 {
-			domains = append(domains, label)
+			labels = append(labels, label)
 			label = ""
+			if handlingPointer {
+				pos = oldPos
+				handlingPointer = false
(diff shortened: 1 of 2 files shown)


rfc1035label/label_test.go
```diff
@@ -52,15 +42,85 @@ func TestLabelToBytes(t *testing.T) {
...
+		// slackware.it
+		9, 's', 'l', 'a', 'c', 'k', 'w', 'a', 'r', 'e',
+		2, 'i', 't',
+		0,
+		// insomniac.slackware.it
+		9, 'i', 'n', 's', 'o', 'm', 'n', 'i', 'a', 'c',
+		192,
+	}
+
+	_, err := LabelsFromBytes(data)
```
pmazzini: Should the return value also be tested?

rfc1035label/label_test.go
```diff
@@ -52,15 +42,85 @@ func TestLabelToBytes(t *testing.T) {
...
+		3, 'i', 't',
+		0,
+		// slackware.it
+		9, 's', 'l', 'a', 'c', 'k', 'w', 'a', 'r', 'e',
+		192, 0,
+		// insomniac.slackware.it
+		9, 'i', 'n', 's', 'o', 'm', 'n', 'i', 'a', 'c',
+		192, 5,
 	}
+	_, err := LabelsFromBytes(data)
```
pmazzini: Should the return value also be tested?

rfc1035label/label.go
```diff
@@ -1,39 +1,62 @@
...
+				return nil, errors.New("rfc1035label: cannot handle nested pointers")
+			}
+			handlingPointer = true
+			if pos+1 > len(buf) {
+				return nil, errors.New("rfc1035label: pointer buffer too short")
+			}
+			off := int(buf[pos-1]&^0xc0)<<8 + int(buf[pos])
+			oldPos = pos + 1
+			pos = off
+			continue
```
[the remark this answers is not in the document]
insomniacslk: not anymore, it's a leftover from the previous iteration, thanks!

rfc1035label/label_test.go
insomniacslk: not in this case - the function returns an error from the invalid data, no valid data is expected to be returned

rfc1035label/label_test.go
insomniacslk: same as above

rfc1035label/label_test.go
pmazzini: Oh right

PR #205, opened 2018-11-20
rfc1035label: using a structure to hold original data

insomniacslk:
rfc1035label now is implemented on top of a `Labels` struct that contains the original data in case no changes are made. This is required to parse responses from DHCP servers that use label compression.


pmazzini:
Looks good. I will do some more testing before stamping it.

PR #208, opened 2018-11-20
Parsing Vendor Opts lib for ztpv6

gorhamc:
This mimics the DHCPv4 vendor option parsing done currently in the DHCPLB algo code.  This should align with the planned directory structure of moving the parsing out of the dhcplb algo and into dhcpv[4|6] directories.

### DIFF
--- a/dhcpv6/ztpv6/README.md
+++ b/dhcpv6/ztpv6/README.md
@@ -1,0 +1,16 @@
+# Zero Touch Provisioning (ZTP) DHCPv6 Parsing for Network Hardware Vendors
+
+## Currently Supported Vendors For DHCPv6 ZTP
+ - Arista
+ - ZPE
+
+## Why Do We Need This?
+Many network hardware vendors support features that allow network devices to provision themselves with proper supporting automation/tools. Network devices can rely on DHCP and other methods to gather bootfile info, IPs, etc. DHCPv6 Vendor options provides us Vendor Name, Make, Model, and Serial Number data. This data can be used to uniquely identify individual network devices at provisioning time and can be used by tooling to make decisions necessary to correctly and reliably provision a network device.
+
+For more details on a large-scale ZTP deployment, check out how this is done at Facebook, [Scaling Backbone Networks Through Zero Touch Provisioning](https://code.fb.com/networking-traffic/scaling-the-facebook-backbone-through-zero-touch-provisioning/).
+
+
+### Example Data
+Vendor specific data is commonly in a delimiter separated format containing Vendor Name, Model, Make, and Serial Number. This of course will vary per vendor and there could be more or less data.
+Vendor;Model;Version;SerialNumber
+`Arista;DCS-7060;01.011;ZZZ00000000`
--- a/dhcpv6/ztpv6/parse_remote_id.go
+++ b/dhcpv6/ztpv6/parse_remote_id.go
@@ -1,0 +1,84 @@
+package ztpv6
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+
+	"github.com/insomniacslk/dhcp/dhcpv6"
+)
+
+var (
+	// Arista Port, Vlan Pattern
+	aristaPVPattern = regexp.MustCompile("Ethernet(?P<port>[0-9]+):(?P<vlan>[0-9]+)")
+	// Arista Slot, Mod, Port Pattern
+	aristaSMPPattern = regexp.MustCompile("Ethernet(?P<slot>[0-9]+)/(?P<module>[0-9]+)/(?P<port>[0-9]+)")
+)
+
+// CircuitID represents the structure of network vendor interface formats
+type CircuitID struct {
+	Slot    string
+	Module  string
+	Port    string
+	SubPort string
+	Vlan    string
+}
+
+// ParseRemoteId will parse the RemoteId Option data for Vendor Specific data
+func ParseRemoteId(packet dhcpv6.DHCPv6) (*CircuitID, error) {
+	// Need to decapsulate the packet after multiple relays in order to reach RemoteId data
+	inner, err := dhcpv6.DecapsulateRelayIndex(packet, -1)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decapsulate relay index: %v", err)
+	}
+
+	if rid := inner.GetOneOption(dhcpv6.OptionRemoteID); rid != nil {
+		remoteID := string(rid.(*dhcpv6.OptRemoteId).RemoteID())
+		circ, err := matchCircuitId(remoteID)
+		if err != nil {
+			return nil, err
+		}
+		return circ, nil
+	}
+	return nil, errors.New("failed to parse RemoteID option data")
+}
+
+func matchCircuitId(remoteID string) (*CircuitID, error) {
+	var names, matches []string
+
+	switch {
+	case aristaPVPattern.MatchString(remoteID):
+		matches = aristaPVPattern.FindStringSubmatch(remoteID)
+		names = aristaPVPattern.SubexpNames()
+	case aristaSMPPattern.MatchString(remoteID):
+		matches = aristaSMPPattern.FindStringSubmatch(remoteID)
+		names = aristaSMPPattern.SubexpNames()
+	}
+
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("no circuitId regex matches for %v", remoteID)
+	}
+
+	var circuit CircuitID
+	for i, match := range matches {
+		switch names[i] {
+		case "port":
+			circuit.Port = match
+		case "slot":
+			circuit.Slot = match
+		case "module":
+			circuit.Module = match
+		case "subport":
+			circuit.SubPort = match
+		case "vlan":
+			circuit.Vlan = match
+		}
+	}
+
+	return &circuit, nil
+}
+
+// FormatCircuitID is the CircuitID format we send in our Bootfile URL for ZTP devices
+func (c *CircuitID) FormatCircuitID() string {
+	return fmt.Sprintf("%v,%v,%v,%v,%v", c.Slot, c.Module, c.Port, c.SubPort, c.Vlan)
+}
--- a/dhcpv6/ztpv6/parse_remote_id_test.go
+++ b/dhcpv6/ztpv6/parse_remote_id_test.go
@@ -1,0 +1,88 @@
+package ztpv6
+
+import (
+	"testing"
+
+	"github.com/insomniacslk/dhcp/dhcpv6"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCircuitID(t *testing.T) {
+	tt := []struct {
+		name    string
+		circuit string
+		want    *CircuitID
+		fail    bool
+	}{
+		{name: "Bogus string", circuit: "ope/1/2/3:ope", fail: true, want: nil},
+		{name: "Arista Port Vlan Pattern", circuit: "Ethernet13:2001", want: &CircuitID{Port: "13", Vlan: "2001"}},
+		{name: "Arista Slot Module Port Pattern", circuit: "Ethernet1/3/4", want: &CircuitID{Slot: "1", Module: "3", Port: "4"}},
+	}
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			circuit, err := matchCircuitId(tc.circuit)
+			if err != nil && !tc.fail {
+				t.Errorf("unexpected failure: %v", err)
+			}
+			if circuit != nil {
+				require.Equal(t, *tc.want, *circuit, "comparing remoteID data")
+			}
+		})
+	}
+}
+
+func TestFormatCircuitID(t *testing.T) {
+	tt := []struct {
+		name    string
+		circuit *CircuitID
+		want    string
+		fail    bool
+	}{
+		{name: "empty", circuit: &CircuitID{}, want: ",,,,"},
+		{name: "Arista format Port/Vlan", circuit: &CircuitID{Port: "13", Vlan: "2001"}, want: ",,13,,2001"},
+		{name: "Arista format Slot/Module/Port", circuit: &CircuitID{Slot: "1", Module: "3", Port: "4"}, want: "1,3,4,,"},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			circuit := tc.circuit.FormatCircuitID()
+			require.Equal(t, tc.want, circuit, "FormatRemoteID data")
+		})
+	}
+
+}
+
+func TestParseRemoteID(t *testing.T) {
+	tt := []struct {
+		name    string
+		circuit []byte
+		want    *CircuitID
+		fail    bool
+	}{
+		{name: "Bogus string", circuit: []byte("ope/1/2/3:ope.1"), fail: true, want: nil},
+		{name: "Arista Port Vlan Pattern", circuit: []byte("Ethernet13:2001"), want: &CircuitID{Port: "13", Vlan: "2001"}},
+		{name: "Arista Slot Module Port Pattern", circuit: []byte("Ethernet1/3/4"), want: &CircuitID{Slot: "1", Module: "3", Port: "4"}},
+	}
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			packet, err := dhcpv6.NewMessage()
+			if err != nil {
+				t.Fatalf("failed to creat dhcpv6 packet object: %v", err)
+			}
+			opt := dhcpv6.OptRemoteId{}
+			opt.SetRemoteID(tc.circuit)
+			opt.SetEnterpriseNumber(1234)
+			packet.AddOption(&opt)
+
+			circuit, err := ParseRemoteId(packet)
+			if err != nil && !tc.fail {
+				t.Errorf("unexpected failure: %v", err)
+			}
+			if circuit != nil {
+				require.Equal(t, *tc.want, *circuit, "ZTPRemoteID data")
+			} else {
+				require.Equal(t, tc.want, circuit, "ZTPRemoteID data")
+			}
+		})
+	}
+}
--- a/dhcpv6/ztpv6/parse_vendor_options.go
+++ b/dhcpv6/ztpv6/parse_vendor_options.go
@@ -1,0 +1,60 @@
+package ztpv6
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/insomniacslk/dhcp/dhcpv6"
+)
+
+var (
+	errVendorOptionMalformed = errors.New("malformed vendor option")
+)
+
+// VendorData contains fields extracted from Option 17 data
+type VendorData struct {
+	VendorName, Model, Serial string
+}
+
+// ParseVendorData will try to parse dhcp6 Vendor Specific Information options data
+// looking for more specific vendor data (like model, serial number, etc).
+// If the options are missing we will just return nil
+func ParseVendorData(packet dhcpv6.DHCPv6) (*VendorData, error) {
+	opt := packet.GetOneOption(dhcpv6.OptionVendorOpts)
+	if opt == nil {
+		return nil, errors.New("vendor options not found")
+	}
+
+	vd := VendorData{}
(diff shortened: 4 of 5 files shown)


dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,74 @@
...
+			if len(p) < 4 {
+				return nil, errVendorOptionMalformed
+			}
+
+			vd.VendorName = p[0]
+			vd.Model = p[1]
+			vd.Serial = p[3]
+			vendData = append(vendData, vd)
+
+			// ZPESystems:NSC:000000000
```
pmazzini: Indentation seems wrong here.

dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,74 @@
+package ztpv6
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/golang/glog"
```
pmazzini: In the library we are using the regular `log` module instead of `glog`.

dhcpv6/ztp/parse_vendor_options_test.go
```diff
@@ -0,0 +1,53 @@
+package ztpv6
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
```
pmazzini: In the library we are using `stretchr/testify` for tests.

dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,74 @@
...
+var errVendorOptionMalformed = errors.New("malformed vendor option")
+
+type VendorData struct {
+	VendorName, Model, Serial string
+}
+
+// VendorDataV6 will try to parse dhcp6 Vendor Specific Information options data
+// looking for more specific vendor data (like model, serial number, etc).
+// If the options are missing we will just return an empty
+func VendorDataV6(packet dhcpv6.DHCPv6) ([]VendorData, error) {
```
pmazzini: For consistency with the ztpv4 module lets call this `ParseVendorData`. The V6 is redundant since it is part of the dhcpv6 package already.

pmazzini: Do we need separate functions for `VendorDataV6` and `parseV6VendorSpecificInfo`? Maybe we can combine them.

pmazzini: Why does it returns `[]VendorData`? Shouldn't it return one VendorData per packet?

pmazzini:
Are we going to add ztpv6 support for ZPE? Otherwise I would only add what we need for now.

Will we need to extract remote id and circuit id as well? Should that be added to the library?

dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,74 @@
+package ztpv6
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/gorhamc/dhcp/dhcpv6"
```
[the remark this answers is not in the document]
gorhamc: Whoops! That was certainly not intentional. Good catch!

dhcpv6/ztp/parse_vendor_options_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/gorhamc/dhcp/dhcpv6"
```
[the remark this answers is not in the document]
gorhamc: Same. This was not intentional. Good catch!

dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,72 @@
...
+
+// VendData contains fields extracted from Option 17 data
+type VendData struct {
+	VendorName, Model, Serial string
+}
+
+// VendorData will try to parse dhcp6 Vendor Specific Information options data
+// looking for more specific vendor data (like model, serial number, etc).
+// If the options are missing we will just return an empty struct
+func VendorData(packet dhcpv6.DHCPv6) (VendData, error) {
```
pmazzini: For consistency with the ztpv4 module lets call this `ParseVendorData`. Do we need separate functions for `VendorData` and `parseV6VendorSpecificInfo`? Maybe we can combine them.

pmazzini: How about returning a reference to the `VendData` to avoid copying the whole struct?

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,80 @@
...
+type CircuitID struct {
+	Slot    string
+	Module  string
+	Port    string
+	SubPort string
+	Vlan    string
+}
+
+// ZTPRemoteId will parse the RemoteId Option data for Vendor Specific data
+func ZTPRemoteId(packet dhcpv6.DHCPv6) (CircuitID, error) {
```
pmazzini: How about returning a reference to the C`ircuitID` to avoid copying the whole struct?

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,80 @@
...
+			circuit.SubPort = match
+		case "vlan":
+			circuit.Vlan = match
+		}
+	}
+	return circuit, nil
+}
+
+// FormatCircuitID is the CircuitID format we send in our Bootfile URL for ZTP devices
+func FormatCircuitID(circuit CircuitID) string {
```
pmazzini: How about using a reference of the `CircuitID` to avoid copying the whole struct?

pmazzini:
Is the idea to add support for both ZPE and Arista? Does the `ZTPRemoteId` function supports both of them?

gorhamc:
Yeah we will want to support ZPE and Arista. The ZTPRemoteId function will support both because the relay agent for ZPE is through an Arista which is the current interface format I'm matching. There will be more patterns whenever we get around to adding Juniper support, but this should cover us for now.

dhcpv6/ztp/parse_remote_id.go
gorhamc: I like it, I'll update, along with the above suggested comments.

pmazzini:
Ok, great, thanks for the clarification. Are the remote-id patterns standard or do they depend on a switch config?

gorhamc:
The remote Id patterns depend on vendor and hardware platform

pmazzini:
There are still some unaddressed comments. Is it ready for review?

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// ParseZTPRemoteId will parse the RemoteId Option data for Vendor Specific data
+func ParseZTPRemoteId(packet dhcpv6.DHCPv6) (*CircuitID, error) {
+	// Need to decapsulate the packet after multiple relays in order to reach RemoteId data
+	inner, err := dhcpv6.DecapsulateRelayIndex(packet, -1)
+	if err != nil {
+		return &CircuitID{}, fmt.Errorf("failed to decapsulate relay index: %v", err)
+	}
+	var remoteID string
```
pmazzini: Is there a need for declaring this variable here? How about doing a short variable declaration at line 37?

dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,58 @@
...
+
+			vd.VendorName = p[0]
+			vd.Model = p[1]
+			vd.Serial = p[3]
+
+			// ZPESystems:NSC:000000000
+		case strings.HasPrefix(optData, "ZPESystems:"):
+			p := strings.Split(optData, ":")
+			if len(p) < 3 {
+				return &VendorData{}, errVendorOptionMalformed
```
pmazzini: How about returning a nil `VendorData` here?

dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,58 @@
...
+	vo := opt.(*dhcpv6.OptVendorOpts).VendorOpts
+
+	for _, opt := range vo {
+		optData := string(opt.(*dhcpv6.OptionGeneric).OptionData)
+		switch {
+		// Arista;DCS-0000;00.00;ZZZ00000000
+		case strings.HasPrefix(optData, "Arista;"):
+			p := strings.Split(optData, ";")
+			if len(p) < 4 {
+				return &VendorData{}, errVendorOptionMalformed
```
pmazzini: How about returning a nil `VendorData` here?

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,80 @@
...
+		remoteID = string(rid.(*dhcpv6.OptRemoteId).RemoteID())
+		cid, err = parseZTPRemoteID(remoteID)
+		if err != nil {
+			return &CircuitID{}, err
+		}
+	}
+	return cid, nil
+}
+
+func parseZTPRemoteID(remoteID string) (*CircuitID, error) {
```
pmazzini: Both functions have the same name, one exported and the other one not, I am not a great fan of this. How about changing its name?

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,80 @@
...
+// ParseZTPRemoteId will parse the RemoteId Option data for Vendor Specific data
+func ParseZTPRemoteId(packet dhcpv6.DHCPv6) (*CircuitID, error) {
+	// Need to decapsulate the packet after multiple relays in order to reach RemoteId data
+	inner, err := dhcpv6.DecapsulateRelayIndex(packet, -1)
+	if err != nil {
+		return &CircuitID{}, fmt.Errorf("failed to decapsulate relay index: %v", err)
+	}
+	var remoteID string
+	var cid *CircuitID
+	rid := inner.GetOneOption(dhcpv6.OptionRemoteID)
```
pmazzini: This could be a single line: `if rid := inner.GetOneOption(dhcpv6.OptionRemoteID); rid != nil {`

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,80 @@
...
+		cid, err = parseZTPRemoteID(remoteID)
+		if err != nil {
+			return &CircuitID{}, err
+		}
+	}
+	return cid, nil
+}
+
+func parseZTPRemoteID(remoteID string) (*CircuitID, error) {
+	var names []string
```
pmazzini: This could be a single line: `var names, matches []string`

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,80 @@
...
+	SubPort string
+	Vlan    string
+}
+
+// ParseZTPRemoteId will parse the RemoteId Option data for Vendor Specific data
+func ParseZTPRemoteId(packet dhcpv6.DHCPv6) (*CircuitID, error) {
+	// Need to decapsulate the packet after multiple relays in order to reach RemoteId data
+	inner, err := dhcpv6.DecapsulateRelayIndex(packet, -1)
+	if err != nil {
+		return &CircuitID{}, fmt.Errorf("failed to decapsulate relay index: %v", err)
```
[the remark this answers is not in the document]
gorhamc: Good call. Not sure why I didn't do that in the first place.

dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,60 @@
...
+			if len(p) < 4 {
+				return nil, errVendorOptionMalformed
+			}
+
+			vd.VendorName = p[0]
+			vd.Model = p[1]
+			vd.Serial = p[3]
+			return &vd, nil
+
+			// ZPESystems:NSC:000000000
```
pmazzini: Indentation seems wrong here.

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,85 @@
...
+		case "vlan":
+			circuit.Vlan = match
+		}
+	}
+
+	return &circuit, nil
+}
+
+// FormatCircuitID is the CircuitID format we send in our Bootfile URL for ZTP devices
+func FormatCircuitID(circuit CircuitID) string {
```
pmazzini: How about using a reference of the `CircuitID` to avoid copying the whole struct?

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,85 @@
...
+	switch {
+	case aristaPVPattern.MatchString(remoteID):
+		matches = aristaPVPattern.FindStringSubmatch(remoteID)
+		names = aristaPVPattern.SubexpNames()
+	case aristaSMPPattern.MatchString(remoteID):
+		matches = aristaSMPPattern.FindStringSubmatch(remoteID)
+		names = aristaSMPPattern.SubexpNames()
+	}
+
+	fmt.Println(matches)
```
pmazzini: Is this a leftover from debugging?

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,85 @@
...
+		circ, err := circuitId(remoteID)
+		if err != nil {
+			return nil, err
+		}
+		return circ, nil
+	}
+	return nil, errors.New("failed to parse RemoteID option data")
+}
+
+func circuitId(remoteID string) (*CircuitID, error) {
```
pmazzini: Maybe rename this to `matchCircuitId` or `createCircuitId` to avoid having the same name as the type.

dhcpv6/ztp/parse_vendor_options.go
```diff
@@ -0,0 +1,60 @@
...
+	VendorName, Model, Serial string
+}
+
+// ParseVendorData will try to parse dhcp6 Vendor Specific Information options data
+// looking for more specific vendor data (like model, serial number, etc).
+// If the options are missing we will just return an empty struct
+func ParseVendorData(packet dhcpv6.DHCPv6) (*VendorData, error) {
+	opt := packet.GetOneOption(dhcpv6.OptionVendorOpts)
+	if opt == nil {
+		return &VendorData{}, nil
```
pmazzini: Should we return an error here and a nil `VendorData`? Returning an empty `VendorData` is probably not very useful.

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,84 @@
...
+type CircuitID struct {
+	Slot    string
+	Module  string
+	Port    string
+	SubPort string
+	Vlan    string
+}
+
+// ParseZTPRemoteId will parse the RemoteId Option data for Vendor Specific data
+func ParseZTPRemoteId(packet dhcpv6.DHCPv6) (*CircuitID, error) {
```
pmazzini: Should `ZTP` be removed from the name of this function? I find it a bit redundant since that is already in the name of the package.

pmazzini:
Some minor NITs but looks good in general.

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,84 @@
...
+		case "vlan":
+			circuit.Vlan = match
+		}
+	}
+
+	return &circuit, nil
+}
+
+// FormatCircuitID is the CircuitID format we send in our Bootfile URL for ZTP devices
+func FormatCircuitID(circuit *CircuitID) string {
```
pmazzini: What do you think of making this a method of `CircuitID`? Similar to https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/duid.go#L88

pmazzini:
I am trying to think a little bit ahead how these functions are going to be used. Should `ParseVendorData` call `ParseRemoteId` or better to have them as separate functions?

dhcpv6/ztp/parse_remote_id.go
gorhamc: I like it, good idea!

gorhamc:
I struggled with the same question actually. The reason I split them up was because the remote id isn't specific to the vendor data of the ztp device, but is related to the relay agent's vendor/hw so I decided to keep them separate. The RemoteID data is really only used to add into the Bootfile URL, whereas the Vendor Data will be used in a number of places (BootURL, Serf Lookups, ZTP Gating).

However, I'm not opposed to having Vendor Data call Remote ID parsing and populate a circuit ID field or something. I've honestly gone back and forth on this a number of times. If you have a preference, I will implement it.

pmazzini:
Sounds good to me, lets leave it that way, thanks for explaining it. If needed we can change it afterwards.

insomniacslk:
I'm looking at this today - meanwhile I see that there are merge conflicts that should be resolved first

pmazzini:
Thanks! Are there merge conflicts? They are all new files and a new directory.

insomniacslk:
I see this on the github UI:

```
This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.
```

insomniacslk:
Maybe the README?

pmazzini:
Interesting, the README is also a new file inside the ztp directory.

insomniacslk:
I found why - in the individual commits, README is created twice. @gorhamc should squash all the commits in one to remove that duplication

dhcpv6/ztp/parse_remote_id.go
```diff
@@ -0,0 +1,84 @@
+package ztpv6
```
insomniacslk: package name is different from directory name

dhcpv6/ztp/README.md
```diff
@@ -0,0 +1,16 @@
+# Zero Touch Provisioning (ZTP) DHCPv6 Parsing for Network Hardware Vendors
+
+## Currently Supported Vendors For DHCPv6 ZTP
+ - Arista
+ - ZPE
+
+## Why Do We Need This?
+Many network hardware vendors support features that allow network devices to provision themselves with proper supporting automation/tools. Network devices can rely on DHCP and other methods to gather bootfile info, IP's, etc. DHCPv6 Vendor options provides us Vendor Name, Make, Model, and Serial Number data. This data can be used to uniquely identify individual network devices at provisioning time and can be used by tooling to make decisions necessary to correctly and reliably provision a network device.
```
insomniacslk:
```suggestion
Many network hardware vendors support features that allow network devices to provision themselves with proper supporting automation/tools. Network devices can rely on DHCP and other methods to gather bootfile info, IPs, etc. DHCPv6 Vendor options provides us Vendor Name, Make, Model, and Serial Number data. This data can be used to uniquely identify individual network devices at provisioning time and can be used by tooling to make decisions necessary to correctly and reliably provision a network device.
```

dhcpv6/ztp/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+For more detail on how this is used at Facebook you can check out our ZTP Wiki article, [Scaling Backbone Networks Through Zero Touch Provisioning](https://code.fb.com/networking-traffic/scaling-the-facebook-backbone-through-zero-touch-provisioning/)
```
insomniacslk: is this relevant? This is not a FB-specific library. The article link is still useful, so it should be slightly  rephrased (I added a suggestion)

```suggestion
For more details on a large-scale ZTP deployment, check out how this is done at Facebook, [Scaling Backbone Networks Through Zero Touch Provisioning](https://code.fb.com/networking-traffic/scaling-the-facebook-backbone-through-zero-touch-provisioning/).
```

dhcpv6/ztp/parse_remote_id.go
gorhamc: Do you have a preference on ztpv6 vs ztp for directory/package name? I changed package to ztp in my last update but that may not be the best idea.

pmazzini: I would use `ztpv6` to differentiate from the v4 package.

insomniacslk: I agree that it would be confusing to have the same name for both. At the same time I don't like having "ztp" for v4 and "ztpv6" for v6, because it's inconsistent

insomniacslk: However I don't see an easy solution without renaming ztp to ztpv6

PR #209, opened 2018-11-26
Add missing error check in GenerateTransactionID

earlgreyz:
Added missing error check, which might return more descriptive error to the user when generation error occurs (alternatively we can change the `Read` line to avoid ineffectual assignment)

```go
n, _ := rand.Read(tidBytes)
```

Additionally fixed the tests to work on both LE and BE (previously a BE was assumed).

### DIFF
--- a/dhcpv6/dhcpv6message.go
+++ b/dhcpv6/dhcpv6message.go
@@ -39,15 +39,18 @@ func GenerateTransactionID() (*uint32, error) {
 	for {
 		tidBytes := make([]byte, 4)
 		n, err := rand.Read(tidBytes)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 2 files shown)


dhcpv6/dhcpv6message.go
```diff
@@ -47,7 +50,7 @@ func GenerateTransactionID() (*uint32, error) {
 			return nil, err
 		}
 		if tid == nil {
-			return nil, fmt.Errorf("Error: got a nil Transaction ID")
+			return nil, fmt.Errorf("Got a nil Transaction ID")
```
insomniacslk: since we are here, make it lowercase as per https://github.com/golang/go/wiki/CodeReviewComments#error-strings

PR #211, opened 2018-11-27
Test TransactionID generation

earlgreyz:
Allow for mocking a `"crypto/rand"` function in dhcpv6 and improve unit tests.

`randomReadMock` is a standalone function without `mock.Mock` from the testify suite due to multiple errors of `testify/mock` library not properly supporting type aliases (`byte = uint8`) https://github.com/stretchr/testify/issues/692, and the same occurs when using `mock.Anything` (for `[]byte`).

### DIFF
--- a/dhcpv6/dhcpv6_test.go
+++ b/dhcpv6/dhcpv6_test.go
@@ -1,12 +1,15 @@
 package dhcpv6
 
 import (
+	"crypto/rand"
 	"encoding/binary"
+	"errors"
 	"net"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


earlgreyz:
Suite takes care of cleaning up the function pointer, since it's a global var I want to ensure it's not affecting any other test that might use the actual generator. If it wasn't for the suite we'd need a manual `recover` function (when a `require.*` is not met since it stops the function execution and the function pointer wouldn't have been replaced back).

PR #217, opened 2018-11-29
interfaces: added package with interface facilities

insomniacslk:
There are several places that may benefit from a package with interface facilities, see  https://github.com/systemboot/systemboot/pull/73 for context

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ The library is split into several parts:
 * `iana`: several IANA constants, and helpers used by `dhcpv6` and `dhcpv4`
 * `rfc1035label`: simple implementation of RFC1035 labels, used by `dhcpv6` and
   `dhcpv4`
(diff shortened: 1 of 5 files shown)


pmazzini:
The same changes will need to be applied to `dhcpv4/server_test.go`: https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/server_test.go#L96

insomniacslk:
> The same changes will need to be applied to `dhcpv4/server_test.go`: https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/server_test.go#L96

silly me, I didn't add it to the commit. Added now

pmazzini:
Should the `README.md` also be updated if a new package is added?

insomniacslk:
> Should the `README.md` also be updated if a new package is added?

Done

PR #219, opened 2018-12-08
Added netconf v4 tests and required modifiers

insomniacslk:
Same as https://github.com/insomniacslk/dhcp/pull/218 , but for DHCPv4.

I have also simplified the corresponding v6 modifiers, they were unnecessarily complex in #218.

### DIFF
--- a/dhcpv6/modifiers.go
+++ b/dhcpv6/modifiers.go
@@ -69,7 +69,11 @@ func WithArchType(at iana.ArchType) Modifier {
 // options
 func WithIANA(addrs ...OptIAAddress) Modifier {
 	return func(d DHCPv6) DHCPv6 {
-		iaNa := &OptIANA{}
+		opt := d.GetOneOption(OptionIANA)
+		if opt == nil {
+			opt = &OptIANA{}
+		}
+		iaNa := opt.(*OptIANA)
 		for _, addr := range addrs {
 			iaNa.AddOption(&addr)
 		}
--- a/dhcpv6/modifiers.go
+++ b/dhcpv6/modifiers.go
@@ -69,11 +69,7 @@ func WithArchType(at iana.ArchType) Modifier {
 // options
 func WithIANA(addrs ...OptIAAddress) Modifier {
 	return func(d DHCPv6) DHCPv6 {
-		opt := d.GetOneOption(OptionIANA)
-		if opt == nil {
-			opt = &OptIANA{}
-		}
-		iaNa := opt.(*OptIANA)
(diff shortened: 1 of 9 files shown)


dhcpv6/modifiers.go
```diff
@@ -69,11 +69,7 @@ func WithArchType(at iana.ArchType) Modifier {
 // options
 func WithIANA(addrs ...OptIAAddress) Modifier {
 	return func(d DHCPv6) DHCPv6 {
-		opt := d.GetOneOption(OptionIANA)
-		if opt == nil {
-			opt = &OptIANA{}
-		}
-		iaNa := opt.(*OptIANA)
+		iaNa := &OptIANA{}
```
pmazzini: The simplification of the v6 modifiers look good to me except for the `WithIANA` one. Ideally when adding a IAAddr if there was a previous IANA option we should preserve the IaId which we are missing in this case.

pmazzini:
Except for the `WithIANA` modifier, the rest looks good to me.

dhcpv6/modifiers.go
insomniacslk: I thought about that, and it's the reason why I didn't simplify other functions in that file. However I think we should use different names for modifiers that add or that replace, to avoid having misleading names. E.g. "WithIANA" in my mind sounds like something that sets IA_NA to what I want regardless of what's there already. We could use "Add*" instead of "With*" for the other use case

pmazzini: > it's the reason why I didn't simplify other functions in that file

One of the modifiers which was not changed is `WithNetboot` which I find it similar to this one since it adds/modifies only one of the suboptions of OPTION_ORO.

> However I think we should use different names for modifiers that add or that replace, to avoid having misleading names.

I am focusing on what the modifier is modifying/replacing. In this case the modifier only takes as parameter one of the IA_NA suboptions fields, so I wouldn't replace the parent option in that case.
 
> "WithIANA" in my mind sounds like something that sets IA_NA to what I want regardless of what's there already.

The problem is that IA_NA is a complex option which also has nested options and this modified `WithIANA` doesn't let you completely manipulate all the fields of that option. It is only receiving a IAAddr which is one of the suboptions of IA_NA. In practice I think this modifier is not super useful and would break things more than simplifying them.

Given the parameters that it takes I think it should be renamed to `WithIAAddress`.

insomniacslk: At the same time, what do you think of using more consistent names for what replaces and what adds stuff?

PR #221, opened 2018-12-22
[ztpv4] make it consistent with ztpv6

pmazzini:
* Rename the directory from `ztp` to `ztpv4` to match the package name
* Return an error when calling `ParseVendorData` and no vendor data is found

### DIFF
--- a/dhcpv4/ztpv4/ztp.go
+++ b/dhcpv4/ztpv4/ztp.go
@@ -10,9 +10,7 @@ import (
 // VendorData is optional data a particular vendor may or may not include
 // in the Vendor Class options.
 type VendorData struct {
-	VendorName string
-	Model      string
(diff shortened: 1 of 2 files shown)


dhcpv4/ztpv4/ztp.go
```diff
@@ -76,5 +74,5 @@ func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
 	}
 
 	// We didn't match anything.
-	return nil, nil
+	return nil, errors.New("failed to parse vendor option data")
```
insomniacslk: nit: this is not a parsing failure, rather no known ZTP vendor is found, I would change the error string

insomniacslk:
thanks for this PR!

PR #222, opened 2018-12-29
dhcpv4: nicer parsing of messages.

hugelgupf:
All tests pass locally, as far as I can tell.

Less code.

API changes to make easier to use:
- No setters and getters on DHCPv4 packet anymore, just access fields
  directly.
- TransactionID changed from uint32 -> [4]byte to avoid parsing it at
  all (fixed-length array can be used as a map key).
- ServerHostName and BootFileName are translated to the correct type
  while parsing.
- Client HW Addr len does not have to be set manually anymore, is
  derived from the length of the byte slice (and automatically set when
  parsing).

Less error prone:
- IPs in packet can be nil or 0.0.0.0, get serialized all the same.
- MessageType returns 0 rather than nil, which was unchecked in netboot
  code.
- Setters and Getters eliminated in favor of exported member names,
  since there was no validation of values anyway.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -100,7 +100,7 @@ func GenerateTransactionID() (*TransactionID, error) {
 	var b TransactionID
 	n, err := rand.Read(b[:])
 	if n != 4 {
-		return nil, fmt.Errorf("invalid random sequence: smaller than 32 bits")
+		return nil, errors.New("invalid random sequence: smaller than 32 bits")
 	}
 	if err != nil {
 		return nil, err
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -10,6 +10,20 @@ import (
 	"strings"
 
 	"github.com/insomniacslk/dhcp/iana"
+	"github.com/u-root/u-root/pkg/uio"
+)
+
+const (
+	minPacketLen = 236
+
+	// Maximum length of the ClientHWAddr (client hardware address) according to
+	// RFC 2131, Section 2. This is the link-layer destination a server
+	// must send responses to.
+	chaddrLen = 16
+
+	// flagBroadcast is the broadcast bit in the flag field as defined by
+	// RFC 2131, Section 2, Figure 2.
+	flagBroadcast = 1 << 15
 )
 
 // HeaderSize is the DHCPv4 header size in bytes.
@@ -21,21 +35,20 @@ const MaxMessageSize = 576
 // DHCPv4 represents a DHCPv4 packet header and options. See the New* functions
 // to build DHCPv4 packets.
 type DHCPv4 struct {
-	opcode         OpcodeType
-	hwType         iana.HwTypeType
-	hwAddrLen      uint8
-	hopCount       uint8
-	transactionID  uint32
-	numSeconds     uint16
-	flags          uint16
-	clientIPAddr   net.IP
-	yourIPAddr     net.IP
-	serverIPAddr   net.IP
-	gatewayIPAddr  net.IP
-	clientHwAddr   [16]byte
-	serverHostName [64]byte
-	bootFileName   [128]byte
-	options        []Option
+	OpCode         OpcodeType
+	HWType         iana.HwTypeType
+	HopCount       uint8
+	TransactionID  [4]byte
+	NumSeconds     uint16
+	Flags          uint16
+	ClientIPAddr   net.IP
+	YourIPAddr     net.IP
+	ServerIPAddr   net.IP
+	GatewayIPAddr  net.IP
+	ClientHWAddr   net.HardwareAddr
+	ServerHostName string
+	BootFileName   string
+	Options        []Option
 }
 
 // Modifier defines the signature for functions that can modify DHCPv4
@@ -83,17 +96,16 @@ func GetExternalIPv4Addrs(addrs []net.Addr) ([]net.IP, error) {
 
 // GenerateTransactionID generates a random 32-bits number suitable for use as
 // TransactionID
-func GenerateTransactionID() (*uint32, error) {
-	b := make([]byte, 4)
-	n, err := rand.Read(b)
+func GenerateTransactionID() (*[4]byte, error) {
+	var b [4]byte
+	n, err := rand.Read(b[:])
 	if n != 4 {
-		return nil, errors.New("Invalid random sequence: smaller than 32 bits")
+		return nil, fmt.Errorf("invalid random sequence: smaller than 32 bits")
 	}
 	if err != nil {
 		return nil, err
 	}
-	tid := binary.LittleEndian.Uint32(b)
-	return &tid, nil
+	return &b, nil
 }
 
 // New creates a new DHCPv4 structure and fill it up with default values. It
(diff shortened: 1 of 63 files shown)


hugelgupf:
This looks like a huge PR now, but most of these changes remove boilerplate code. This *should be really easy* to review.

insomniacslk:
Thanks @hugelgupf ! We needed this change for a long time :) For context: the v6 library is similar to what you did here, with no setters/getters and a simpler API in general. There should be an issue to track the things we want to "backport" from v6 to v4, but I cannot find it now.
This PR is also a necessary step to release a version 1.0.0 with a stable API that is consistent across v6 and v4

insomniacslk:
CC @pmazzini @get9 @earlgreyz , FYI for the relevant parts that are touched in this PR

dhcpv4/option_archtype.go
```diff
@@ -53,24 +54,11 @@ func (o *OptClientArchType) String() string {
...
-		b := data[2+idx : 2+idx+2]
-		archTypes = append(archTypes, iana.ArchType(binary.BigEndian.Uint16(b)))
+	buf := uio.NewBigEndianBuffer(data)
+
+	archTypes := make([]iana.ArchType, 0, buf.Len()/2)
+	for buf.Has(2) {
+		archTypes = append(archTypes, iana.ArchType(buf.Read16()))
 	}
-	return &OptClientArchType{ArchTypes: archTypes}, nil
+	return &OptClientArchType{ArchTypes: archTypes}, buf.FinError()
```
insomniacslk: the RFC 4578 section 2.1 requires that at least one arch type is present. Here (or above) we should return an error if `len(archTypes)` is 0

dhcpv4/dhcpv4.go
```diff
@@ -21,21 +35,20 @@ const MaxMessageSize = 576
...
-	flags          uint16
-	clientIPAddr   net.IP
-	yourIPAddr     net.IP
-	serverIPAddr   net.IP
-	gatewayIPAddr  net.IP
-	clientHwAddr   [16]byte
-	serverHostName [64]byte
-	bootFileName   [128]byte
-	options        []Option
+	OpCode         OpcodeType
```
earlgreyz: For consistency maybe `Opcode`

dhcpv4/dhcpv4.go
```diff
@@ -83,17 +96,16 @@ func GetExternalIPv4Addrs(addrs []net.Addr) ([]net.IP, error) {
...
 // TransactionID
-func GenerateTransactionID() (*uint32, error) {
-	b := make([]byte, 4)
-	n, err := rand.Read(b)
+func GenerateTransactionID() (*[4]byte, error) {
+	var b [4]byte
+	n, err := rand.Read(b[:])
 	if n != 4 {
-		return nil, errors.New("Invalid random sequence: smaller than 32 bits")
+		return nil, fmt.Errorf("invalid random sequence: smaller than 32 bits")
```
earlgreyz: the message is not formatted so `errors.New` is a better choice here

dhcpv4/options.go
```diff
@@ -85,70 +89,82 @@ func ParseOption(data []byte) (Option, error) {
...
+
+		// N bytes: option data
+		data := buf.Consume(length)
+		if data == nil {
+			return nil, fmt.Errorf("error collecting options: %v", buf.Error())
 		}
-		options = append(options, opt)
-		if opt.Code() == OptionEnd {
-			break
+		data = data[:length:length]
```
pmazzini: Is this different from `data = data[:length]`?

dhcpv4/dhcpv4.go
```diff
@@ -10,6 +10,20 @@ import (
 	"strings"
 
 	"github.com/insomniacslk/dhcp/iana"
+	"github.com/u-root/u-root/pkg/uio"
```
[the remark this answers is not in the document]
pmazzini: I am ok with it, it simplifies a lot the parsing.

dhcpv4/dhcpv4.go
hugelgupf: I'm planning on renaming the types to remove the "Type" suffix as well, because that's kind of unlike Go. And HwTypeType should really be HWType, e.g.

dhcpv4/dhcpv4.go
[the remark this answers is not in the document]
hugelgupf: I can guarantee uio will always be a self-contained library.

dhcpv4/options.go
hugelgupf: Yes -- data is a slice into the overall array containing the packet. If you were to append more bytes here, you'd write over the packet if there are more bytes left in the packet. So you have to cap the capacity as well as the length.

hugelgupf:
See https://github.com/rtr7/router7/pull/19 for CI breakage.

Dependencies should use modules... then you don't have to guarantee API compatibility.

insomniacslk:
reviewing commit by commit as suggested - the first one (nicer parsing of messages) looks good to me

dhcpv4/dhcpv4.go
```diff
@@ -371,9 +371,32 @@ func (d *DHCPv4) SetUnicast() {
...
+// Options is a collection of options.
+type Options []Option
+
+// GetOption will attempt to get all options that match a DHCPv4 option
+// from its OptionCode.  If the option was not found it will return an
+// empty list.
+//
+// According to RFC 3396, options that are specified more than once are
+// concatenated, and hence this should always just return one option.
+func (o Options) GetOption(code OptionCode) []Option {
```
insomniacslk: I'm sure I'll get confused by this `Options` vs `[]Option` in the future, but this is OK for me

dhcpv4/option_archtype.go
```diff
@@ -53,24 +54,11 @@ func (o *OptClientArchType) String() string {
...
-	if len(data) < 2+length {
-		return nil, ErrShortByteStream
-	}
-	archTypes := make([]iana.ArchType, 0, length%2)
-	for idx := 0; idx < length; idx += 2 {
-		b := data[2+idx : 2+idx+2]
-		archTypes = append(archTypes, iana.ArchType(binary.BigEndian.Uint16(b)))
+	buf := uio.NewBigEndianBuffer(data)
+
+	archTypes := make([]iana.ArchType, 0, buf.Len()/2)
```
insomniacslk: we still have the problem of not checking if there is at least one arch type here?

hugelgupf: later commit fixes that

dhcpv4/bsdp/types.go
```diff
@@ -26,9 +26,9 @@ const (
 	OptionMachineName                   dhcpv4.OptionCode = 130
 )
 
-// OptionCodeToString maps BSDP OptionCodes to human-readable strings
+// optionCodeToString maps BSDP OptionCodes to human-readable strings
 // describing what they are.
-var OptionCodeToString = map[dhcpv4.OptionCode]string{
+var optionCodeToString = map[dhcpv4.OptionCode]string{
```
get9: Ditto here as above.

insomniacslk:
As agreed in person let's split this PR to simplify the review,  because it's large and contains multiple unrelated changes

dhcpv4/bsdp/bsdp_option_message_type.go
```diff
@@ -20,14 +21,14 @@ const (
...
 func (m MessageType) String() string {
-	if s, ok := MessageTypeToString[m]; ok {
+	if s, ok := messageTypeToString[m]; ok {
 		return s
 	}
 	return "Unknown"
 }
 
-// MessageTypeToString maps each BSDP message type to a human-readable string.
-var MessageTypeToString = map[MessageType]string{
```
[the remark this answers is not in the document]
hugelgupf: Why was it exported to begin with? What user would need that?

dhcpv4/bsdp/types.go
hugelgupf: Ditto. I don't see why this needs to be in the API.

dhcpv4/dhcpv4.go
```diff
@@ -416,6 +416,52 @@ func (o Options) GetOneOption(code OptionCode) Option {
...
 }
 
+// Marshal writes options binary representations to b.
+func (o Options) Marshal(b *uio.Lexer, writeEnd bool) {
+	for _, opt := range o {
+		code := opt.Code()
+
+		// Even if the End option is in there, don't marshal it until
+		// the end.
+		if code == OptionEnd {
```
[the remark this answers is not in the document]
hugelgupf: I'll move this to a different PR, but yes, and early End should always be considered an anomaly.

The other side should never send you an early End. I mean, you wouldn't even *know* if it's an early End. You have no idea if the data after the End is garbage or not.

And a user of this library should never be given the opportunity to include an early End in the options list, because it's invalid anyway. A user of this library shouldn't be able to insert the End option at all, I'd assert, and the library itself should take care of that while marshaling.

dhcpv4/dhcpv4.go
```diff
@@ -511,29 +557,6 @@ func (d *DHCPv4) Summary() string {
 	return ret
 }
 
-// ValidateOptions runs sanity checks on the DHCPv4 packet and prints a number
-// of warnings if something is incorrect.
-func (d *DHCPv4) ValidateOptions() {
```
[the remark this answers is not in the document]
hugelgupf: Marshaling and unmarshaling will take care of OptionEnd, so no need to validate options - you can never get in this state anyway.

dhcpv4/dhcpv4.go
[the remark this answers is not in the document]
insomniacslk: Your reasoning is valid, but there is another one equally valid in my opinion - this library was initially inspired by `scapy`'s concepts, where you can intentionally craft malformed packets for security testing purposes, if you want to. Then sane (hopefully) defaults have been implemented for the regular user, and validation methods have been added for the paranoid

dhcpv4/dhcpv4.go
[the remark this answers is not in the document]
insomniacslk: see comment above for my point of view on validation methods

hugelgupf:
Closing in favor of #223, #224, and PRs yet-to-be-opened.

Issue #22, opened 2018-03-26
Uniform dhcpv4 and dhcpv6 Parse methods


insomniacslk:
The dhcpv4 and dhcpv6 packages use two different conventions for option parsing. In `dhcpv4` the option is parsed along with option code and length, in `dhcpv6` these are handled earlier, and the `Parse*` methods only handle the data, assuming that the option code is the same as the used option.

Let's uniform the two interfaces for consistency.

pmazzini:
Do you have any preference of one over the other?

insomniacslk:
I prefer the dhcpv6 convention but happy to discuss

PR #224, opened 2019-01-09
dhcpv4: simplify option parsing.

hugelgupf:
option's codes and lengths were being parsed twice: once in ParseOption
and once in each option type's Parse implementation. Consolidate such
that it only happens once.

Additionally, only pass data to options that they should parse -- we
know the length before the Parse function is called, so the option only
gets to see the data it needs to see.

Also, use uio.Lexer to simplify parsing code in general. Easier to read
and reason about.


hugelgupf:
This will have minor merge conflicts with #223. I'll resolve depending on whichever gets approved & submitted first.

dhcpv4/bsdp/bsdp_option_default_boot_image_id.go
```diff
@@ -4,35 +4,26 @@ import (
 	"fmt"
 
 	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/u-root/u-root/pkg/uio"
 )
 
+// OptDefaultBootImageID contains the selected boot image ID.
```
get9: This should be moved back down.

dhcpv4/bsdp/bsdp_option_machine_name.go
```diff
@@ -1,34 +1,21 @@
 package bsdp
 
 import (
-	"fmt"
-
 	"github.com/insomniacslk/dhcp/dhcpv4"
 )
 
+// OptMachineName represents a BSDP message type.
```
get9: Ditto this.

dhcpv4/bsdp/option_vendor_specific_information.go
```diff
@@ -11,20 +9,17 @@ import (
 // OptVendorSpecificInformation encapsulates the BSDP-specific options used for
 // the protocol.
 type OptVendorSpecificInformation struct {
-	Options []dhcpv4.Option
+	Options dhcpv4.Options
```
get9: Not sure that I like this particular type alias - what's wrong with just having it as a slice? it's more clear that it's a slice of a struct/interface, rather than having to look at what a `dhcpv4.Options` is.

dhcpv4/options.go
```diff
@@ -25,15 +36,10 @@ type Option interface {
...
-	if len(data) == 0 {
-		return nil, errors.New("invalid zero-length DHCPv4 option")
-	}
-	var (
-		opt Option
-		err error
-	)
-	switch OptionCode(data[0]) {
+func ParseOption(code OptionCode, data []byte) (Option, error) {
+	var opt Option
```
get9:
```
var (
    opt Option
    err error
)

File: dhcpv4/bsdp/bsdp_option_default_boot_image_id.go (hunk 1 above)
Comment (hugelgupf) [reply dhcpv4/bsdp/bsdp_option_default_boot_image_id.go]: That's not how Go comments work?

The diff may make it weird to read, but this is

```go
// OptDefaultBootImageID contains the selected boot image ID.
//
// Implements the BSDP option default boot image ID, which tells the client
// which image is the default boot image if one is not selected.
type OptDefaultBootImageID ...
```

dhcpv4/bsdp/option_vendor_specific_information.go
hugelgupf: So here's the deal. You've got the []dhcpv4.Option type in a couple of places: DHCPv4, OptRelayAgentInfo, and here.

A dhcpv4.Options type can be the one place to implement the following:
```go
func (Options) Get(...)
func (Options) Has(...)
func (Options) Add(...)
func (Options) Update(...)
```

rather than reimplementing GetOption and GetOneOption on multiple types and having the OptionGetter interface, just implement it on Options and spread Options around.

hugelgupf: I.e. there is no need to have both https://github.com/insomniacslk/dhcp/blob/108ed92e1c9901936541020bc3214533acce77bb/dhcpv4/bsdp/option_vendor_specific_information.go#L135 and https://github.com/insomniacslk/dhcp/blob/108ed92e1c9901936541020bc3214533acce77bb/dhcpv4/dhcpv4.go#L366

and eventually you'll start replicating HasOption and UpdateOption and AddOption to OptVendorSpecificInformation as well as the relay agent stuff... so an Options type saves you that work.

dhcpv4/bsdp/types.go
```diff
@@ -26,9 +26,9 @@ const (
 	OptionMachineName                   dhcpv4.OptionCode = 130
 )
 
-// OptionCodeToString maps BSDP OptionCodes to human-readable strings
+// optionCodeToString maps BSDP OptionCodes to human-readable strings
 // describing what they are.
-var OptionCodeToString = map[dhcpv4.OptionCode]string{
+var optionCodeToString = map[dhcpv4.OptionCode]string{
```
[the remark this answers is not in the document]
hugelgupf: OptionCode.String()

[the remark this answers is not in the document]
hugelgupf: Argh, yeah, for bsdp this is kinda weird. There's a commit way down the line to make the printing right, but for now, this is a weird thing to expose.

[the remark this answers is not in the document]
hugelgupf: I suppose I should try to fix this in this PR. Let me see what I can do.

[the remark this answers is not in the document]
hugelgupf: Actually, in this PR it doesn't matter yet -- each Option in BSDP prints its own option name statically anyway: https://github.com/insomniacslk/dhcp/pull/224/files#diff-d173dc6740eaa1ad94d4f356b77a647eR32

pmazzini:
Thanks! This fixes #22. It looks good to me.

dhcpv4/bsdp/option_vendor_specific_information.go
insomniacslk: yeah I get how this works, and in an isolated context it makes perfectly sense from a compactness perspective. I think it introduces a bit more confusion, but it's probably acceptable compared to the benefits of a better / more compact interface. As long as we use always `Options` instead of `[]Option` I guess this is OK. If we don't like it at any stage it's always possible to go back. Let's go with `Options` now

dhcpv4/bsdp/types.go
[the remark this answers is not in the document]
insomniacslk: I hate BSDP -.-

insomniacslk:
Thanks a lot for this PR @hugelgupf !

PR #225, opened 2019-01-10
dhcpv4: simplify option marshaling

hugelgupf:
- Consolidate writing the option code and length to Options.Marshal
  rather than doing it in each individual option.
- Support RFC 3396.
- Use uio in marshaling code to simplify.
- Remove Option.Length() - unused and wrong to use anyway due to RFC 3396.

### DIFF
--- a/dhcpv4/bsdp/boot_image.go
+++ b/dhcpv4/bsdp/boot_image.go
@@ -1,7 +1,6 @@
 package bsdp
 
 import (
-	"encoding/binary"
 	"fmt"
 
 	"github.com/u-root/u-root/pkg/uio"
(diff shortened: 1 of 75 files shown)


insomniacslk:
can you rebase to remove the redundant commits?

hugelgupf:
rebased + increased test coverage (accidentally deleted a test)

hugelgupf:
Updated and updated description

dhcpv4/bsdp/boot_image.go
```diff
@@ -79,11 +79,10 @@ type BootImage struct {
 }
 
 // ToBytes converts a BootImage to a slice of bytes.
```
pmazzini: I am not sure why lint did't warn about the method comment. It should be updated.

dhcpv4/bsdp/boot_image.go
```diff
@@ -79,11 +79,10 @@ type BootImage struct {
...
-func (b *BootImage) ToBytes() []byte {
-	bytes := b.ID.ToBytes()
-	bytes = append(bytes, byte(len(b.Name)))
-	bytes = append(bytes, []byte(b.Name)...)
-	return bytes
+func (b BootImage) Marshal(buf *uio.Lexer) {
```
pmazzini: Why not using a reference of `BootImage`?

pmazzini:
Commented. The rest LGTM. Thanks!

dhcpv4/bsdp/boot_image.go
hugelgupf: b doesn't need to be modified - i think of this like a const in c++. compiler can still optimize it with a reference.

hugelgupf:
update. ptal

dhcpv4/bsdp/boot_image.go
pmazzini: I would use a reference even if it doesn't need to be modified but I don't have a strong opinion about it.

hugelgupf:
@pmazzini can you push the merge button, too? I don't have permission.

PR #226, opened 2019-01-10
Misc cleanup

hugelgupf:
Based on #225. Some random stuff.

PR #227, opened 2019-01-11
dhcpv4: nicer Options API.

hugelgupf:
From:
```go
  r := d.GetOneOption(OptionRouter).(*OptRouter).Routers
  d.AddOption(&OptRouter{Routers: []net.IP{net.IP{192, 168, 0, 1}}})
```

To:
```go
  r := GetRouter(d.Options)
  d.AddOption(OptRouter(net.IP{192, 168, 0, 1}, ...))
```

Reasoning:

Options, as contained in Vendor-Specific Options and Relay Agent Options and in a DHCPv4 packet itself are all laid out the same way. <byte code> <byte length> <option value>. It should stand to reason that there should be a type to operate on them, but the interpretation of the options must be up to the standard itself.

Hence, Options become the following map:
```go
type Options map[uint8][]byte
```

This removes interpretation of the uint8 option code, and the value itself.

It is then on the user to know where they got their Options map from (which should be easy - are you writing a DHCP or BSDP client? Is it the Relay Agent Options struct or the DHCP message one?)

Now, casting options to their specific type as we get them from the options slice also has a bit of a smell to it. When you do `packet.GetOption(OptionRouter)` you *know* exactly what type you're expecting when you write the code. So the compiler should as well.

Proposed: `GetRouter(packet.Options)`
Alternative considered: `packet.Options.GetRouter()`
Alternative considered: `packet.GetRouter()`

The other thing is that this library does not implement all possible option values. It literally can't: vendor-specific options are up to each individual vendor, and some companies use their own company-specific values inside those. So you have to give users of the library the ability to interpret an option on their own *without contributing back to the library*. Users should be able to write their own BSDP library without touching the original code. Hence, the external approach of `GetOptionThing(dhcpv4.Options)`.

This is based on #226, and still in progress. The last commit contains all the code. I'm hoping to kind of split that up, but it's proving to be hard.


hugelgupf:
Let's call this a proposal. No need to review the code yet.

@insomniacslk agreed to this in Slack a couple days ago. Any objections from the others?

get9:
Seems like a great alternative to me.

pmazzini:
In general, I like the idea. Some questions that came to mind are:

* Will using a map preserve the order of the options? Ideally, after doing a FromBytes/Unmashal followed by a ToBytes/Marshal we should get the same bytes. If the options are stored in a map, is this going to be the case?

* Before this change we were parsing all the options straight away, which we will stop doing after this PR. I am not sure what is the best approach, I only wanted to bring this up to know everybody's opinion about it. Not parsing them straight away obviously has its benefits.

* The function `GetRouter` will return the same either if there is no such option in the packet or if it is malformed. Do we want to distinguish those 2 different cases or not really?

hugelgupf:
> * Will using a map preserve the order of the options? Ideally, after doing a FromBytes/Unmashal followed by a ToBytes/Marshal we should get the same bytes. If the options are stored in a map, is this going to be the case?

I preserve reproducibility of packets by sorting the options by option code in the marshaled packet. We could preserve the order they were added in as well, but RFC 2131 requires that some options be marshaled before other options (e.g. Router and Subnet - Subnet must come first), and they happen to coincide with sorting by option code number.

> * Before this change we were parsing all the options straight away, which we will stop doing after this PR. I am not sure what is the best approach, I only wanted to bring this up to know everybody's opinion about it. Not parsing them straight away obviously has its benefits.

Given that we don't know *how* to parse options until we know their context (DHCP or BSDP or BOOTP or some other vendor-specific thing), I actually think it's required we not parse them. It may also save time -- options that you don't use never get unmarshaled. (Though performance is a secondary concern for me right now; there's probably other low-hanging fruit and we should profile instead of throwing out statements like that...)

> * The function `GetRouter` will return the same either if there is no such option in the packet or if it is malformed. Do we want to distinguish those 2 different cases or not really?

I'm somewhat deciding that based on each option on its own. When you call `GetRouter` as the client, is there really a difference between "no router specified" and "router parsing failed" to the client? Imho, there isn't - you just don't have a router in either case. Why you don't have it doesn't really leave you with any distinctive remedy options. If you need a router, you resend your original packet with the router in the parameter request list, and if you still don't get a valid one, then you fall over ¯\_(ツ)_/¯ Would client behavior change if it knew the difference? Imo, probably not.

There are some that will have a good reason to return either - the ones where the 0 value of the type isn't equivalent to what happens in an error. Or where "option is not present" and "option was 0" imply distinct behavior. For example:

https://godoc.org/github.com/u-root/dhcp4/dhcp4opts#GetMaximumDHCPMessageSize
https://godoc.org/github.com/u-root/dhcp4/dhcp4opts#GetIPAddressLeaseTime

So most of them will look like `GetRouter`, but some may not. I think it's an option-specific decision to make.

pmazzini:
@hugelgupf Thanks for your answers! The proposal looks good to me. Are you going to move forward with it or which is the plan for now?

hugelgupf:
Going to move forward, just have other stuff to do, too - it'll be picked
up again soon.

hugelgupf:
So...

This is ready for review.

I apologize for it being so huge. Unfortunately, the switch from one options API to the other is an all or nothing thing.

The one thing that still bothers me is the handling of the bsdp.Packet type. I think the type has to exist to make the String() and Summary() work right, but I'm not sure how to do it better. Ideas?

dhcpv4/bsdp/boot_image.go
```diff
@@ -99,3 +110,37 @@ func (b *BootImage) Unmarshal(buf *uio.Lexer) error {
...
 	return buf.Error()
 }
+
+func getBootImageID(code dhcpv4.OptionCode, o dhcpv4.Options) *BootImageID {
+	v := o.Get(code)
+	if v == nil {
+		return nil
+	}
+	var b BootImageID
+	if err := uio.FromBigEndian(&b, v); err != nil {
```
pmazzini: Should this be `b.FromBytes(v)` instead of `uio.FromBigEndian(&b, v)`?

hugelgupf: Could be, but b.FromBytes(v) is just uio.FromBigEdnain(&b, v) - no difference really

pmazzini:
I like the API simplification and from what I have checked so far the PR lgtm. Additional pair of eyes for this big PR is welcomed :)

pmazzini:
Not approving it yet to give @get9 the chance to review it as well since it includes changes in BSDP and yesterday was a holiday in the US.

insomniacslk:
I didn't have time to review the code, but the proposal looks good to me, and @pmazzini has already reviewed the code. For me this is OK to merge

PR #228, opened 2019-01-13
Conform to RFC 3396 & RFC 2131 wrt Options

hugelgupf:
Removes AddOption and GetOption.

RFC 2131 specifies that options may only appear once (Section 4.1). If
an option does appear more than once, its byte values must be
concatenated.

RFC 3396 further specifies that to send options longer than 255 bytes,
one option may be split into multiple option codes, which must be
concatenated back together by the receiver.

Both of these are concerned with the byte representation of options.
Fact is, based on both RFCs one can say that an option may only appear
once, but may be composed of multiple values. (Like the multiple IP options.)

Because an option may appear only once logically in any case, we remove
the AddOption and GetOption functions and leave only UpdateOption and
GetOneOption.

Also remove all additions & checks of the End option - the marshaling
and unmarshaling code is exclusively responsible for that now.

(will rebase after #226 is merged.)

The full RFC 3396 is not supported at the moment, as the Overload option is not respected. (Who TF even uses that.)

### DIFF
--- a/dhcpv4/bsdp/bsdp.go
+++ b/dhcpv4/bsdp/bsdp.go
@@ -31,7 +31,6 @@ type ReplyConfig struct {
 // ParseBootImageListFromAck parses the list of boot images presented in the
 // ACK[LIST] packet and returns them as a list of BootImages.
 func ParseBootImageListFromAck(ack dhcpv4.DHCPv4) ([]BootImage, error) {
-	var images []BootImage
 	opt := ack.GetOneOption(dhcpv4.OptionVendorSpecificInformation)
 	if opt == nil {
 		return nil, errors.New("ParseBootImageListFromAck: could not find vendor-specific option")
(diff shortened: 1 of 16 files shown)


pmazzini:
This looks good to me, thanks for the cleanup and improvements.

`dhcpv4/ztpv4/ztp_test.go` will need to be updated as well since it is calling `AddOption`.

pmazzini:
Also approving but not merging yet to give @insomniacslk a chance to look at it.

hugelgupf:
rebased :)

hugelgupf:
> Just one question - what about the Pad option? Per concept of padding I would expect that Pad could be added more than once, but this clashes with what described in the RFC.

I think padding should be done by the marshaling routines where necessary. (To pad the packet to its minimum length or to pad the Overload fields.) One more thing for the user not to worry about.

insomniacslk:
> I think padding should be done by the marshaling routines where necessary. (To pad the packet to its minimum length or to pad the Overload fields.) One more thing for the user not to worry about.

Sounds good. This also confirms that after these changes this becomes a production-only library and not allowing anymore scapy-like packet crafting (which is OK, just acknowledging here for the future choices)

pmazzini:
Most things can still be crafted but you are right, we won't control the padding. For me it is OK too.

PR #233, opened 2019-01-19
dhcpv4: build all packets with modifiers

hugelgupf:
Also drop unnecessary return value of Modifier.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -50,7 +50,7 @@ type DHCPv4 struct {
 
 // Modifier defines the signature for functions that can modify DHCPv4
 // structures. This is used to simplify packet manipulation
-type Modifier func(d *DHCPv4) *DHCPv4
+type Modifier func(d *DHCPv4)
 
 // IPv4AddrsForInterface obtains the currently-configured, non-loopback IPv4
 // addresses for iface.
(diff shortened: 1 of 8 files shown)


insomniacslk:
oh, wait, can you first address the linter warnings?

hugelgupf:
> And the build is failing too, `dhcpv4/async/client.go:199:16: mod(message) used as value`

I removed that argument, 'cause I don't think that belongs in the abstraction. SG?

insomniacslk:
> > And the build is failing too, `dhcpv4/async/client.go:199:16: mod(message) used as value`
> 
> I removed that argument, 'cause I don't think that belongs in the abstraction. SG?

Correct, it can go away, it's a leftover of the past

dhcpv4/dhcpv4.go
```diff
@@ -190,74 +185,46 @@ func NewInformForInterface(ifname string, needsBroadcast bool) (*DHCPv4, error)
...
-// NewInform builds a new DHCPv4 Informational message with default Ethernet HW
-// type and specified hardware address. It does NOT put a DHCP End option at the
-// end.
-func NewInform(hwaddr net.HardwareAddr, localIP net.IP) (*DHCPv4, error) {
-	d, err := New()
-	if err != nil {
-		return nil, err
-	}
+// PrependModifiers prepends other to m.
+func PrependModifiers(m []Modifier, other ...Modifier) []Modifier {
```
insomniacslk: is this really necessary? It's basically a synonym of `append`, can't we just do `New(append(modifiers, WithHwAddr(..), ... )` ?

hugelgupf: For one, it's intentionally the other way (prepend, instead of append), so that if you do 

NewReplyForRequest(..., WithSomething())

the WithSomething can override options that NewReplyForRequest enacts.

For example,

NewReplyForRequest(ack, WithHwAddr(somethingelse))

in the prepend world, somethingelse becomes the ClientHWAddr. in the append world, ack.ClientHWAddr becomes the ClientHWAddr.

insomniacslk: sorry, I didn't read it right, makes sense now

hugelgupf:
updated - finally passed all the checks :)

PR #235, opened 2019-01-20
dhcpv6: misc cleanups

hugelgupf:
starting with some of the easy stuff to bring v4 and v6 to parity.


pmazzini:
lgtm, a PR should be created to fix https://github.com/rtr7/router7/ build after merging this.

Do you know if there is anything like RFC 3396 for DHCPv6? Do you think we should also move to using a map in v6?

hugelgupf:
> lgtm, a PR should be created to fix [rtr7/router7](https://github.com/rtr7/router7) build after merging this.

Yeah, will do.

> Do you know if there is anything like RFC 3396 for DHCPv6? Do you think we should also move to using a map in v6?

No, not quite. The same restriction about option appearance exists, but multiple options must not be combined:

> Unless otherwise noted, each option may appear only in the options
   area of a DHCP message and may appear only once.  If an option does
   appear multiple times, each instance is considered separate and the
   data areas of the options MUST NOT be concatenated or otherwise
   combined. (RFC 3315, Section 22)

Now in DHCPv4, there are no "otherwise noted" options that may appear more than once. That's not the case for v6:

> An IA_NA option may only appear in the options area of a DHCP
   message.  A DHCP message may contain multiple IA_NA options. (3315, Section 22.4)

(Same goes for IATAs.)

I personally would still move toward a map, though I'm actually wondering if Getters for each option on each type would be useful for v6. Something like:

```go

type DHCPv6Message struct {
  Options
}

type (m MessageOptions) GetIANAs() []IANA
type (m MessageOptions) GetIATAs() []IATA

type IANA struct {
  Options
}

func (i IANA) GetIAAddresses() []IAAddress
```

This is where maybe I decide to throw the reasoning I used for v4 out the window, because you have to read the RFC carefully to figure out which options may appear where and how many times. That's a PITA. So the disjunct model of option parsing separate from options themselves as I proposed in v4 would kind of suck here. Maybe that's a case for v4 to not use it either.

PR #239, opened 2019-01-25
[async] re-add modifiers to Send method

pmazzini:
It was removed in fe6f307df5d78a54ddd4a56a275043317148fe5a.

It is needed in https://github.com/pinterest/bender/blob/master/dhcpv4/dhcpv4.go#L59.


insomniacslk:
This will also bring the two (sync and async) client interfaces uniform again, looks good to me

PR #241, opened 2019-01-27
Split DHCPv6 and DHCPv4 client into subpackages

### LINKED ISSUE #231: Split client and server from parsing
This will allow finer-grained imports for who only needs a subset of the functionalities (e.g. no need for server or client)

insomniacslk:
Fixes #231


insomniacslk:
this breaks the router7 build. I'll make a PR

insomniacslk:
PR for router7 at https://github.com/rtr7/router7/pull/23

insomniacslk:
This is ready to review. The CI failures are due to router7, and fixed in the related PR

dhcpv4/client4/client.go
```diff
@@ -88,7 +89,7 @@ func makeRawSocket(ifname string) (int, error) {
 	if err != nil {
 		return fd, err
 	}
-	err = BindToInterface(fd, ifname)
+	err = dhcpv4.BindToInterface(fd, ifname)
```
pmazzini: Is the `BindToInterface` going to stay in the `dhcpv4` package?

insomniacslk: yes, for now at least, unless we have a better option. I thought of moving it to `client4` but it's also required for `server4`, and it feels awkward to call `client4.BindToInterface` from the server package

pmazzini: I only see it being called from the client code. Is the plan to start using it for the server as well?

pmazzini: As discussed offline: the `server` code will also use it. This will be shared by `server` and `client` so leaving it in the main directory for now.

dhcpv6/client6/iputils.go
```diff
@@ -1,8 +1,10 @@
-package dhcpv6
+package client6
```
pmazzini: I would leave this in `dhcpv6` since it may also be used by the `server` code.

pmazzini:
Commented about `iputils.go`. The rest looks good to me.

dhcpv6/client6/iputils.go
insomniacslk: good point - there are several things that may go to a separate common package, but I guess `dhcpv6` is still a better option than adding another one

insomniacslk:
@pmazzini done

insomniacslk:
> I wonder if we should move the BSDP client to a separate directory as well. Anyway, approving.

Probably the right thing to do is to move `bsdp` and `async` under `client6`, otherwise leaving them in the main namespace

PR #242, opened 2019-01-28
[dhcpv4] move default to main directory

pmazzini:
This is going to be shared by the server and client code so it should be moved to the main directory. This also makes it consistent with what was done for dhcpv6.


insomniacslk:
dhcpv4/bsdp/client.go:38:55: undefined: client4.ClientPort
dhcpv4/bsdp/client.go:63:65: undefined: client4.ClientPort

pmazzini:
> dhcpv4/bsdp/client.go:38:55: undefined: client4.ClientPort
> dhcpv4/bsdp/client.go:63:65: undefined: client4.ClientPort

Updated.

PR #250, opened 2019-02-16
client4: add new DHCPv4 client.

hugelgupf:
- Able to broadcast UDP before interface is configured.
- RFC2131-compliant retransmission logic.
- Tests.
- Race-condition-averse.

Previous clients (both mine and the ones here) are prone to race
condition errors.

Having one and only one place that calls receive on the socket
"continuously" without having to coordinate hand-offs makes the logic
way easier to follow. We can rely on goroutine scheduling to take care
of some of the blocking semantics.

!! Warning

This PR contains some Apache 2 code taken from gVisor, and the respective file has the header for it.

!! Warning 2

This PR requires Go 1.12.

### DIFF
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -1,0 +1,412 @@
+// Copyright 2018 the u-root Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build go1.12
+
+// Package nclient4 is a small, minimum-functionality client for DHCPv4.
+//
+// It only supports the 4-way DHCPv4 Discover-Offer-Request-Ack handshake as
+// well as the Request-Ack renewal process.
+package nclient4
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"net"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/insomniacslk/dhcp/dhcpv4"
+)
+
+const (
+	defaultTimeout   = 5 * time.Second
+	defaultRetries   = 3
+	defaultBufferCap = 5
+	maxMessageSize   = 1500
+
+	// ClientPort is the port that DHCP clients listen on.
+	ClientPort = 68
+
+	// ServerPort is the port that DHCP servers and relay agents listen on.
+	ServerPort = 67
+)
+
+var (
+	// DefaultServers is the address of all link-local DHCP servers and
+	// relay agents.
+	DefaultServers = &net.UDPAddr{
+		IP:   net.IPv4bcast,
+		Port: ServerPort,
+	}
+)
+
+var (
+	// ErrNoResponse is returned when no response packet is received.
+	ErrNoResponse = errors.New("no matching response packet received")
+)
+
+// pendingCh is a channel associated with a pending TransactionID.
+type pendingCh struct {
+	// SendAndRead closes done to indicate that it wishes for no more
+	// messages for this particular XID.
+	done <-chan struct{}
+
+	// ch is used by the receive loop to distribute DHCP messages.
+	ch chan<- *dhcpv4.DHCPv4
+}
+
+// Client is an IPv4 DHCP client.
+type Client struct {
+	ifaceHWAddr net.HardwareAddr
+	conn        net.PacketConn
+	timeout     time.Duration
+	retry       int
+
+	// bufferCap is the channel capacity for each TransactionID.
+	bufferCap int
+
+	// serverAddr is the UDP address to send all packets to.
+	//
+	// This may be an actual broadcast address, or a unicast address.
+	serverAddr *net.UDPAddr
+
+	// closed is an atomic bool set to 1 when done is closed.
+	closed uint32
+
+	// done is closed to unblock the receive loop.
+	done chan struct{}
+
+	// wg protects any spawned goroutines, namely the receiveLoop.
+	wg sync.WaitGroup
+
+	pendingMu sync.Mutex
+	// pending stores the distribution channels for each pending
+	// TransactionID. receiveLoop uses this map to determine which channel
+	// to send a new DHCP message to.
+	pending map[dhcpv4.TransactionID]*pendingCh
+}
+
+// New returns a client usable with an unconfigured interface.
+func New(ifaceName string, ifaceHWAddr net.HardwareAddr, opts ...ClientOpt) (*Client, error) {
+	c := NewWithConn(nil, ifaceHWAddr, opts...)
+
+	// Do this after so that a caller can still use a WithConn to override
+	// the connection.
+	if c.conn == nil {
+		pc, err := NewRawUDPConn(ifaceName, ClientPort)
+		if err != nil {
+			return nil, err
+		}
+		c.conn = pc
+	}
+	return c, nil
+}
+
+// NewWithConn creates a new DHCP client that sends and receives packets on the
+// given interface.
+func NewWithConn(conn net.PacketConn, ifaceHWAddr net.HardwareAddr, opts ...ClientOpt) *Client {
+	c := &Client{
+		ifaceHWAddr: ifaceHWAddr,
+		timeout:     defaultTimeout,
+		retry:       defaultRetries,
+		serverAddr:  DefaultServers,
+		bufferCap:   defaultBufferCap,
+		conn:        conn,
+
+		done:    make(chan struct{}),
+		pending: make(map[dhcpv4.TransactionID]*pendingCh),
+	}
+
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	c.wg.Add(1)
+	go c.receiveLoop()
+	return c
+}
+
+// Close closes the underlying connection.
+func (c *Client) Close() error {
+	// Make sure not to close done twice.
+	if !atomic.CompareAndSwapUint32(&c.closed, 0, 1) {
+		return nil
+	}
+
+	err := c.conn.Close()
+
+	// Closing c.done sets off a chain reaction:
+	//
+	// Any SendAndRead unblocks trying to receive more messages, which
+	// means rem() gets called.
+	//
+	// rem() should be unblocking receiveLoop if it is blocked.
+	//
+	// receiveLoop should then exit gracefully.
+	close(c.done)
+
+	// Wait for receiveLoop to stop.
+	c.wg.Wait()
+
+	return err
+}
+
+func isErrClosing(err error) bool {
+	// Unfortunately, the epoll-connection-closed error is internal to the
+	// net library.
+	return strings.Contains(err.Error(), "use of closed network connection")
+}
+
+func (c *Client) receiveLoop() {
+	defer c.wg.Done()
+	for {
+		// TODO: Clients can send a "max packet size" option in their
+		// packets, IIRC. Choose a reasonable size and set it.
+		b := make([]byte, maxMessageSize)
+		n, _, err := c.conn.ReadFrom(b)
+		if err != nil {
+			if !isErrClosing(err) {
+				log.Printf("error reading from UDP connection: %v", err)
+			}
+			return
+		}
+
+		msg, err := dhcpv4.FromBytes(b[:n])
+		if err != nil {
+			// Not a valid DHCP packet; keep listening.
+			continue
+		}
+
+		if msg.OpCode != dhcpv4.OpcodeBootReply {
+			// Not a response message.
+			continue
+		}
+
+		// This is a somewhat non-standard check, by the looks
+		// of RFC 2131. It should work as long as the DHCP
+		// server is spec-compliant for the HWAddr field.
+		if c.ifaceHWAddr != nil && !bytes.Equal(c.ifaceHWAddr, msg.ClientHWAddr) {
+			// Not for us.
+			continue
+		}
+
+		c.pendingMu.Lock()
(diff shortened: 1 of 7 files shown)


hugelgupf:
Small update: got rid of the dependency on the netlink package.

hugelgupf:
review this as an in-progress thing? and tell me what to do for darwin and freebsd... mdlayher/raw has support for them, but i'm using a specific linux-only feature (which I actually added to raw) that lets me not have to assemble ethernet frames. but maybe i need to bite the bullet and just do it.

also still need to fix other usages of the old clients.

insomniacslk:
I think I misunderstood when you asked how much we need the promises interface - I didn't think you were going to remove the async client entirely. It is required by the https://github.com/facebookincubator/fbender project to have an asynchronous client with a persistent listener

hugelgupf:
Yeah, still on that - I plan on exposing the send interface directly, I think. The semantics of closing the channel still bother me, trying to find something different.

hugelgupf:
So... I was about to go add what you're asking for, but it's not even really necessary. See 

https://github.com/pinterest/bender/blob/master/dhcpv4/dhcpv4.go#L59

It immediately tries to evaluate the promise. So it doesn't actually need an asynchronous interface, it seems? It would just use a SendAndReadOne (which will automatically use the configured client timeout and retry mechanism).

pmazzini:
> So... I was about to go add what you're asking for, but it's not even really necessary. See
> 
> https://github.com/pinterest/bender/blob/master/dhcpv4/dhcpv4.go#L59
> 
> It immediately tries to evaluate the promise. So it doesn't actually need an asynchronous interface, it seems? It would just use a SendAndReadOne (which will automatically use the configured client timeout and retry mechanism).

cc @earlgreyz

earlgreyz:
> So... I was about to go add what you're asking for, but it's not even really necessary. See
> 
> https://github.com/pinterest/bender/blob/master/dhcpv4/dhcpv4.go#L59
> 
> It immediately tries to evaluate the promise. So it doesn't actually need an asynchronous interface, it seems? It would just use a SendAndReadOne (which will automatically use the configured client timeout and retry mechanism).

I believe the problem with using synchronous client here was that we don't really know the order of answers. The immediate wait for timeout is not for the next message received by the client, but for an answer with a specific `transactionID`.

hugelgupf:
The client in this PR will allow you to do that (I'm going to add a way to get the array of messages sent back and forth).

The timeout in this case will be for a message that has 

- OpcodeBootReply
- matching TransactionID
- matching client hwaddr (this is not in the RFC, but different clients asking for broadcast messages may choose the same XID at the same time... not sure what you think of this, I've not yet tested with major DHCP servers).

hugelgupf:
This should be ready to review now.

netboot.RequestNetbootv4 now returns the offer message, not the whole conversation. Hence, netboot.ConversationToNetconfv4 is now netboot.OfferToNetconfv4.

pmazzini:
CI is failing due to:

```
dhcpv4/server4/server_test.go:74:7: undefined: client4.NewClient
dhcpv4/server4/server_test.go:121:16: c.Exchange undefined (type *client4.Client has no field or method Exchange)
```

The server_test code need to be adjusted.

hugelgupf:
Also removed the integration tag for the server - unnecessary

pmazzini:
I would like to get @insomniacslk's input with regards to the newly added dependencies `github.com/mdlayher/ethernet`, `github.com/mdlayher/raw` and the Apache 2 code. The rest LGTM.

hugelgupf:
This depends on https://github.com/mdlayher/raw/pull/40, pls don't merge before then.

hugelgupf:
The raw PR is merged -- this is good to go, too!

corny:
I have written a DHCPv4 client in Go in 2017. In case you want to take a look at it: https://github.com/digineo/go-dhclient

pmazzini:
The new client is not as friendly for doing unicast requests as the previous one, is it?

hugelgupf:
If you have an already configured network and use that as your PacketConn (WithConn), then SendAndReadOne takes a net.Addr to unicast to. (It does in any case, but with the default PacketConn all packets are broadcast.)

So equal functionality

insomniacslk:
jumping on this a bit late (I'm just back from holidays). I will have time to review tomorrow while on a plane unless this is urgent, in which case let me know

hugelgupf:
Please note that the updated mdlayher/raw *requires Go 1.12*.

insomniacslk:
I'm checking internally whether we can safely drop 1.11.x , sorry for the delay :(

mdlayher:
I'm open to adding compatibility shims for older versions of Go, but I have no plans to do so myself at the moment.

hugelgupf:
@mdlayher while I considered that, this client specifically makes use of the epoll feature, i.e. expects that if net.PacketConn.Close is called in one goroutine, net.PacketConn.Read is released and returns in a different thread.

That's not a given for the previous model of blocking read system calls.

mdlayher:
Ah, right. Yeah, then 1.12 really is the best bet. Just wanted to offer more options.

hugelgupf:
Andrea, you can hold off on reviewing for a bit - I think I'm staring at a deadlock. I'll write a couple more tests.

insomniacslk:
sure - there is no hurry on my side

hugelgupf:
Good to review again, @insomniacslk @pmazzini

hugelgupf:
Couple things changed:

- I am not deleting the old clients for now, but I'm starting to move stuff to the new one at least within this repo.
- I pulled apart the bits which need go 1.12 and the bits that can live without it. The packet's tests can be completed without Go 1.12.
- If you can come up with a net.PacketConn that follows the PacketConn spec for Go 1.11 using raw sockets, it can be used with nclient4.NewWithConn. If you already have a configured interface, you can use this client in Go 1.11 and earlier.

hugelgupf:
(Please do review the code; the test failures are small stuff I'll fix tomorrow. Going to bed for now.)

pmazzini:
> I am not deleting the old clients for now, but I'm starting to move stuff to the new one at least within this repo.

I like the idea of keeping the old client while we transition to the new one. After fixing the tests it LGTM.

hugelgupf:
> > I am not deleting the old clients for now, but I'm starting to move stuff to the new one at least within this repo.
> 
> I like the idea of keeping the old client while we transition to the new one. After fixing the tests it LGTM.

I went this 100% now, and I'm not yet converting netboot or bsdp until y'all are good to switch to Go 1.12.

dhcpv4/nclient4/client.go
```diff
@@ -0,0 +1,396 @@
...
+
+		done:    make(chan struct{}),
+		pending: make(map[dhcpv4.TransactionID]*pendingCh),
+	}
+
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	c.receiveLoop()
```
insomniacslk: nit, feel free to ignore: should the goroutine be created here instead of being in receiveLoop? When I read `receiveLoop` I think of a blocking routine, and I had to look at the implementation to check that it wasn't actually blocking

insomniacslk: and on a second thought - NewClient* makes me think of just *creating* an object, without actually running any non-initialization code. Would it be better to leave the call to receiveLoop to the caller instead?

dhcpv4/nclient4/client.go
```diff
@@ -0,0 +1,396 @@
+// Copyright 2018 the u-root Authors and Andrea Barberio. All rights reserved.
```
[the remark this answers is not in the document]
hugelgupf: haha, you are, I added you explicitly because that's what the top-level license in this repo says - we can just do u-root if you want, idc

dhcpv4/nclient4/client.go
hugelgupf: i've seen this with a lot of New() methods in Go code, I think it's pretty normal to run initialization code. I don't think the caller should have to care about the implementation details of the library (like, relying on a goroutine)

dhcpv4/nclient4/client.go
```diff
@@ -0,0 +1,396 @@
...
+	// to send a new DHCP message to.
+	pending map[dhcpv4.TransactionID]*pendingCh
+}
+
+// NewWithConn creates a new DHCP client that sends and receives packets on the
+// given interface.
+func NewWithConn(conn net.PacketConn, ifaceHWAddr net.HardwareAddr, opts ...ClientOpt) (*Client, error) {
+	c := &Client{
+		ifaceHWAddr: ifaceHWAddr,
+		timeout:     5 * time.Second,
```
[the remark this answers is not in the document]
hugelgupf: done

dhcpv4/nclient4/client.go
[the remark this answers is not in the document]
insomniacslk: yup, let's do just u-root Authors

dhcpv4/nclient4/client.go
```diff
@@ -0,0 +1,415 @@
...
+// Request completes the 4-way Discover-Offer-Request-Ack handshake.
+//
+// Note that modifiers will be applied *both* to Discover and Request packets.
+func (c *Client) Request(ctx context.Context, modifiers ...dhcpv4.Modifier) (offer, ack *dhcpv4.DHCPv4, err error) {
+	offer, err = c.DiscoverOffer(ctx, modifiers...)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO(chrisko): should this be unicast to the server?
```
insomniacslk: maybe tunable? As far as I know there's no recommended default

dhcpv4/nclient4/client.go
```diff
@@ -0,0 +1,415 @@
...
+
+	if _, err := c.conn.WriteTo(msg.ToBytes(), dest); err != nil {
+		cancel()
+		return nil, nil, fmt.Errorf("error writing packet to connection: %v", err)
+	}
+	return ch, cancel, nil
+}
+
+// This error should never be visible to users.
+var errDeadlineExceeded = errors.New("INTERNAL ERROR: deadline exceeded")
```
insomniacslk: why never visible? Sounds like a good way to determine why a transaction failed

dhcpv4/nclient4/client.go
```diff
@@ -0,0 +1,415 @@
...
+		}
+		defer rem()
+
+		for {
+			select {
+			case <-c.done:
+				return ErrNoResponse
+
+			case <-time.After(timeout):
+				return errDeadlineExceeded
```
insomniacslk: nit (i.e. feel free to ignore): if you accept the above on the error not being internal. This could be a `net.Error`, and the caller could check for `err.Timeout()` after a type assertion. It's more verbose, but it reflects an existing workflow in Go

dhcpv4/nclient4/client.go
hugelgupf: Users can write their own Request method, and without DiscoverOffer, you don't know what server the packet came from, so you have nowhere to unicast it to (so it wouldn't make sense to pass it in).

dhcpv4/nclient4/client.go
hugelgupf: This is only used to indicate to retryFn to increase the timeout and retry

insomniacslk: ok, sounds good

dhcpv4/nclient4/client.go
hugelgupf: no need, since this is specifically for the retry logic (adding a comment to that effect in a sec)

dhcpv4/nclient4/client.go
hugelgupf: (in principle, you're right, though, i just... later lol)

dhcpv4/nclient4/conn_linux.go
```diff
@@ -0,0 +1,168 @@
...
+		PacketConn: rawPacketConn,
+		boundAddr:  boundAddr,
+	}
+}
+
+func udpMatch(addr *net.UDPAddr, bound *net.UDPAddr) bool {
+	if bound == nil {
+		return true
+	}
+	if bound.IP != nil && !bound.IP.Equal(addr.IP) {
```
insomniacslk: I assume it's intentional that if addr is nil here this will panic (which shouldn't, being internal), is it right?

dhcpv4/nclient4/ipv4.go
```diff
@@ -0,0 +1,376 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
insomniacslk: this makes me uncomfortable, I'd rather avoid pulling Apache code owned by Google if all we need is just IP/UDP header encoding and decoding.

For similar reasons, weeks ago I have been writing an extremely simplified IPv4/IPv6/UDP/ICMP library, which is still in a non-merged branch of another project. The code is ugly, probably overcomplicated, and definitely work-in-progress, but I have to clean it up and it may be an easier option from a ownership and licensing perspective. Should we use it instead?

The code (shame on me) is currently here, https://github.com/insomniacslk/dublin-traceroute/tree/v6/go/dublintraceroute/net . It is not as simple as the code below but we can improve it.

insomniacslk: as discussed in person, let's try the BSD-licensed net library first

dhcpv4/nclient4/conn_linux.go
hugelgupf: yeah. since addr comes from the packet, and the packet must have an addr or it wouldn't have parsed...

hugelgupf:
ptal

PR #253, opened 2019-02-28
V6 cleanups

hugelgupf:


pmazzini:
Thanks for your contribution!

CI is failing due to:

```
examples/packetcrafting6/main.go:47:33: dhcpv6.WithClientID(duid)(msg) used as value
```

The example needs to be fixed after changing the `Modifiers` func.

There are some lint warnings that need to be fixed as well.

pmazzini:
CI is failing due to:

```
netboot/netboot.go:26:20: cannot use dhcpv6.WithNetboot (type func(dhcpv6.DHCPv6) dhcpv6.DHCPv6) as type dhcpv6.Modifier in append
```

The netboot package needs to be fixed after changing the `Modifiers` func.

pmazzini:
Ci is failing:

```
dhcpv6/server6/server_test.go:61:45: cannot use m (type dhcpv6.DHCPv6) as type *dhcpv6.Message in argument to dhcpv6.NewAdvertiseFromSolicit: need type assertion
```

The server test needs to be updated.

hugelgupf:
Alright, we're finally ready.

dhcpv6/dhcpv6relay.go
```diff
@@ -10,125 +10,94 @@ import (
...
+func (r *RelayMessage) IsRelay() bool {
 	return true
 }
 
-// Recurse into a relay message and extract and return the inner DHCPv6Message.
-// Return nil if none found (e.g. not a relay message).
-func (d *DHCPv6Relay) GetInnerMessage() (DHCPv6, error) {
+// GetInnerMessage recurses into a relay message and extract and return the
+// inner Message.  Return nil if none found (e.g. not a relay message).
+func (r *RelayMessage) GetInnerMessage() (DHCPv6, error) {
```
pmazzini: In order to make it consistent with the rest of the changes, should this return a *Message instead of an interface?

dhcpv6/client6/client.go
```diff
@@ -163,16 +163,16 @@ func (c *Client) sendReceive(ifname string, packet dhcpv6.DHCPv6, expectedType d
 		if err != nil {
 			return nil, err
 		}
-		adv, err = dhcpv6.FromBytes(buf[:n])
+		adv, err = dhcpv6.MessageFromBytes(buf[:n])
```
pmazzini: I think this should be `FromBytes` instead of `MessageFromBytes`.
For example in the case of using a modifier that encapsulates a message inside a relaymessage to simulate a relay. The reply will come as a relaymessage.

hugelgupf: Given how relays encap and decap messages, clients should only ever get raw messages, or your relays are screwed up.

pmazzini: There is a valid use case where you want to simulate a relay, to do a unicast request to a particular server and bypassing the local relay. This can be done with a modifier that encapsulates the message in a relaymessage. That is why we have https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/client6/client.go#L60. The client should support receiving a relaymessage.

pmazzini: And that is why we have https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/client6/client.go#L89

hugelgupf: I see. Included again.

dhcpv6/dhcpv6relay.go
hugelgupf: done

dhcpv6/dhcpv6message.go
```diff
@@ -42,37 +42,37 @@ func GetTime() uint32 {
 	return uint32((now.Nanoseconds() / 1000000000) % 0xffffffff)
 }
 
-// NewSolicitWithCID creates a new SOLICIT message with CID.
-func NewSolicitWithCID(duid Duid, modifiers ...Modifier) (DHCPv6, error) {
-	d, err := NewMessage()
+// NewSolicitWithCID creates a new SOLICIT message with CIm.
```
pmazzini: Is this change a typo?

dhcpv6/dhcpv6message.go
```diff
@@ -14,15 +13,16 @@ import (
...
+type Message struct {
+	MessageType   MessageType
+	TransactionID TransactionID
+	Options       Options
 }
 
 var randomRead = rand.Read
 
-// GenerateTransactionID generates a random 3-byte transaction ID.
+// GenerateTransactionID generates a random 3-byte transaction Im.
```
pmazzini: Is this change a typo?

pmazzini:
Commented about 2 function's comments. The rest LGTM. Thanks for your contribution!
I will approve it after changing those comments.

hugelgupf:
@pmazzini good to go.

will send a PR to rtr7 when this is in

pmazzini:
This PR changed the modifier from

```
type Modifier func(d DHCPv6) DHCPv6
```

to 

```
type Modifier func(d DHCPv6)
```

I think I will have to revert that. We can no longer have a modifier that encapsulates a packet.

PR #254, opened 2019-03-07
Allow Unknown OperState of the link interface

lsiudut:
We hit a bug when Netconf library was failing to bring interface up
despite the fact that it was actually up. It turned out that it's oper
state was not set to UP, what is expected by the library.

According to kernel documentation it is ok proceed if interface state
is Up or Unknown:

```
  Interface is in RFC2863 operational state UP or UNKNOWN. This is for
  backward compatibility, routing daemons, dhcp clients can use this
  flag to determine whether they should use the interface.
```

Also, resaon why operational state may remain Unknown:

```
IF_OPER_UNKNOWN (0):
 Interface is in unknown state, neither driver nor userspace has set
 operational state. Interface must be considered for user data as
 setting operational state has not been implemented in every driver.
```

I modified our code to try DHCP transaction even if `IfUp` failed, but
the OperState was equal to Unknown - it worked perfectly.

### DIFF
--- a/.travis/tests.sh
+++ b/.travis/tests.sh
@@ -28,7 +28,7 @@ done
 
 # Skip go1.9 for this check. rtr7/router7 depends on miekg/dns, which does not
 # support go1.9
-if [ "$TRAVIS_GO_VERSION" = "1.9" ]
+if [[ "$TRAVIS_GO_VERSION" =~ ^1.(9|10|11)$ ]]
 then
     exit 0
 fi
--- a/netboot/netconf.go
+++ b/netboot/netconf.go
@@ -144,8 +144,14 @@ func IfUp(ifname string, timeout time.Duration) (netlink.Link, error) {
 			return nil, fmt.Errorf("cannot get interface %q by name: %v", ifname, err)
 		}
 
-		// if the interface is up, return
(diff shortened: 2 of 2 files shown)


lsiudut:
I don't believe those failures are related to my change?

hugelgupf:
See https://github.com/u-root/u-root/pull/1121

(Another good reason why I'm working on merging your work and our work)

pmazzini:
Although it is not related to this diff, It would be nice to get travis to build. `rt7` depends on `mdlayher/raw` which now requires go 1.12. Can you skip the `rt7` build test when using go 1.10 and 1.11 in https://github.com/insomniacslk/dhcp/blob/master/.travis/tests.sh#L31?

lsiudut:
Yes, I was also considering just checking interface state. The reason I went with comparing operational state is because I kernel docs mentioned it explicitly.

Also, do you think checking just interface state, not OperState, has any advantage here? It seems that checking on OperState is more robust, as it will wait/fail on events like OperDormant, OperLowerLayerDown or OperNotPresent.

pmazzini:
I am going to merge this to unblock @lsiudut. We can modify it afterwards if needed.

PR #256, opened 2019-03-12
client6: new async DHCPv6 client like #250.

hugelgupf:
- Race-condition-averse.
- Supports multiple concurrent requests.
- Tested.
- Requires a fully compatible net.PacketConn.


hugelgupf:
This should be almost the same as #250, but for v6.

dhcpv6/nclient6/conn_linux.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"fmt"
+	"net"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// NewIPv6UDPConn returns a UDP connection bound to both the interface and port
+// given based on a IPv4 DGRAM socket. The UDP connection allows broadcasting.
```
insomniacslk: IPv4?

dhcpv6/nclient6/conn_linux.go
```diff
@@ -0,0 +1,42 @@
...
+func NewIPv6UDPConn(iface string, port int) (net.PacketConn, error) {
+	fd, err := unix.Socket(unix.AF_INET6, unix.SOCK_DGRAM, unix.IPPROTO_UDP)
```
insomniacslk: question (not a request to change): do we need an explicit socket at all with v6? We are guaranteed to always have a link-local address, so net.ListenUDP should be enough.

Partial answer to my own question is that with unix.Socket we can use SO_REUSEADDR and BindToDevice, which is only necessary in case of misconfigurations (e.g. two NICs with the same MAC address, but I'd call it extremely rare) or interface aliases

dhcpv6/nclient6/conn_linux.go
```diff
@@ -0,0 +1,42 @@
...
+	// net.FilePacketConn dups the FD, so we have to close this in any case.
+	defer f.Close()
+
+	// Allow reusing the addr to aid debugging.
+	if err := unix.SetsockoptInt(fd, unix.SOL_SOCKET, unix.SO_REUSEADDR, 1); err != nil {
+		return nil, fmt.Errorf("cannot set reuseaddr on socket: %v", err)
+	}
+	if len(iface) != 0 {
+		// Bind directly to the interface.
+		if err := unix.BindToDevice(fd, iface); err != nil {
```
insomniacslk: what about the portable BindToInterface that we also used in dhcpv4? We need to be compatible at least with darwin :(

pmazzini:
The PR contains the binary file `examples/server6/server6`. Should it be removed?

dhcpv6/nclient6/client.go
```diff
@@ -0,0 +1,364 @@
...
+			if err != nil {
+				if !isErrClosing(err) {
+					log.Printf("error reading from UDP connection: %v", err)
+				}
+				return
+			}
+
+			msg, err := dhcpv6.MessageFromBytes(b[:n])
+			if err != nil {
+				// Not a valid DHCP packet; keep listening.
```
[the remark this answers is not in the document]
hugelgupf: invalid options should never fail decoding of the package itself anyway -- once options move to the DHCPv4 model of decoding on demand that'll be true here.

dhcpv6/nclient6/conn_linux.go
hugelgupf: good point -- BindToDevice is taken care off by the Zone config, and we could add reuseaddr to the UDPConn we get, too, if needed.

hugelgupf:
> The PR contains the binary file `examples/server6/server6`. Should it be removed?

Actually, I *am* removing it, because it's already in the repo: https://github.com/insomniacslk/dhcp/blob/master/examples/server6/server6

insomniacslk:
the CI failure is on rtr7, which breaks with this PR. I'm sending another PR to @stapelberg to fix that

```
# github.com/rtr7/router7/internal/dhcp6
../../rtr7/router7/internal/dhcp6/dhcp6.go:220:18: undefined: dhcpv6.NewSolicitForInterface
```

PR #258, opened 2019-03-12
revert modifiers definition

pmazzini:
Revert the change to the modifiers so that we can have a modifier that encapsulates packets.

The PR in question is https://github.com/insomniacslk/dhcp/pull/253.

### DIFF
--- a/dhcpv6/modifiers.go
+++ b/dhcpv6/modifiers.go
@@ -10,26 +10,28 @@ import (
 
 // WithClientID adds a client ID option to a DHCPv6 packet
 func WithClientID(duid Duid) Modifier {
-	return func(d DHCPv6) {
+	return func(d DHCPv6) DHCPv6 {
 		cid := OptClientId{Cid: duid}
 		d.UpdateOption(&cid)
+		return d
 	}
 }
 
 // WithServerID adds a client ID option to a DHCPv6 packet
 func WithServerID(duid Duid) Modifier {
-	return func(d DHCPv6) {
+	return func(d DHCPv6) DHCPv6 {
 		sid := OptServerId{Sid: duid}
 		d.UpdateOption(&sid)
+		return d
 	}
 }
 
 // WithNetboot adds bootfile URL and bootfile param options to a DHCPv6 packet.
-func WithNetboot(d DHCPv6) {
+func WithNetboot(d DHCPv6) DHCPv6 {
 	msg, ok := d.(*Message)
 	if !ok {
 		log.Printf("WithNetboot: not a Message")
-		return
+		return d
 	}
 	// add OptionBootfileURL and OptionBootfileParam
 	opt := msg.GetOneOption(OptionORO)
@@ -41,30 +43,32 @@ func WithNetboot(d DHCPv6) {
 	oro.AddRequestedOption(OptionBootfileURL)
 	oro.AddRequestedOption(OptionBootfileParam)
 	msg.UpdateOption(oro)
-	return
+	return d
 }
 
 // WithUserClass adds a user class option to the packet
 func WithUserClass(uc []byte) Modifier {
 	// TODO let the user specify multiple user classes
-	return func(d DHCPv6) {
+	return func(d DHCPv6) DHCPv6 {
 		ouc := OptUserClass{UserClasses: [][]byte{uc}}
 		d.AddOption(&ouc)
+		return d
 	}
 }
 
 // WithArchType adds an arch type option to the packet
 func WithArchType(at iana.Arch) Modifier {
-	return func(d DHCPv6) {
+	return func(d DHCPv6) DHCPv6 {
 		ao := OptClientArchType{ArchTypes: []iana.Arch{at}}
 		d.AddOption(&ao)
+		return d
 	}
 }
 
 // WithIANA adds or updates an OptIANA option with the provided IAAddress
 // options
 func WithIANA(addrs ...OptIAAddress) Modifier {
-	return func(d DHCPv6) {
+	return func(d DHCPv6) DHCPv6 {
 		opt := d.GetOneOption(OptionIANA)
 		if opt == nil {
 			opt = &OptIANA{}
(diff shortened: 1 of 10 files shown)


dhcpv6/modifiers.go
```diff
@@ -10,26 +10,28 @@ import (
 
 // WithClientID adds a client ID option to a DHCPv6 packet
 func WithClientID(duid Duid) Modifier {
-	return func(d DHCPv6) {
+	return func(d DHCPv6) DHCPv6 {
```
hugelgupf: :/

This means we can't use specific types with modifiers, or you have to ignore the return value.

I feel like encapsulation should not be done by a modifier. Modifiers as they are here are used to add stuff to the inside of the packet, not to redefine the packet itself. I'd just call that an Encapsulator and make it a separate thing?

dhcpv6/modifiers_test.go
```diff
@@ -46,32 +46,30 @@ func TestWithRequestedOptions(t *testing.T) {
...
-	WithIANA(OptIAAddress{
+	d := WithIANA(OptIAAddress{
 		IPv6Addr:          net.ParseIP("::1"),
 		PreferredLifetime: 3600,
 		ValidLifetime:     5200,
-	})(&d)
-	require.Equal(t, 1, len(d.Options))
-	require.Equal(t, OptionIANA, d.Options[0].Code())
+	})(&Message{})
+	require.Equal(t, 1, len(d.(*Message).Options))
```
hugelgupf: Now you have to do all these casts, and that's crummy. I'd prefer to separate Modifiers and Encaps rather than reintroduce the need for casting.

hugelgupf: In my mind, casting like this will only lead to production panics and reboots.

dhcpv6/modifiers.go
insomniacslk: originally the modifiers were intended to be able to also modify the packet if necessary, but I agree that this is not the right way to use them

pmazzini: > I'd just call that an Encapsulator and make it a separate thing?

What about the client code? We send Relay messages using modifiers.

dhcpv6/modifiers_test.go
pmazzini: > Now you have to do all these casts

I am reverting as few things as possible to get some internal code to work again.
These casts were previously there.

> In my mind, casting like this will only lead to production panics and reboots.

I agree but these are only tests.

dhcpv6/modifiers.go
pmazzini: I am not against it but until we come up with a solution, I would revert it.
This is what we are currenty using to do unicast tests to specific servers.

dhcpv6/modifiers_test.go
hugelgupf: > > Now you have to do all these casts
> 
> I am reverting as few things as possible to get some internal code to work again.
> These casts were previously there.
> 
> > In my mind, casting like this will only lead to production panics and reboots.
> 
> I agree but these are only tests.

There is no such thing as "only tests". These tests use exported features of the library. That means any user that wants to use the library has to cast like this to make it work. Yes, it was previously there, but that's the reason for the improvement.

dhcpv6/modifiers.go
hugelgupf: > I am not against it but until we come up with a solution, I would revert it.
> This is what we are currenty using to do unicast tests to specific servers.

Can you wait to import the new code until you come up with a solution?

I'd love to come up with one for you, but without seeing the code, it's hard to say what would be a good way to deal with it.

pmazzini: > Can you wait to import the new code until you come up with a solution?

Sure!

> I'd love to come up with one for you, but without seeing the code, it's hard to say what would be a good way to deal with it.

We have a modifier similar to:

```
func encapsulateRelay(peerAddr net.IP) dhcpv6.Modifier {
	return func(packet dhcpv6.DHCPv6) dhcpv6.DHCPv6 {
		encapsulatedPacket, err := dhcpv6.EncapsulateRelay(packet, dhcpv6.MessageTypeRelayForward, net.IPv6zero, peerAddr)
		return encapsulatedPacket
	}
}
```

Which is passed to the client's `Exchange` method.

So the client creates a Solicit, applies the modifier and sends a RelayForw.
Then it receives a RelayRepl and decapsulates it:
https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/client6/client.go#L61

With the advertise, the client creates a request, applies the modifier and sends again a RelayForw.

So basically with 1 modifier we were emulating a relay, which broke after changing how modifiers behave.

What we need is a client supporting exchange of relay messages, which we were doing with a modifier.

hugelgupf: > func encapsulateRelay(peerAddr net.IP) dhcpv6.Modifier {
> 	return func(packet dhcpv6.DHCPv6) dhcpv6.DHCPv6 {
> 		encapsulatedPacket, err := dhcpv6.EncapsulateRelay(packet, dhcpv6.MessageTypeRelayForward, net.IPv6zero, peerAddr)
> 		return encapsulatedPacket
> 	}
> }

I feel like this is kind of a dangerous modifier, because it's the only modifier I've seen where it's ordering with /other/ modifiers matters. E.g.

```go
client.Exchange(ifname, encapsulateRelay(nil), WithClientID(foobar))
```
is a different packet than
```go
client.Exchange(ifname, msg, WithClientID(foobar), encapsulateRelay(nil))
```

And I feel this is not intuitive. I think It wouldn't be that hard to do something like
```go
client.Exchange(ifname, Encap(msg, WithClientID(foo)))
```
instead, and it's *very clear* which message the modifier applies to in this model.

This would also mean that you can drop the modifiers on client.Exchange. It might be useful to have a helper like this instead:

```go
func Apply(msg DHCPv6, modifiers ...Modifier) DHCPv6

func Encap(msg DHCPv6, modifiers ...Modifier) *RelayMessage

func EncapWithFoo(...) *RelayMessage
```

and then you can
```go
client.Exchange(ifname, Apply(msg, WithClientID(foo)))
client.Exchange(ifname, Apply(Encap(msg, WithClientID(bar)), WithClientID(foo)))
client.Exchange(ifname, Encap(Apply(msg, WithClientID(bar)))
```

I'm also confused about clients that act like relays. This is just for testing, right? Because I imagine a relay is actually made using the `server6` package rather than `client6`.

pmazzini: > I feel like this is kind of a dangerous modifier, because it's the only modifier I've seen where it's ordering with /other/ modifiers matters. E.g.

Well, the order of the rest of the modifiers also matters, since we have a list of options instead of a map, it affects the ordering of the options.

> and then you can

```
client.Exchange(ifname, Apply(msg, WithClientID(foo)))
```

Ok, this means changing the client API so that `Exchange` can now take `DHCPv6` packets.

```
client.Exchange(ifname, Encap(Apply(msg, WithClientID(bar)))
```

How is the Request generated in this case by the client? I mean, how is it encapsulated again?

> This is just for testing, right?

No, for doing checks to servers (monitoring).

hugelgupf: > > I feel like this is kind of a dangerous modifier, because it's the only modifier I've seen where it's ordering with /other/ modifiers matters. E.g.
> 
> Well, the order of the rest of the modifiers also matters, since we have a list of options instead of a map, it affects the ordering of the options.

I feel the same about that. The order should not matter. The marshaling order of options *should* not affect DHCP clients & servers either, and it doesn't per spec. Modifiers should be applicable in any order.

For encapsulation, it's not the marshaling order that's affected, but *which* packet the modifier gets used on. Is it an option for the encapsulated message, or the relay message? Depends on whether it's before or after the relay modifier. That's too subtle for me from an API design perspective.

> 
> > and then you can
> 
> ```
> client.Exchange(ifname, Apply(msg, WithClientID(foo)))
> ```
> Ok, this means changing the client API so that `Exchange` can now take `DHCPv6` packets.
> 
> ```
> client.Exchange(ifname, Encap(Apply(msg, WithClientID(bar)))
> ```
> How is the Request generated in this case by the client? I mean, how is it encapsulated again?

Ah, sorry, you're right. I think for something like this, a specialized client like yours should build its Relay message the right way, and then just use SendAndRead to send the packet. I.e. I think to send a different kind of packet, you should not use Exchange.

> 
> > This is just for testing, right?
> 
> No, for doing checks to servers (monitoring).

pmazzini: How about adding the `PeerAddr`attribute to the client? Whenever that is set, the packet gets entcapsulated before sending it, that should solve our use case.

pmazzini: It should be a simple change.

pmazzini:
Sorry for the noise. I am going to put up different PR to fix this.

PR #259, opened 2019-03-13
client: simulate relay

pmazzini:
Follow-up of https://github.com/insomniacslk/dhcp/pull/258. Add support for the client to behaving as a relay.


pmazzini:
Thoughts?

pmazzini:
Since this is disabled by default, I'll go ahead and merge it to get back the functionality we lost with https://github.com/insomniacslk/dhcp/pull/253. Later we can check better alternatives.

PR #261, opened 2019-03-29
dhcpv4: pad v4 messages to a 300 byte minimum

hugelgupf:
Certain older DHCP servers and relay agents follow the RFC 951 BOOTP
standard in which BOOTP/DHCP messages have a 300 byte minimum length.

cc @ganshun

I'd appreciate the fast path on this review :)

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -16,6 +16,7 @@
 package dhcpv4
 
 import (
+	"bytes"
 	"crypto/rand"
 	"errors"
 	"fmt"
@@ -39,6 +40,9 @@ const (
 
 	// MaxMessageSize is the maximum size in bytes that a DHCPv4 packet can hold.
 	MaxMessageSize = 576
+
+	// Per RFC 951, the minimum length of a packet is 300 bytes.
+	bootpMinLen = 300
 )
 
 // magicCookie is the magic 4-byte value at the beginning of the list of options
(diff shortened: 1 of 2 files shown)


dhcpv4/dhcpv4.go
```diff
@@ -468,6 +469,19 @@ func (d *DHCPv4) ToBytes() []byte {
...
 
+	// DHCP is based on BOOTP, and BOOTP messages have a minimum length of
+	// 300 bytes per RFC 951. This not stated explicitly, but if you sum up
+	// all the bytes in the message layout, you'll get 300 bytes.
+	//
+	// Some DHCP servers and relay agents care about this BOOTP legacy B.S.
+	// and "conveniently" drop messages that are less than 300 bytes long.
+	//
+	// We subtract one byte for the OptionEnd option.
+	const bootpMinLen = 300 - 1
```
pmazzini: Should this const be added at the top of the file together with `minPacketLen`?

insomniacslk: could also be `len(OptionEnd)` (but really feel free to ignore this comment)

dhcpv4/dhcpv4.go
```diff
@@ -468,6 +469,19 @@ func (d *DHCPv4) ToBytes() []byte {
...
+	if buf.Len() < bootpMinLen {
```
insomniacslk: nit: what about subtracting the size of OptionEnd here and below, instead? I'd rather leave `bootpMinLen` semantically correct (i.e. 300) even if it's unexported and clearly commented

dhcpv4/dhcpv4.go
hugelgupf: done

hugelgupf: it cannot, it'd have to be sizeof, and i'd rather not pull in reflect or any of that stuff

dhcpv4/dhcpv4.go
hugelgupf: done

PR #263, opened 2019-03-29
server4 test: pick ports > 32k

hugelgupf:
This is dumb, but when I import this stuff to Google, I can only use ports 32k-64k. I'd like to avoid maintaining an internal patchset on top of dhcp... would y'all be fine taking this?


insomniacslk:
I got to this too late, I wish we could make this a tunable rather than a hardcoded value

hugelgupf:
We can make some flags

PR #269, opened 2019-04-06
nclient6: small fixes

hugelgupf:
- RapidCommit solicits wait for Reply messages, not Advertise.
- Default recipient should be all relay agents and servers, not just
  servers.
- Make New() and NewWithConn() interface same as in nclient4.

### DIFF
--- a/dhcpv6/nclient6/client.go
+++ b/dhcpv6/nclient6/client.go
@@ -90,15 +90,30 @@ func NewIPv6UDPConn(iface string, port int) (net.PacketConn, error) {
 	})
 }
 
-// New creates a new DHCP client that sends and receives packets on the given
-// interface.
-func New(ifaceHWAddr net.HardwareAddr, opts ...ClientOpt) (*Client, error) {
+// New returns a new DHCPv6 client for the given network interface.
+func New(iface string, opts ...ClientOpt) (*Client, error) {
+	c, err := NewIPv6UDPConn(iface, dhcpv6.DefaultClientPort)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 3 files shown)


dhcpv6/nclient6/client.go
```diff
@@ -90,15 +90,26 @@ func NewIPv6UDPConn(iface string, port int) (net.PacketConn, error) {
 	})
 }
 
-// New creates a new DHCP client that sends and receives packets on the given
-// interface.
-func New(ifaceHWAddr net.HardwareAddr, opts ...ClientOpt) (*Client, error) {
+// New returns a new DHCPv6 client.
+func New(ifaceName string, ifaceHWAddr net.HardwareAddr, opts ...ClientOpt) (*Client, error) {
```
insomniacslk: A question that slipped through the cracks in the v4 PR - do we need to specify interface name and hardware address explicitly? We could get the hardware address from the interface name. Am I missing something?

insomniacslk:
also needs to fix server tests:
```
dhcpv6/server6/server_test.go:46:41: cannot use net.HardwareAddr literal (type net.HardwareAddr) as type string in argument to nclient6.New
dhcpv6/server6/server_test.go:47:20: cannot use nclient6.WithConn(fakeUnconnectedConn literal) (type nclient6.ClientOpt) as type net.HardwareAddr in argument to nclient6.New
```

dhcpv6/nclient6/client.go
hugelgupf: not missing anything, fixed. will post a v4 PR.

PR #271, opened 2019-04-09
Replace vishvananda/netlink with jsimonetti/rtnetlink

### LINKED ISSUE #257: Swap out netlink library
The most used netlink library out there kind of plays fast and loose with unsafe pointers, sometimes leading to memory corruption.

(E.g. https://github.com/vishvananda/netlink/blob/master/ioctl_linux.go#L89 - should be using binary marshaling instead, since `e` may be freed anytime after - luckily, it gets returned and used *after* Ifreq is given to a syscall, but this kind of stuff is all over that codebase.)

It's also a huge grab-bag of netlink code, which means it's quite a large dependency in our embedded environment. I'd like to recommend switching to a netlink library based on https://github.com/mdlayher/netlink, such as https://godoc.org/github.com/jsimonetti/rtnetlink, which is more targeted (and has no memory corruption).

I'm happy to do this switch when I get to these items.

insomniacslk:
Fixes #257

Note: this PR also introduces a transitional `RTNL` object, that will be later removed as part of the plan of implementing it into `jsimonetti/rtnetlink`.

And saves ~100kb (and probably more once migrated everything else).
Tested both v6 and v4 netboot, using qemu + kea. The client was able in both cases to configure the network as returned by kea, and to download and kexec the payload via bootfile URL / bootfile name.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: go
 sudo: required
 
 go:
-  - "1.9"
   - "1.10"
   - "1.11"
+  - "1.12"
   - tip
 
 before_install:
--- a/.travis/tests.sh
+++ b/.travis/tests.sh
@@ -13,13 +13,13 @@ for d in $(go list ./... | grep -v vendor); do
         rm profile.out
     fi
     # integration tests
-    go test -c -tags=integration -race -coverprofile=profile.out -covermode=atomic $d
+    go test -c -cover -tags=integration -race -covermode=atomic $d
     testbin="./$(basename $d).test"
     # only run it if it was built - i.e. if there are integ tests
-    test -x "${testbin}" && sudo "./${testbin}"
+    test -x "${testbin}" && sudo "./${testbin}" -test.coverprofile=profile.out
     if [ -f profile.out ]; then
         cat profile.out >> coverage.txt
-        rm profile.out
+        rm -f profile.out
     fi
 done
 
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -94,13 +94,17 @@ type Client struct {
 }
 
 // New returns a client usable with an unconfigured interface.
-func New(ifaceName string, ifaceHWAddr net.HardwareAddr, opts ...ClientOpt) (*Client, error) {
-	c := NewWithConn(nil, ifaceHWAddr, opts...)
+func New(iface string, opts ...ClientOpt) (*Client, error) {
+	i, err := net.InterfaceByName(iface)
+	if err != nil {
+		return nil, err
+	}
+	c := NewWithConn(nil, i.HardwareAddr, opts...)
 
 	// Do this after so that a caller can still use a WithConn to override
 	// the connection.
 	if c.conn == nil {
-		pc, err := NewRawUDPConn(ifaceName, ClientPort)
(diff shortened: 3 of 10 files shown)


insomniacslk:
Space-wise:

with this change:
```
$ goweight -j | jq '[.[].size] | add' 
17496098
```

before this change:
```
$ goweight -j | jq '[.[].size] | add' 
20094696
```

which is 2598598 bytes of difference according to `goweight`. After compression it's ~160kb smaller

insomniacslk:
needed another fix, https://github.com/jsimonetti/rtnetlink/pull/27 (now merged)

hugelgupf:
Wow. That suggests to me there are problems with the dead code elimination in Go.

insomniacslk:
@hugelgupf I think we will reclaim more space when u-root's `ip` is migrated to the rtnetlink library. I plan to do that, meanwhile @jsimonetti is helping by implementing some of the required features (e.g. for `neigh`, just merged today)

insomniacslk:
meanwhile @jsimonetti is helping on the same thing on u-root[1] :) we may want to unify the code base. I will probably refactor some stuff under some u-root pkg and reuse here and there

[1] https://github.com/u-root/u-root/pull/1187

insomniacslk:
I added a `rtnetnlink.go` file that exposes a higher level interface with just the methods we need here. However this is transitional - I plan to implement a proper higher level interface either directly in `rtnetlink` (if @jsimonetti agrees), or in u-root.

insomniacslk:
the coverage decrease is due to this code being less compact. To test it properly I could write integ tests similarly to what @mdlayher does for `netlink`

mdlayher:
> I plan to implement a proper higher level interface either directly in rtnetlink (if @jsimonetti agrees), or in u-root.

I definitely think it'd make sense to have an optional wrapper package like `github.com/jsimonetti/rtnetlink/rtnlquick` (name TBD) for these types of operations.

> To test it properly I could write integ tests similarly to what @mdlayher does for netlink

Yeah this can be a bit of a hassle, but it's probably worth doing.  I also have packages `nltest` and `genltest` for testing netlink interactions without actually touching the kernel; you can see it in use here: https://github.com/mdlayher/wireguardctrl/blob/master/internal/wgnl/client_linux_test.go.

It may be worth making an `rtnltest` as well in the same style, so users of `rtnetlink` can do some quick and easy verification of their code.

https://godoc.org/github.com/mdlayher/netlink/nltest
https://godoc.org/github.com/mdlayher/genetlink/genltest

insomniacslk:
> I definitely think it'd make sense to have an optional wrapper package 

I completely agree. After a chat with @jsimonetti we agreed we'll go for an optional `rtnl` package under the `jsimonetti/rtnetlink` repo

Re. the integ testing I also agree, they are absolutely necessary :) thanks for pointing out some good reference too!

insomniacslk:
For what concerns this PR instead I'd rather get it merged so I can get unblocked on shrinking the size of LinuxBoot images, to then transition to the upcoming `rtnl` library and get rid of this temporary wrapper

insomniacslk:
just tested also the DHCPv6 code path (with the usual qemu + kea combo), and the host is able to configure the network as returned by kea, and download and kexec the payload.

PR #273, opened 2019-04-16
[server] Serve returns an error, and closes when done

insomniacslk:
Server6 and Server4 had Serve methods that return without reporting an
error, changed in this PR.
Serve now also closes the connection when done.

~~The Server6.Close and Server4.Close methods now also check for nil conn.~~

~~Doing this safely requires an RWLock around `conn`, and doesn't guarantee that the handler will properly lock around using `conn`. I am open to suggestions.~~ EDIT: will reconsider for another PR

### DIFF
--- a/dhcpv4/server4/server.go
+++ b/dhcpv4/server4/server.go
@@ -3,24 +3,27 @@ package server4
 import (
 	"log"
 	"net"
-	"sync"
 
 	"github.com/insomniacslk/dhcp/dhcpv4"
 )
 
 /*
   To use the DHCPv4 server code you have to call NewServer with two arguments:
+  - an address to listen on, and
   - a handler function, that will be called every time a valid DHCPv4 packet is
-    received, and
-  - an address to listen on.
+    received.
+
+  The address to listen on is used to know IP address, port and optionally the
+  scope to create and UDP socket to listen on for DHCPv4 traffic.
 
   The handler is a function that takes as input a packet connection, that can be
   used to reply to the client; a peer address, that identifies the client sending
   the request, and the DHCPv4 packet itself. Just implement your custom logic in
   the handler.
 
-  The address to listen on is used to know IP address, port and optionally the
-  scope to create and UDP socket to listen on for DHCPv4 traffic.
+  Optionally, NewServer can receive options that will modify the server object.
+  Some options already exist, for example WithConn. If this option is passed with
+  a valid connection, the listening address argument is ignored.
 
   Example program:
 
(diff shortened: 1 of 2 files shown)


dhcpv4/server4/server.go
```diff
@@ -117,10 +128,13 @@ func NewServer(addr *net.UDPAddr, handler Handler, opt ...ServerOpt) (*Server, e
 	}
 	if s.conn == nil {
 		var err error
-		s.conn, err = net.ListenUDP("udp4", addr)
+		conn, err := net.ListenUDP("udp4", addr)
 		if err != nil {
 			return nil, err
 		}
+		s.mu.Lock()
```
hugelgupf: Why is there a connection mutex at all? You aren't able to change it at runtime

insomniacslk: I added it because there was a race condition when I added the check in `Close` for a nil conn (in case someone calls `Close` twice), but I will remove it

insomniacslk: sounds like the same thing that you fixed in https://github.com/insomniacslk/dhcp/pull/274 ? Except that there was no reason here until I changed how `Close` works :D (which I reverted for now)

dhcpv4/server4/server.go
```diff
@@ -62,22 +66,22 @@ type Handler func(conn net.PacketConn, peer net.Addr, m *dhcpv4.DHCPv4)
 
 // Server represents a DHCPv4 server object
 type Server struct {
-	conn       net.PacketConn
-	connMutex  sync.Mutex
-	shouldStop chan bool
```
insomniacslk: I removed this because dead code

dhcpv4/server4/server.go
```diff
@@ -3,24 +3,27 @@ package server4
...
+  scope to create and UDP socket to listen on for DHCPv4 traffic.
 
   The handler is a function that takes as input a packet connection, that can be
   used to reply to the client; a peer address, that identifies the client sending
   the request, and the DHCPv4 packet itself. Just implement your custom logic in
   the handler.
 
-  The address to listen on is used to know IP address, port and optionally the
-  scope to create and UDP socket to listen on for DHCPv4 traffic.
+  Optionally, NewServer can receive options that will modify the server object.
```
insomniacslk: this block just fixes documentation. It's an unnecessary burden to update docs at every API change, so in the future I'll remove example comments and switch to an examples directory

insomniacslk:
@hugelgupf I'll merge this after removing the mutex thing, but please comment here if you spot some other issue

PR #274, opened 2019-04-17
nclient4: fix conn race condition

hugelgupf:
Why did I ever think this would work??

Signed-off-by: Christopher Koch <chrisko@google.com>

### DIFF
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -99,23 +99,16 @@ func New(iface string, opts ...ClientOpt) (*Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	c := NewWithConn(nil, i.HardwareAddr, opts...)
-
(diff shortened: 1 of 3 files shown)


hugelgupf:
we should set up a 

go test -race

invocation on travis

insomniacslk:
> we should set up a go test -race invocation on travis

But it's there: https://github.com/insomniacslk/dhcp/edit/master/.travis/tests.sh

dhcpv4/nclient4/client.go
```diff
@@ -132,9 +125,12 @@ func NewWithConn(conn net.PacketConn, ifaceHWAddr net.HardwareAddr, opts ...Clie
 		opt(c)
 	}
 
+	if c.conn == nil {
```
insomniacslk: nit: could check early if `conn != nil` at the beginning of the function instead?

PR #276, opened 2019-04-23
[dhcpv6] Solicit messages derive default IAID from MAC address

insomniacslk:
IAID must be set by the client. This patch generates the IAID from the
MAC address of the interface. To do so, a new WithIAID modifier is
added, the interface of NewSolicitWithCID now requires a hwaddr
parameter, and NewAdvertiseFromSolicit copies the IA_NA option from the
solicit if present.

### DIFF
--- a/dhcpv6/async/client_test.go
+++ b/dhcpv6/async/client_test.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/insomniacslk/dhcp/dhcpv6"
 	"github.com/insomniacslk/dhcp/dhcpv6/client6"
-	"github.com/insomniacslk/dhcp/iana"
 	"github.com/stretchr/testify/require"
 )
 
@@ -20,13 +19,7 @@ func solicit(input string) (*dhcpv6.Message, error) {
 	if err != nil {
 		return nil, err
 	}
-	duid := dhcpv6.Duid{
-		Type:          dhcpv6.DUID_LLT,
-		HwType:        iana.HWTypeEthernet,
-		Time:          dhcpv6.GetTime(),
-		LinkLayerAddr: mac,
-	}
-	return dhcpv6.NewSolicitWithCID(duid)
+	return dhcpv6.NewSolicit(mac)
 }
 
 // server creates a server which responds with a predefined response
--- a/dhcpv6/dhcpv6_test.go
+++ b/dhcpv6/dhcpv6_test.go
@@ -197,7 +197,7 @@ func TestNewReplyFromMessage(t *testing.T) {
 	require.Error(t, err)
 }
 
-func TestNewMessageTypeSolicitWithCID(t *testing.T) {
+func TestNewMessageTypeSolicit(t *testing.T) {
 	hwAddr, err := net.ParseMAC("24:0A:9E:9F:EB:2B")
 	require.NoError(t, err)
 
@@ -207,7 +207,7 @@ func TestNewMessageTypeSolicitWithCID(t *testing.T) {
 		LinkLayerAddr: hwAddr,
 	}
 
-	s, err := NewSolicitWithCID(duid)
+	s, err := NewSolicit(hwAddr, WithClientID(duid))
 	require.NoError(t, err)
 
 	require.Equal(t, s.Type(), MessageTypeSolicit)
@@ -227,6 +227,14 @@ func TestNewMessageTypeSolicitWithCID(t *testing.T) {
 	require.Contains(t, opts, OptionDNSRecursiveNameServer)
 	require.Contains(t, opts, OptionDomainSearchList)
 	require.Equal(t, len(opts), 2)
+
+	// Check IA_NA
+	iaid := [4]byte{hwAddr[2], hwAddr[3], hwAddr[4], hwAddr[5]}
+	iaNaOption := s.GetOneOption(OptionIANA)
+	require.NotNil(t, iaNaOption)
+	iaNa, ok := iaNaOption.(*OptIANA)
+	require.True(t, ok)
+	require.Equal(t, iaid, iaNa.IaId)
 }
 
 func TestIsUsingUEFIArchTypeTrue(t *testing.T) {
--- a/dhcpv6/dhcpv6message.go
+++ b/dhcpv6/dhcpv6message.go
@@ -42,8 +42,15 @@ func GetTime() uint32 {
 	return uint32((now.Nanoseconds() / 1000000000) % 0xffffffff)
 }
 
-// NewSolicitWithCID creates a new SOLICIT message with CID.
-func NewSolicitWithCID(duid Duid, modifiers ...Modifier) (*Message, error) {
+// NewSolicit creates a new SOLICIT message, using the given hardware address to
+// derive the IAID in the IA_NA option.
+func NewSolicit(hwaddr net.HardwareAddr, modifiers ...Modifier) (*Message, error) {
+	duid := Duid{
+		Type:          DUID_LLT,
+		HwType:        iana.HWTypeEthernet,
+		Time:          GetTime(),
+		LinkLayerAddr: hwaddr,
+	}
 	m, err := NewMessage()
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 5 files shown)


dhcpv6/async/client_test.go
```diff
@@ -26,7 +26,7 @@ func solicit(input string) (*dhcpv6.Message, error) {
 		Time:          dhcpv6.GetTime(),
 		LinkLayerAddr: mac,
 	}
-	return dhcpv6.NewSolicitWithCID(duid)
+	return dhcpv6.NewSolicitWithCID(duid, mac, dhcpv6.WithIAID([4]byte{1, 2, 3, 4}))
```
pmazzini: How about using the duid to generate the IAID instead of having to pass the mac?

dhcpv6/dhcpv6message.go
```diff
@@ -101,6 +102,12 @@ func NewAdvertiseFromSolicit(sol *Message, modifiers ...Modifier) (*Message, err
 		return nil, errors.New("Client ID cannot be nil in SOLICIT when building ADVERTISE")
 	}
 	adv.AddOption(cid)
+	// add IA_NA
```
pmazzini: I would leave this to the server implementation. Not having a IA_NA is not a reason to hard fail, client may be requesting IA_PD, etc.

dhcpv6/modifiers.go
```diff
@@ -77,6 +77,21 @@ func WithIANA(addrs ...OptIAAddress) Modifier {
 	}
 }
 
+// WithIAID updates an OptIANA option with the provided IAID
+func WithIAID(iaid [4]byte) Modifier {
```
pmazzini: I find this a little bit confusing, both `WithIANA` and `WithIAID` may create a IA_NA option.

dhcpv6/async/client_test.go
insomniacslk: I am using the same approach as dhclient, I'd rather stick with it. This can be changed with a modifier if necessary

dhcpv6/dhcpv6message.go
insomniacslk: good point. I will remove this

dhcpv6/modifiers.go
insomniacslk: I know, and I did it because WithIANA is confusing to me in the first place - it doesn't change IANA, but it's a shortcut to extend the IAAddress options within the IANA option. I think both use cases are good, and the documentation should be enough to avoid misunderstandings

dhcpv6/dhcpv6message.go
insomniacslk: done

dhcpv6/modifiers.go
insomniacslk: the alternative in my opinion is to rename WithIANA to WithIAAddrs. WithIAID itself does what it says - it sets the IAID

dhcpv6/async/client_test.go
pmazzini: The DUID has the MAC inside it so it feels redundant.

dhcpv6/modifiers.go
pmazzini: Its fine, lets leave it like that

dhcpv6/dhcpv6message.go
```diff
@@ -43,7 +43,7 @@ func GetTime() uint32 {
 }
 
 // NewSolicitWithCID creates a new SOLICIT message with CID.
-func NewSolicitWithCID(duid Duid, modifiers ...Modifier) (*Message, error) {
+func NewSolicitWithCID(duid Duid, hwaddr net.HardwareAddr, modifiers ...Modifier) (*Message, error) {
```
pmazzini: How about removing `NewSolicitWithCID` and leaving only `NewSolicit` which takes a mac as parameter?

It doesn't make much sense to me to have a `NewSolicitWithCID` which takes a duid and a mac.

If we want to change the CID we can use the `WithClientID` modifier.

pmazzini: Otherwise, the default duid will be DUID_LLT as specified in line 77.

insomniacslk: I am OK with that, I also think it makes more sense. CC @hugelgupf in case you may be impacted by this

insomniacslk: I am OK with that, I also think it makes more sense. CC @hugelgupf in case you may be impacted by this

insomniacslk: worth doing in a separate PR?

insomniacslk: updated

dhcpv6/dhcpv6message.go
```diff
@@ -42,8 +41,9 @@ func GetTime() uint32 {
 	return uint32((now.Nanoseconds() / 1000000000) % 0xffffffff)
 }
 
-// NewSolicitWithCID creates a new SOLICIT message with CID.
-func NewSolicitWithCID(duid Duid, modifiers ...Modifier) (*Message, error) {
+// NewSolicit creates a new SOLICIT message, using the given hardware address to
+// derive the IAID in the IA_NA option.
+func NewSolicit(duid Duid, hwaddr net.HardwareAddr, modifiers ...Modifier) (*Message, error) {
```
pmazzini: What about removing the duid parameter and generating a default one?

pmazzini: Based on the hwaddr

insomniacslk: yeah, DUID_LLT generated with GetTime and MAC address, sounds good

dhcpv6/dhcpv6_test.go
```diff
@@ -144,6 +144,7 @@ func TestNewAdvertiseFromSolicit(t *testing.T) {
 	cid := OptClientId{}
 	s.AddOption(&cid)
 	duid := Duid{}
+	WithIAID([4]byte{1, 2, 3, 4})(&s)
```
pmazzini: Is this needed?

dhcpv6/dhcpv6message.go
```diff
@@ -57,31 +64,20 @@ func NewSolicitWithCID(duid Duid, modifiers ...Modifier) (*Message, error) {
...
-	iaNa.T1 = 0xe10
-	iaNa.T2 = 0x1518
-	m.AddOption(iaNa)
+	if len(hwaddr) < 4 {
+		return nil, errors.New("short hardware addrss: less than 4 bytes")
+	}
+	l := len(hwaddr)
+	var iaid [4]byte
+	copy(iaid[:], hwaddr[l-4:l])
+	modifiers = append([]Modifier{WithIAID(iaid)}, modifiers...)
```
pmazzini: I am worried that if we append a modifier, it will always be applied after the ones passed by parameter. How about directly applying the modifier here?

dhcpv6/async/client_test.go
```diff
@@ -20,13 +19,7 @@ func solicit(input string) (*dhcpv6.Message, error) {
...
 		return nil, err
 	}
-	duid := dhcpv6.Duid{
-		Type:          dhcpv6.DUID_LLT,
-		HwType:        iana.HWTypeEthernet,
-		Time:          dhcpv6.GetTime(),
-		LinkLayerAddr: mac,
-	}
-	return dhcpv6.NewSolicitWithCID(duid)
+	return dhcpv6.NewSolicit(mac, dhcpv6.WithIAID([4]byte{1, 2, 3, 4}))
```
pmazzini: Is the `WithIAID` needed?

dhcpv6/dhcpv6relay_test.go
```diff
@@ -90,6 +90,7 @@ func TestNewRelayRepFromRelayForw(t *testing.T) {
 	s, err := NewMessage()
 	require.NoError(t, err)
 	s.AddOption(&OptClientId{})
+	WithIAID([4]byte{1, 2, 3, 4})(s)
```
pmazzini: Is this needed?

netboot/netconf_test.go
```diff
@@ -23,25 +23,29 @@ func getAdv(modifiers ...dhcpv6.Modifier) *dhcpv6.Message {
 		Time:          dhcpv6.GetTime(),
 		LinkLayerAddr: hwaddr,
 	}
-	sol, err := dhcpv6.NewSolicitWithCID(duid, modifiers...)
+	mac, err := net.ParseMAC("aa:bb:cc:dd:ee:ff")
 	if err != nil {
 		log.Panic(err)
 	}
-	d, err := dhcpv6.NewAdvertiseFromSolicit(sol, modifiers...)
+	sol, err := dhcpv6.NewSolicitWithCID(duid, mac, solModifiers...)
```
pmazzini: We no longer have `NewSolicitWithCID`.

dhcpv6/dhcpv6message.go
insomniacslk: the modifier here is prepended, not appended, so it will always be applied first (and can be overridden)

pmazzini: Oh, right, thanks for the clarification.

netboot/netconf_test.go
insomniacslk: done

netboot/netconf_test.go
```diff
@@ -7,41 +7,37 @@ import (
 
 	"github.com/insomniacslk/dhcp/dhcpv4"
 	"github.com/insomniacslk/dhcp/dhcpv6"
-	"github.com/insomniacslk/dhcp/iana"
 	"github.com/stretchr/testify/require"
 )
 
-func getAdv(modifiers ...dhcpv6.Modifier) *dhcpv6.Message {
+func getAdv(solModifiers, advModifiers []dhcpv6.Modifier) *dhcpv6.Message {
```
pmazzini: Why does it take two different lists of modifiers? If we are not using it, I'll rather leave it how it was before.

pmazzini:
One last comment, the rest looks good.

netboot/netconf_test.go
insomniacslk: some modifiers must be applied only to the advertise - they were applied to the solicit too. With the latest iteration no modifiers are applied to the solicit anymore, so this could go away (until needed again)

pmazzini: If it is not used, I would remove it :P

insomniacslk: it's removed in the last push

netboot/netconf_test.go
```diff
@@ -7,27 +7,23 @@ import (
...
 		log.Panic(err)
 	}
 
-	duid := dhcpv6.Duid{
-		Type:          dhcpv6.DUID_LLT,
-		HwType:        iana.HWTypeEthernet,
-		Time:          dhcpv6.GetTime(),
-		LinkLayerAddr: hwaddr,
+	if err != nil {
+		log.Panic(err)
```
pmazzini: This is duplicated

pmazzini: Probably a left over

insomniacslk: copy-pasta error, good catch

insomniacslk:
thanks for the thorough review @pmazzini !

PR #278, opened 2019-04-29
[server6] Server should join multicast address

insomniacslk:
The previous logic was wrong - there's no "listening" on multicast
address, the server should listen on the given address, and join the
multicast group. This PR fixes it.
Also moved the multicast addresses to a common package.

Tested with unit/integ tests, and with coredhcp.

### DIFF
--- a/dhcpv6/server6/server.go
+++ b/dhcpv6/server6/server.go
@@ -58,6 +58,7 @@ import (
 	"net"
 
 	"github.com/insomniacslk/dhcp/dhcpv6"
+	"golang.org/x/net/ipv6"
 )
 
 // Handler is a type that defines the handler function to be called every time a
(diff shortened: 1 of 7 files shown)


dhcpv6/server6/server.go
```diff
@@ -122,10 +127,29 @@ func NewServer(addr *net.UDPAddr, handler Handler, opt ...ServerOpt) (*Server, e
...
+			iface, err = net.InterfaceByName(ifname)
+			if err != nil {
+				return nil, err
+			}
+		}
+		group := net.UDPAddr{
+			IP:   dhcpv6.AllDHCPRelayAgentsAndServers,
+			Port: dhcpv6.DefaultServerPort,
+		}
+		p := ipv6.NewPacketConn(conn)
```
hugelgupf: This seems weird to me - you can make a packet conn out of an existing UDP packet conn?

This doesn't do another layer of marshaling UDP packets on top of UDP packets, does it?

insomniacslk: I believe this just joins the multicast group and discards, I'll double-check in the code before merging, after the wall of meetings I have now :(

insomniacslk: (and the CI failures of course)

insomniacslk: oh the CI failure is for router7, I'll make a PR

insomniacslk:
CI failure related to router7, I'll make a PR.
I'll also check the code of `ipv6.NewPacketConn` soon to see if nothing bad happens with the conn

dhcpv6/server6/server.go
insomniacslk: According to the documentation at https://godoc.org/golang.org/x/net/ipv6 it should be fine, looking at the code now

insomniacslk: seems just a wrapper, I think it's safe. I'll merge for now (we can unland if we find subtleties later)

insomniacslk:
dependency fixed at https://github.com/rtr7/router7/pull/30

Issue #123, opened 2018-08-05
Versions for Go modules?


stapelberg:
Commit https://github.com/insomniacslk/dhcp/commit/c6894ea160d82b4a326ebab94faaebd090192a7b broke my project https://github.com/rtr7/router7

I realize that currently, this repository does not have any tagged versions.

As per https://research.swtch.com/vgo-module, could you please start releasing a new major version for breaking changes?

Thanks for considering,

insomniacslk:
Hey Michael,
I'm sorry that this broke your build, I see that you have already committed an update on your side. We planned already to use vgo, as this project is now more widely used. For now I will commit to avoid breaking changes until we have switched to versioning. Would that be OK for you?

stapelberg:
Sure thing, thanks for your reply! :)

insomniacslk:
We are not far from a stable 1.0 release :) Will get back to this soon

stapelberg:
After PR #279, `github.com/u-root/u-root` is the only dependency showing up as incompatible when enabling Go modules:

```
cat go.mod
module github.com/insomniacslk/dhcp

go 1.12

require (
	github.com/fanliao/go-promise v0.0.0-20141029170127-1890db352a72
	github.com/hugelgupf/socketpair v0.0.0-20190312070548-ac4bdefbf625
	github.com/jsimonetti/rtnetlink v0.0.0-20190429090657-99ed6b5c17d2
	github.com/mdlayher/ethernet v0.0.0-20190313224307-5b5fc417d966
	github.com/mdlayher/netlink v0.0.0-20190423160922-382af5ec7c57
	github.com/mdlayher/raw v0.0.0-20190419142535-64193704e472
	github.com/stretchr/testify v1.3.0
	github.com/u-root/u-root v4.0.0+incompatible
	golang.org/x/net v0.0.0-20190424112056-4829fb13d2c6
	golang.org/x/sys v0.0.0-20190429094411-2cc0cad0ac78
)
```

I think this is because u-root is using major version numbers without following the semantic import versioning convention (i.e. /v4 in the import path for major version 4). I would guess that u-root doesn’t really consider itself a library package.

However, it seems like u-root is only pulled in for the uio package, which provides endianness conversion. Those conversions are easily expressed in a helper package (see e.g. https://github.com/google/nftables/blob/master/binaryutil/binaryutil.go), which seems preferable to an external dependency.

hugelgupf:
[...]

PR #279, opened 2019-04-29
switch from u-root/pkg/rand to crypto/rand

stapelberg:
As per its documentation, u-root’s rand package implements cancelable reads,
which is not something that this DHCP client uses.

The crypto/rand package has the advantage of using getrandom(2) on platforms
where that is possible, and otherwise behaves exactly like u-root/pkg/rand.

Notably, this commit fixes the build when running with GO111MODULES=on, so it’s
related to issue #123.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -17,6 +17,7 @@ package dhcpv4
 
 import (
 	"bytes"
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"net"
@@ -25,7 +26,6 @@ import (
 
 	"github.com/insomniacslk/dhcp/iana"
 	"github.com/insomniacslk/dhcp/rfc1035label"
-	"github.com/u-root/u-root/pkg/rand"
 	"github.com/u-root/u-root/pkg/uio"
 )
 
--- a/dhcpv6/dhcpv6_test.go
+++ b/dhcpv6/dhcpv6_test.go
@@ -1,6 +1,7 @@
 package dhcpv6
 
 import (
+	"crypto/rand"
 	"encoding/binary"
 	"errors"
 	"net"
@@ -9,7 +10,6 @@ import (
 	"github.com/insomniacslk/dhcp/iana"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
-	"github.com/u-root/u-root/pkg/rand"
 )
 
 func randomReadMock(value []byte, n int, err error) func([]byte) (int, error) {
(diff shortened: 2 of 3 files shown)


hugelgupf:
u-root's rand uses getrandom.

The package was requested by @insomniacslk - is your usage of it internal-only at this point?

I do plan to make use of the cancelable functionality. We often end up on systems that do not have a random number source initially available, seeded, or configured, and this package provides nicer diagnostics than the print-and-hang-forever that crypto/rand will do.

stapelberg:
> u-root's rand uses getrandom.

Ah, I missed this because godoc referred me to random_urandom.go. Nevermind :)

> I do plan to make use of the cancelable functionality. We often end up on systems that do not have a random number source initially available and this package provides nicer diagnostics than the print-and-hang-forever that crypto/rand will do.

That’s fair. Might be good to suggest said nicer diagnostics upstream for crypto/rand, too, though :)

insomniacslk:
> Might be good to suggest said nicer diagnostics upstream for crypto/rand, too, though :)

Go 1.12 introduced a 60-seconds timeout before printing a warning, which *may* be good enough if we only wanted to know what's going on. I still think it's valuable to cancel the random number generation, and I like your idea of submitting the proposition upstream. What do you think @hugelgupf ?

(Also CC @mdlayher )

mdlayher:
+1 for at least filing an issue with upstream to explain the use case.

stapelberg:
> still think it's valuable to cancel the random number generation, and I like your idea of submitting the proposition upstream

To be clear, I only proposed submitting better diagnostics upstream. I’m not sure yet whether canceling random number reading is really a good idea. If random numbers are optional, why read them in the first place? If they’re not, why cancel?

insomniacslk:
> To be clear, I only proposed submitting better diagnostics upstream.

I know, sorry for the confusion. In my (mis-worded) reply I meant that some form of diagnostic is already available starting with Go 1.12, and that I'd like the idea of submitting upstream a feature I find useful (cancellation).

> I’m not sure yet whether canceling random number reading is really a good idea. If random numbers are optional, why read them in the first place? If they’re not, why cancel?

A scenario where this may be useful is when you need to do something with those numbers within a certain amount of time, or fail and do something else. This is exactly my use case: in the early seconds of booting, a host may or may not have enough entropy (that I use for DHCP transactions and other cryptographic operations), and I need to:
* detect whether we are stuck on random number generation (Go 1.12 now handles this, but it wasn't available before)
* react and report the error in some way.

However I agree that the feature may be considered not useful enough to be part of the standard library, as there are not many scenarios with these needs.

PR #281, opened 2019-05-02
[dhcpv6] Handle SOLICIT with rapid commit in NewReplyFromMessage

insomniacslk:
NewReplyFromMessage creates a reply from a message. With this patch it
can handle Solicit messages with rapid-commit.

RFC3315 states that:
```
   If the client has included a Rapid Commit option in its Solicit
   message, the client terminates the waiting process as soon as a Reply
   message with a Rapid Commit option is received.
```

### DIFF
--- a/dhcpv6/dhcpv6message.go
+++ b/dhcpv6/dhcpv6message.go
@@ -161,17 +161,23 @@ func NewRequestFromAdvertise(adv *Message, modifiers ...Modifier) (*Message, err
 }
 
 // NewReplyFromMessage creates a new REPLY packet based on a
-// Message. The function is to be used when generating a reply to
-// REQUEST, CONFIRM, RENEW, REBIND, RELEASE and INFORMATION-REQUEST packets.
+// Message. The function is to be used when generating a reply to a SOLICIT with
+// rapid-commit, REQUEST, CONFIRM, RENEW, REBIND, RELEASE and INFORMATION-REQUEST
+// packets.
 func NewReplyFromMessage(msg *Message, modifiers ...Modifier) (*Message, error) {
 	if msg == nil {
-		return nil, errors.New("Message cannot be nil")
+		return nil, errors.New("message cannot be nil")
 	}
 	switch msg.Type() {
+	case MessageTypeSolicit:
+		if msg.GetOneOption(OptionRapidCommit) == nil {
+			return nil, errors.New("cannot create REPLY from a SOLICIT without rapid-commit option")
+		}
+		modifiers = append([]Modifier{WithRapidCommit}, modifiers...)
 	case MessageTypeRequest, MessageTypeConfirm, MessageTypeRenew,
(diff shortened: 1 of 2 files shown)


insomniacslk:
the CI failure is unrelated, it complains that the output of `DHCPv4.Summary` is different from the expected. I think this broke in a previous PR on `dhcpv4/options.go`.

insomniacslk:
cc @hugelgupf

dhcpv6/dhcpv6message.go
```diff
@@ -161,17 +161,23 @@ func NewRequestFromAdvertise(adv *Message, modifiers ...Modifier) (*Message, err
 }
 
 // NewReplyFromMessage creates a new REPLY packet based on a
-// Message. The function is to be used when generating a reply to
-// REQUEST, CONFIRM, RENEW, REBIND, RELEASE and INFORMATION-REQUEST packets.
+// Message. The function is to be used when generating a reply to a SOLICIT with
+// rapid-commit, REQUEST, CONFIRM, RENEW, REBIND, RELEASE and INFORMATION-REQUEST
+// packets.
```
pmazzini: I wonder if we should merge `NewReplyFromMessage` and `NewAdvertiseFromSolicit` to avoid having to figure out which one to use.

insomniacslk: I think not - NewReplyFromMessage explicitly builds a Reply from any message, and fails if it's not acceptable by the protocol. NewAdvertiseFromSolicit instead is specific about solicit, and only to be used for the "old" four-way transactions. Collapsing Advertise/Reply from Solicit to me sounds more complicated

insomniacslk:
Re. the CI, interestingly this only fails on `tip`, but it works on Go 1.10, 1.11 and 1.12

dhcpv6/dhcpv6message.go
pmazzini: I am thinking about a server handler when it receives a solicit: How should it build a response? Using `NewReplyFromMessage` or `NewAdvertiseFromSolicit`? Should the server handler check if the solicit has a rapid commit option? If so it will duplicate what we are already doing here. Why not merging them instead and having a function that takes a solicit and returns whatever needs to be returned.

insomniacslk:
as suggested by @hugelgupf , it *may* be a bug in tip. I think this is safe to merge, since no production version of Go is affected. If this is not a transient thing, it will show up again. At least we don't block this PR. CC @pmazzini

dhcpv6/dhcpv6message.go
insomniacslk: I made this PR exactly for that reason. The server handler does something like
```
if msg.GetOneOption(OptionRapidCommit) {
    resp, err = dhcpv6.NewReplyFromMessage(msg)
} else {
  resp, err = dhcpv6.NewAdvertiseFromSolicit(msg)
}
```
At some point in the future I see NewAdvertiseFromSolicit going away, but until then this seems the straightforward approach

PR #282, opened 2019-05-07
Adding RemoteID option while creating a Relayed Encapsulated Packet 

akshaynawale:
The dhcpv6 client was not updating the Remote ID option to the outer encapsulated packet, for SimulateRelay Flag. This was causing problems in testing dhcp requests with Remote ID. As Remote ID is set by the Relay Agent so it should be present at the outer encapsulated packet. This change will copy the inner message's Remote ID option if present, to the outer encapsulated message.

Tested by running unit tests. 
```
go test 
2019/05/07 15:58:22 DNS OptDNSRecursiveNameServer{nameservers=[fe80::1 fe80::2]}
PASS
ok      _/home/navale/go/src/github.com/akshaynawale/dhcp/dhcpv6        0.009s
```

### DIFF
--- a/dhcpv6/client6/client.go
+++ b/dhcpv6/client6/client.go
@@ -92,7 +92,7 @@ func (c *Client) sendReceive(ifname string, packet dhcpv6.DHCPv6, expectedType d
 	}
 	if c.SimulateRelay {
 		var err error
-		encapPacket, err = dhcpv6.EncapsulateRelay(packet, dhcpv6.MessageTypeRelayForward, net.IPv6zero, laddr.IP)
+		encapPacket, err := dhcpv6.EncapsulateRelay(packet, dhcpv6.MessageTypeRelayForward, net.IPv6zero, laddr.IP)
 		if err != nil {
 			return nil, err
 		}
--- a/dhcpv6/client6/client.go
+++ b/dhcpv6/client6/client.go
@@ -92,10 +92,15 @@ func (c *Client) sendReceive(ifname string, packet dhcpv6.DHCPv6, expectedType d
 	}
 	if c.SimulateRelay {
 		var err error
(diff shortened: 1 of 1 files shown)


insomniacslk:
Thanks @akshaynawale ! Can you please amend the commit to include the Signed-Off-By footer? `git commit --amend -s` should suffice

insomniacslk:
The CI is also failing<;
```
dhcpv6/client6/client.go:95:3: undefined: encapPacket
dhcpv6/client6/client.go:101:4: undefined: encapPacket
dhcpv6/client6/client.go:103:12: undefined: encapPacket
```

akshaynawale:
Yes forgot to initialize the encapPacket. I have fixed it now. I also tried to fix the signoff message but tje DCO is stilling failing. I dont know if I am doing anything wrong.

akshaynawale:
Signed off all the commits.

pmazzini:
> As Remote ID is set by the Relay Agent so it should be present at the outer encapsulated packet.

I am not following this. In this PR you are copying the Remote ID. When simulating a relay, who populates the Remote ID?

pmazzini:
I see the issue now. You are probably using a modifier to add a remote id but it is being added to the `Message` and not to the `Relay`, and with this PR you are copying it over. However, it doesn't make sense to add a remote id to a dhcpv6 `Message`.

I don't see a nice solution for this. How about adding the remote id to the client struct, similar to what we are doing with the `SimulateRelay` bool?

akshaynawale:
Make sense Pablo. I will update this.

akshaynawale:
Added a field in client called RelayOptions which is dhcpv6.Options and can be used to pass options we need to add to the relay message.

pmazzini:
Thanks!

PR #286, opened 2019-05-08
Improve compatibility with some ipv4 networks.

lprylli:
- dnsmasq has been seen to null-terminate the bootfile option.
- for the gateway/dns information to be present in final packet,
the Router option should be queried again in request as
in discover (which matches behavior of udhcpc/dhclient).

Tested: pxeboot with u-root on dnsmask/ipv4 client.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -242,6 +242,12 @@ func NewRequestFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) 
 		WithClientIP(offer.ClientIPAddr),
 		WithOption(OptRequestedIPAddress(offer.YourIPAddr)),
 		WithOption(OptServerIdentifier(serverIP)),
+		WithRequestedOptions(
+			OptionSubnetMask,
+			OptionRouter,
+			OptionDomainName,
+			OptionDomainNameServer,
+		),
 	)...)
 }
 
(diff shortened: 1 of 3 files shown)


insomniacslk:
Thanks @lprylli ! Can you please amend the commit with DCO?

dhcpv4/dhcpv4.go
```diff
@@ -574,7 +580,8 @@ func (d *DHCPv4) RootPath() string {
 //
 // The Bootfile Name option is described by RFC 2132, Section 9.5.
 func (d *DHCPv4) BootFileNameOption() string {
-	return GetString(OptionBootfileName, d.Options)
+	name := GetString(OptionBootfileName, d.Options)
+	return strings.TrimRight(name, "\x00")
```
insomniacslk: that's sad.. I wonder if we should notify the user of this or silently swallow the null bytes. Thank you null-terminated strings? :(

lprylli: As RFC2132 states "Options containing NVT ASCII data SHOULD NOT include a trailing NULL; however, the receiver of such options MUST be prepared to delete trailing nulls if they exist." :-)

insomniacslk: sounds good :)

insomniacslk:
Alright, the PR looks good. I'll wait for tests and DCO before merging

insomniacslk:
Also, we should look for other options using NVT ASCII and do the same. Maybe a separate PR is ok

dhcpv4/dhcpv4.go
hugelgupf: Do the same for the TFTPServerName option?

lprylli: done

lprylli:
Adding tftpservername as suggested (it is really symmetric in that it is the only one dnsmasq also treats specially and null-terminate, although probably not very commonly used).

Added DCO.

There is probably a case for moving up a level and doing null-trimming in GetString, but I was trying to be conservative, minimize the risk of regression, and the scope of changes to case I tested in a "real" common scenario with dnsmasq. So handling all NVT-ASCII string might be better done in separate PR.

lprylli:
Did an extra push (

pmazzini:
Why did you close the PR?

lprylli:
sorry fat finger, not that much practive with github PR, pressed the wrong "comment and close" button instead of just "comment" intended.

insomniacslk:
Thanks @lprylli, I'll review again tomorrow during daytime

pmazzini:
I am not merging it yet to let @insomniacslk have a look as well.

PR #287, opened 2019-05-08
[dhcpv4] Do not set Gateway IP address on packets sent by clients

marcoguerri:
Relays might drop packets coming from clients if they have the Gateway IP set. This modifier is supposed to be used by relays: `WithReply` is used only by clients.

### DIFF
--- a/dhcpv4/dhcpv4_test.go
+++ b/dhcpv4/dhcpv4_test.go
@@ -256,7 +256,6 @@ func TestNewReplyFromRequest(t *testing.T) {
 	reply, err := NewReplyFromRequest(discover)
 	require.NoError(t, err)
 	require.Equal(t, discover.TransactionID, reply.TransactionID)
-	require.Equal(t, discover.GatewayIPAddr, reply.GatewayIPAddr)
 }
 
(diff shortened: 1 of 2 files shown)


insomniacslk:
CI fails on
```
package github.com/u-root/u-root/pkg/rand: cannot find package "github.com/u-root/u-root/pkg/rand" in any of:
	/home/travis/.gimme/versions/go1.11.linux.amd64/src/github.com/u-root/u-root/pkg/rand (from $GOROOT)
	/home/travis/gopath/src/github.com/u-root/u-root/pkg/rand (from $GOPATH)
```
We should drop Go 1.11 soon.

pmazzini:
Thanks for the fix!

marcoguerri:
Closed and reopen PR to trigger again Travis following @pmazzini 's suggestion (failed on a transient issue).

pmazzini:
Go 1.11 is happy now.

insomniacslk:
Oh good, I thought it was due to u-root not supporting 1.11 anymore (which we should drop soon anyway, 1.13 is coming)

PR #289, opened 2019-05-14
Remove SHIFT IN character bytes from Circuit ID

akshaynawale:
Some Vendors like Arista sends SHIFT IN (0x000f) prefix to the circuit id. This used to fail the circuit id parsing logic. This check will remove the 0x000f bytes prefix from the circuit id field if present. 
Added unit test for the same.
Signed-off-by: Akshay Navale <navale@fb.com>

```
go test
PASS
ok      _/home/navale/go/src/github.com/akshaynawale/dhcp/dhcpv4/ztpv4  0.004s
```

### DIFF
--- a/dhcpv4/ztpv4/parse_circuitid.go
+++ b/dhcpv4/ztpv4/parse_circuitid.go
@@ -51,9 +51,7 @@ func ParseCircuitID(packet *dhcpv4.DHCPv4) (*CircuitID, error) {
 	cid := relayOptions.Options.Get(dhcpv4.AgentCircuitIDSubOption)
 	// Some Vendor like Arista sends SHIFT IN character i.e. 0x000f before circuitid
 	// remove it before matching against regexps.
-	if bytes.HasPrefix(cid, []byte{0x00, 0x0f}) {
-		cid = bytes.TrimPrefix(cid, []byte{0x00, 0x0f}) 
-	}
+	cid = bytes.TrimPrefix(cid, []byte{0x00, 0x0f}) 
 	circuitIdStr := string(cid)
 	if circuitIdStr == "" {
 		return nil, fmt.Errorf("no circuit-id suboption found in dhcpv4 packet")
--- a/dhcpv4/ztpv4/parse_circuitid.go
+++ b/dhcpv4/ztpv4/parse_circuitid.go
@@ -1,6 +1,7 @@
 package ztpv4
 
 import (
+	"bytes"
 	"fmt"
 	"regexp"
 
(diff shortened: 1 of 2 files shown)


dhcpv4/ztpv4/parse_circuitid_test.go
```diff
@@ -85,6 +86,7 @@ func TestParseCircuitID(t *testing.T) {
 		{name: "Cisco pattern", circuit: []byte("Gi1/10:2020"), want: &CircuitID{Slot: "1", Port: "10", Vlan: "2020"}},
 		{name: "Cisco Nexus pattern", circuit: []byte("Ethernet1/3"), want: &CircuitID{Slot: "1", Port: "3"}},
 		{name: "Juniper Bundle Pattern", circuit: []byte("ae52.0"), want: &CircuitID{Port: "52", SubPort: "0"}},
+		{name: "Arista Vlan pattern 1 with SHIFT IN", circuit: bytes.Join([][]byte{[]byte{0x00, 0x0f}, []byte("Ethernet14:Vlan2001")}, []byte{}), want: &CircuitID{Port: "14", Vlan: "Vlan2001"}},
```
pmazzini: What is the `[]byte{}` for in the last parameter of bytes.Join?

pmazzini:
Minor NIT. The rest looks good. Thank you @akshaynawale.

dhcpv4/ztpv4/parse_circuitid.go
```diff
@@ -47,7 +48,13 @@ func ParseCircuitID(packet *dhcpv4.DHCPv4) (*CircuitID, error) {
 
 	// As per RFC 3046 sub-Option 1 is circuit-id. Look at 2.0 section in that RFC
 	// https://tools.ietf.org/html/rfc3046
-	circuitIdStr := string(relayOptions.Options.Get(dhcpv4.AgentCircuitIDSubOption))
+	cid := relayOptions.Options.Get(dhcpv4.AgentCircuitIDSubOption)
+	// Some Vendor like Arista sends SHIFT IN character i.e. 0x000f before circuitid
+	// remove it before matching against regexps.
+	if bytes.HasPrefix(cid, []byte{0x00, 0x0f}) {
```
insomniacslk: you can simply do `cid = bytes.TrimPrefix(cid, []byte{0x00, 0x0f})` here, according to the documentation:
```
TrimPrefix returns s without the provided leading prefix string. If s doesn't start
with prefix, s is returned unchanged.
```

dhcpv4/ztpv4/parse_circuitid_test.go
insomniacslk: it's the separator used to join the two byte sequences - empty separator, basically it concatenates.

However I would recommend to change the whole call to `bytes.Join` to a simpler `[]byte("\x00'x0fEthernet14:Vlan2001")`

insomniacslk:
@pmazzini please also have a last look. If you merge, please select squash and merge, so this becomes a single commit

PR #291, opened 2019-05-15
Adding Juniper EX pattern for circuit parsing

akshaynawale:
This change will allow us to parse Juniper EX series interface pattern. 
Adding the regex and unit test for the same.
Tested with unit tests.
```
go test
PASS
ok      _/home/navale/go/src/github.com/akshaynawale/dhcp/dhcpv4/ztpv4  0.004s
```

Signed-off-by: Akshay Navale <navale@fb.com>


pmazzini:
Thanks!

akshaynawale:
Thanks for the quick review Pablo!

PR #292, opened 2019-05-16
Ignore bytes after end of IP packet in BroadcastRawUDPConn.ReadFrom

lprylli:
When reading raw packets from the network, it can happen that the raw
ethernet packet read has undefined bytes after the end of the ip
packet (either from the network or in some cases from the local
stack).

Those bytes should not be passed to the dhcp-receiver otherwise the
option parser which is picky about final padding byte will silently
discard the dhcp-reply.

Tested: ast2500 bmc reproducing the issue + existing go test for coverage.
Signed-off-by: Loic Prylli <lprylli@netflix.com>

### DIFF
--- a/dhcpv4/nclient4/conn_linux.go
+++ b/dhcpv4/nclient4/conn_linux.go
@@ -115,11 +115,11 @@ func udpMatch(addr *net.UDPAddr, bound *net.UDPAddr) bool {
 // ReadFrom reads raw IP packets and will try to match them against
 // upc.boundAddr. Any matching packets are returned via the given buffer.
 func (upc *BroadcastRawUDPConn) ReadFrom(b []byte) (int, net.Addr, error) {
-	ipLen := IPv4MaximumHeaderSize
-	udpLen := UDPMinimumSize
+	ipHdrMaxLen := IPv4MaximumHeaderSize
+	udpHdrLen := UDPMinimumSize
 
 	for {
-		pkt := make([]byte, ipLen+udpLen+len(b))
+		pkt := make([]byte, ipHdrMaxLen+udpHdrLen+len(b))
 		n, _, err := upc.PacketConn.ReadFrom(pkt)
 		if err != nil {
 			return 0, nil, err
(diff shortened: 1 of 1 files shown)


pmazzini:
@lprylli Thanks for your contribution.

The other alternative is to stop reading after the `optEnd` https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/options.go#L158

However, the approach of this PR is probably better.

@hugelgupf What is your call?

dhcpv4/nclient4/conn_linux.go
```diff
@@ -150,7 +150,10 @@ func (upc *BroadcastRawUDPConn) ReadFrom(b []byte) (int, net.Addr, error) {
 			IP:   net.IP(ipHdr.SourceAddress()),
 			Port: int(udpHdr.SourcePort()),
 		}
-		return copy(b, buf.ReadAll()), srcAddr, nil
+		// Extra padding after end of IP packet shoudl be ignored,
```
pmazzini: s/shoudl/should/

dhcpv4/nclient4/conn_linux.go
```diff
@@ -150,7 +150,10 @@ func (upc *BroadcastRawUDPConn) ReadFrom(b []byte) (int, net.Addr, error) {
...
+		// if not dhcp option parsing will fail.
+		dhcpLen := int(ipHdr.PayloadLength()) - UDPMinimumSize
```
pmazzini: Maybe use `udpLen` here to be consistent with line 140.

pmazzini:
I commented with a couple of NITs. The rest looks good to me. Thanks.

dhcpv4/nclient4/conn_linux.go
lprylli: done

dhcpv4/nclient4/conn_linux.go
lprylli: done, renamed ipLen to ipMaxHdrLen, and udpLen to udpHdrLen to avoid any confusion.

pmazzini:
Thanks!

PR #294, opened 2019-06-19
dhcpv4: make the short stringer more useful

hugelgupf:
Nobody gives a hoot about the hardware type.


hugelgupf:
I think I removed that one. I don't understand the CI failure that's there though

insomniacslk:
@hugelgupf , same here as in #295 for CI, this is OK to merge for me, and I'll address the unrelated CI failure in another PR

PR #295, opened 2019-06-19
nclient4: add logging options

hugelgupf:

### DIFF
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"os"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


hugelgupf:
cc @stapelberg

dhcpv4/nclient4/client.go
```diff
@@ -225,6 +260,24 @@ func WithTimeout(d time.Duration) ClientOpt {
...
+func WithSummaryLogger() ClientOpt {
+	return func(c *Client) {
+		c.logger = shortSummaryLogger{
+			Logger: log.New(os.Stderr, "[dhcpv4]", log.LstdFlags),
+		}
+	}
+}
+
+// WithDebugLogger logs multi-line full DHCPv4 messages when sent & received.
+func WithDebugLogger() ClientOpt {
```
stapelberg: You could add another option which takes an exported logger interface containing only Printf so that users can pass in e.g. a custom log.Logger.

insomniacslk:
@hugelgupf the CI error is unrelated to your PR, and I'm OK to merge it if you don't plan to make further changes here. I wonder if eth0 is no more available in Travis-CI, but I will address this in a separate PR

PR #296, opened 2019-06-19
nclient4: real exponential backoff

hugelgupf:

### DIFF
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -16,6 +16,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"math/rand"
 	"net"
 	"strings"
 	"sync"
@@ -67,7 +68,7 @@ type Client struct {
 	ifaceHWAddr net.HardwareAddr
 	conn        net.PacketConn
 	timeout     time.Duration
-	retry       int
+	retries     uint
 
 	// bufferCap is the channel capacity for each TransactionID.
 	bufferCap int
@@ -112,7 +113,7 @@ func NewWithConn(conn net.PacketConn, ifaceHWAddr net.HardwareAddr, opts ...Clie
 	c := &Client{
 		ifaceHWAddr: ifaceHWAddr,
 		timeout:     defaultTimeout,
-		retry:       defaultRetries,
+		retries:     defaultRetries,
 		serverAddr:  DefaultServers,
 		bufferCap:   defaultBufferCap,
 		conn:        conn,
@@ -233,10 +234,12 @@ func withBufferCap(n int) ClientOpt {
 
 // WithRetry configures the number of retransmissions to attempt.
 //
+// Negative retries means infinite retries.
+//
 // Default is 3.
 func WithRetry(r int) ClientOpt {
 	return func(c *Client) {
-		c.retry = r
+		c.retries = u
 	}
 }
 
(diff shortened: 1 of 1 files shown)


dhcpv4/nclient4/client.go
```diff
@@ -233,10 +234,16 @@ func withBufferCap(n int) ClientOpt {
 
 // WithRetry configures the number of retransmissions to attempt.
 //
+// 0 retries means infinite retries.
+//
 // Default is 3.
 func WithRetry(r int) ClientOpt {
```
insomniacslk: what about requiring an uint here? The user will get a compile time error if a negative integer is passed, a bit annoying (more typing) but safer

dhcpv4/nclient4/client.go
```diff
@@ -233,10 +234,16 @@ func withBufferCap(n int) ClientOpt {
 
 // WithRetry configures the number of retransmissions to attempt.
 //
+// 0 retries means infinite retries.
```
insomniacslk: I understand "zero retries" as "try just once, then stop" - same as above with "tries" vs. "retries". I'd rather keep using the concept of retries here

insomniacslk: which also means that 0 is a valid number of retries, and "loop forever" should be achieved differently (any negative number? Exactly -1 ? )

dhcpv4/nclient4/client.go
```diff
@@ -67,7 +68,7 @@ type Client struct {
 	ifaceHWAddr net.HardwareAddr
 	conn        net.PacketConn
 	timeout     time.Duration
-	retry       int
+	tries       uint
```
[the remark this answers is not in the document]
hugelgupf: In our usage below, we seem to actually do number of tries, not retries. I changed the variable name to correct reflect our usage of it

dhcpv4/nclient4/client.go
hugelgupf: I didn't want to disturb existing users, haha

dhcpv4/nclient4/client.go
hugelgupf: went back to retries, but i think 0 is kind of a useless value, so <= 0 is infinite

hugelgupf:
ptal

dhcpv4/nclient4/client.go
insomniacslk: How I see the API as a user:
* if retries is 0, I expect one attempt
* if retries is 1, I expect two attempts
* and so on

The current code treats retries as attempts, so 0 is not meaningful, but I still think there is an inconsistency between naming and behaviour, which could confuse the users. Hope it makes sense, and open to discuss :)

insomniacslk: hm, why marked as resolved?

insomniacslk:
there is also a small merge conflict with #295

tfg13:
Now the comment says `Negative retries means infinite retries`, but the loop uses `c.retries <= 0` :P

tfg13:
Or are we doing `retries = 1` means 1 total attempt now?

dhcpv4/nclient4/client.go
```diff
@@ -386,15 +389,15 @@ func (c *Client) retryFn(fn func(timeout time.Duration) error) error {
...
 		switch err := fn(timeout); err {
 		case nil:
 			// Got it!
 			return nil
 
 		case errDeadlineExceeded:
-			// Double timeout, then retry.
-			timeout *= 2
+			// Exponential backoff.
+			timeout = time.Duration(rand.Intn(1<<i)-1) * c.timeout
```
tfg13: Isn't the timeout always negative for the first iteration (when `i = 0`? Is this on purpose?)

hugelgupf:
Ryan's gonna take this over.

PR #299, opened 2019-06-27
v4 discover: don't ask for broadcast

hugelgupf:
Usually this is used for clients that don't know how to receive any
other packets. We can deal with both a unicast or broadcast response
packet, so let's let the server decide on its own.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -175,7 +175,6 @@ func NewDiscoveryForInterface(ifname string, modifiers ...Modifier) (*DHCPv4, er
 // HW type and specified hardware address.
 func NewDiscovery(hwaddr net.HardwareAddr, modifiers ...Modifier) (*DHCPv4, error) {
 	return New(PrependModifiers(modifiers,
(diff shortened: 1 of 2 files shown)


insomniacslk:
CI is failing - just remove the check for `m.IsBroadcast()` in `TestNewDiscovery`

insomniacslk:
The build looks to be still broken. Also, why there's Pablo's commit from another PR?

pmazzini:
> Also, why there's Pablo's commit from another PR?

It is not from another PR, I think that was me clicking the `update branch` button from the UI.

insomniacslk:
> > Also, why there's Pablo's commit from another PR?
> 
> It is not from another PR, I think that was me clicking the `update branch` button from the UI.

It's a feature I would like to disable because it causes more trouble than help (but I can't find how to disable it). @hugelgupf can you force-push to fix this?

hugelgupf:
If the merge commit is empty, github will automatically drop it when you hit "rebase and merge".

hugelgupf:
Updated, PTAL

PR #310, opened 2019-08-13
Bind interface fix

borna-blazevic:
Created a new bind to interface function for v4 server, and altered v4 server.go to make it more consistent with the v6 server.


pmazzini:
Thanks for the PR @borna-blazevic!

dhcpv4/server4/server.go
```diff
@@ -123,7 +123,7 @@ func WithConn(c net.PacketConn) ServerOpt {
 }
 
 // NewServer initializes and returns a new Server object
-func NewServer(addr *net.UDPAddr, handler Handler, opt ...ServerOpt) (*Server, error) {
+func NewServer(addr *net.UDPAddr, handler Handler, iface string, opt ...ServerOpt) (*Server, error) {
```
pmazzini: Can this be renamed to `func NewServer(ifname string, addr *net.UDPAddr, handler Handler, opt ...ServerOpt) (*Server, error) {`

so that it matches the v6 one: https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/server6/server.go#L120

dhcpv4/bindtointerface_linux.go
```diff
@@ -0,0 +1,47 @@
+// +build linux
```
pmazzini: I think this can be dropped in favor of `NewIPv4UDPConn` https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/nclient4/conn_linux.go#L33

dhcpv4/server4/server.go
```diff
@@ -133,7 +133,7 @@ func NewServer(addr *net.UDPAddr, handler Handler, opt ...ServerOpt) (*Server, e
 	}
 	if s.conn == nil {
 		var err error
-		conn, err := net.ListenUDP("udp4", addr)
+		conn, err := dhcpv4.BindConnectionToInterface(addr.String(), iface)
```
pmazzini: Same as above, I think here you can use `NewIPv4UDPConn` which is already in the lib.

pmazzini:
`nclient4/client_test.go` needs to be updated: https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/nclient4/client_test.go#L64

`server4/server_test.go` needs to be updated: https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/server4/server_test.go#L76

borna-blazevic:
> `nclient4/client_test.go` needs to be updated: https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/nclient4/client_test.go#L64
> 
> `server4/server_test.go` needs to be updated: https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/server4/server_test.go#L76

From where should I pass the interface in client_test.go?

pmazzini:
> From where should I pass the interface in client_test.go?

Which interface are you referring to? In `client_test.go` I think you should change line 64:

From:

```
s, err := server4.NewServer(nil, h.handle, server4.WithConn(serverConn))
```

To:

```
s, err := server4.NewServer("", nil, h.handle, server4.WithConn(serverConn))
```

pmazzini:
For reference, this PR comes from https://github.com/coredhcp/coredhcp/pull/40

borna-blazevic:
@pmazzini made the changes you asked for.

pmazzini:
Thanks for addressing those comments. Travis is now showing a circular dependency:

```
import cycle not allowed in test
package github.com/insomniacslk/dhcp/dhcpv4/nclient4 (test)
	imports github.com/insomniacslk/dhcp/dhcpv4/server4
	imports github.com/insomniacslk/dhcp/dhcpv4/nclient4
FAIL	github.com/insomniacslk/dhcp/dhcpv4/nclient4 [setup failed]
```

How about moving `NewIPv4UDPConn` from `nclient4/conn_linux.go` to `server4/conn_linux.go` since that function is only used by the server package.

I think that should solve the circular dependency problem.

We can later see a better organization for it. Maybe creating a `conn` package.

pmazzini:
`server4/server_test.go` needs to be updated https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/server4/server_test.go#L84

```
dhcpv4/server4/server_test.go:84:21: undefined: nclient4.NewIPv4UDPConn
```

dhcpv4/server4/server_test.go
```diff
@@ -73,15 +73,15 @@ func setUpClientAndServer(t *testing.T, iface net.Interface, handler Handler) (*
...
+	s, err := NewServer(iface.Name, &saddr, handler)
 	if err != nil {
 		t.Fatal(err)
 	}
 	go func() {
 		_ = s.Serve()
 	}()
 
-	clientConn, err := nclient4.NewIPv4UDPConn("", caddr.Port)
+	clientConn, err := dhcpv4.NewIPv4UDPConn("", caddr.Port)
```
pmazzini: This should be `NewIPv4UDPConn("", caddr.Port)` instead of `dhcpv4.NewIPv4UDPConn("", caddr.Port)`.

pmazzini: It is now in the same package.

dhcpv4/server4/server_test.go
```diff
@@ -73,15 +73,15 @@ func setUpClientAndServer(t *testing.T, iface net.Interface, handler Handler) (*
 		IP:   loAddr,
 		Port: randPort(),
 	}
-	s, err := NewServer(&saddr, handler)
+	s, err := NewServer(iface.Name, &saddr, handler)
```
pmazzini: How about doing:

```
s, err := NewServer("", &saddr, handler)
```

to avoid binding to the interface in the test.

```
server_test.go:78: cannot bind to interface lo: operation not permitted
```

dhcpv4/server4/conn_linux.go
```diff
@@ -0,0 +1,45 @@
+package server4
```
pmazzini: I have just noticed this is not linux specific so it can be renamed from `conn_linux` to `conn`

pmazzini:
lgtm after those 2 last comments

pmazzini:
Can `conn_linux.go` inside `server4` be renamed to `conn.go`? I've just noticed that is not linux specific.

pmazzini:
I am not merging it yet to let other people have a look as well.

borna-blazevic:
> Thanks for working on this @borna-blazevic! It looks ok to me.
> 
> Can you check if the v4 work that you are doing in [coredhcp/coredhcp#40](https://github.com/coredhcp/coredhcp/pull/40) works fine with this change?

Yes it works  fine

PR #312, opened 2019-08-14
server4: make conn public

pmazzini:
Use case: When writing a DHCP relay we create 2 `dhcpv4.Server`s. One for communicating with downstream clients and one for communicating with upstream servers. Each one binds to a different interface. In this diff we expose the connection so the opposite handler could reach the opposite bound interface.


pmazzini:
Right, using `WithConn` should be enough.

PR #314, opened 2019-09-12
Increase DHCPv4 IP TTL from 30 to 64

RossHanson:
RFC 1700 recommends a value of 64 for the default IP time to live (TTL)
parameter. This is only necessary for the V4 client because it uses
the net.PacketConn, while the IPv6 implementation uses a net.UDPConn instead.

From https://tools.ietf.org/html/rfc1700:

IP TIME TO LIVE PARAMETER

The current recommended default time to live (TTL) for the
Internet Protocol (IP) [45,105] is 64.


dhcpv4/nclient4/ipv4.go
```diff
@@ -351,7 +351,7 @@ func udp4pkt(packet []byte, dest *net.UDPAddr, src *net.UDPAddr) []byte {
 	ipv4fields := &IPv4Fields{
 		IHL:         IPv4MinimumSize,
 		TotalLength: uint16(ipLen + udpLen + len(packet)),
-		TTL:         30,
+		TTL:         64,
```
hugelgupf: Can you mention the rfc here, too?

RossHanson: Ah of course! Updated by force pushing an amended commit.

PR #230, opened 2019-01-17
Added osx to continuous integration

insomniacslk:
As discussed in https://github.com/insomniacslk/dhcp/issues/229 we need to test on all the OSes we support

Issue #229, opened 2019-01-17
Can't build/run on macOS due to client.go code


begetan:
```
go build
# github.com/insomniacslk/dhcp/dhcpv4
../../insomniacslk/dhcp/dhcpv4/client.go:125:25: undefined: unix.AF_PACKET
../../insomniacslk/dhcp/dhcpv4/client.go:125:68: undefined: unix.ETH_P_IP
../../insomniacslk/dhcp/dhcpv4/client.go:133:12: undefined: unix.SockaddrLinklayer
```
There are no constants in package **golang.org/x/sys/unix** for darwin_amd64

```
grep AF_PACKET *
grep: linux: Is a directory
syscall_linux.go:// SockaddrLinklayer implements the Sockaddr interface for AF_PACKET type sockets.
syscall_linux.go:	sa.raw.Family = AF_PACKET
syscall_linux.go:	case AF_PACKET:
zerrors_linux_386.go:	AF_PACKET                            = 0x11
zerrors_linux_amd64.go:	AF_PACKET                            = 0x11
zerrors_linux_arm.go:	AF_PACKET                            = 0x11
zerrors_linux_arm64.go:	AF_PACKET                            = 0x11
zerrors_linux_mips.go:	AF_PACKET                            = 0x11
zerrors_linux_mips64.go:	AF_PACKET                            = 0x11
zerrors_linux_mips64le.go:	AF_PACKET                            = 0x11
zerrors_linux_mipsle.go:	AF_PACKET                            = 0x11
zerrors_linux_ppc64.go:	AF_PACKET                            = 0x11
zerrors_linux_ppc64le.go:	AF_PACKET                            = 0x11
zerrors_linux_riscv64.go:	AF_PACKET                            = 0x11
zerrors_linux_s390x.go:	AF_PACKET                            = 0x11
zerrors_linux_sparc64.go:	AF_PACKET                        = 0x11
zerrors_solaris_amd64.go:	AF_PACKET                     = 0x20
```

This is a commit probably breaks compilation:  
https://github.com/insomniacslk/dhcp/commit/4eedccfd12105dbcde88298e87bbbcf41aa7c1df

In BSD system AF_PACKET should be replaced by AF_LINK

BTW, why not to move client for separate package? I just removed file client.go from repo in order to compile server code.

insomniacslk:
cc @get9

insomniacslk:
[...]

PR #316, opened 2019-09-17
DHCPv6 Server multicast handling improvements

Natolumin:
#### Summary

* Patch 1 makes listening on an explicit multicast address work, and removes the useless joining on multicast groups when listening on an unicast address. It adds the second RFC-defined multicast group to the default groups joined when listening on the wildcard address
* Patch 2 moves the definitions for dhcpv4.BindToInterface to the interfaces package, which seems more appropriate as they are not ipv4 specific.
From the code comments in [bsd/netinet/ip_output.c#L2715](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/netinet/ip_output.c#L2715) `IP_BOUND_IF` for darwin doesn't do anything on the *receive* path, only send.`IP_RECVIF` however is also supported on darwin (I assume inherited from freebsd) and should do the same thing
 **Note**: I do not have access to either darwin or *bsd machines to test this change, but at least it does build. Any help / pointers for testing on those (esp. darwin, which I can't easily start in a cloud somewhere) would be appreciated
* Patch 3 makes an equivalent of NewIPv4UDPConn for ipv6, using bindtodevice and v6only, to give greater control over listen parameters, and allow binding to a specific multicast address instead of wildcard

#### Testing
Since we can't really use default ports or require privileges in unittests, I tested the additions manually for the handling of bindtodevice.

This is a build of coredhcp with these patches, listening on wildcard addresses:
```
$ sudo ip netns exec dhcp_servers ss -lnpue
State                           Recv-Q                          Send-Q                                                    Local Address:Port                                                     Peer Address:Port
UNCONN                          0                               0                                                         0.0.0.0%srv_u:67                                                            0.0.0.0:*                              users:(("coredhcp",pid=126583,fd=6)) ino:837636 sk:3 <->
UNCONN                          0                               0                                                            [::]%srv_u:547                                                              [::]:*                              users:(("coredhcp",pid=126583,fd=5)) ino:837635 sk:4 v6only:1 <->
```
An strace log shows it joined both multicast groups (and I verified a relayed packet to `ff05::1:3` was handled): 
```
setsockopt(5, SOL_IPV6, MCAST_JOIN_GROUP, {gr_interface=if_nametoindex("srv_u"), gr_group={sa_family=AF_INET6, sin6_port=htons(0), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "ff02::1:2", &sin6_addr), sin6_scope_id=0}}, 136) = 0
setsockopt(5, SOL_IPV6, MCAST_JOIN_GROUP, {gr_interface=if_nametoindex("srv_u"), gr_group={sa_family=AF_INET6, sin6_port=htons(0), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "ff05::1:3", &sin6_addr), sin6_scope_id=0}}, 136) = 0
```

Similarly, I verified that asking explicitly to bind on `ff05::1:3`, interface srv_u, results in a socket listening on `[ff05::1:3]%srv_u:547`, and the mcast group is joined (but not the other one): 
```
132015 setsockopt(5, SOL_IPV6, MCAST_JOIN_GROUP, {gr_interface=if_nametoindex("srv_u"), gr_group={sa_family=AF_INET6, sin6_port=htons(0), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "ff05::1:3", &sin6_addr), sin6_scope_id=0}}, 136) = 0
```


pmazzini:
I don't think darwin is currenly working: #229 and #230.

Natolumin:
OK, I won't worry too much about darwin then. 
Let's not merge this yet though, I want to finish the coredhcp parts that go with it first in case it reveals API issues with the updated behaviour

Natolumin:
v2 changed the last commit, correcting the documentation and handling the case `addr == nil` correctly

pmazzini:
> Let's not merge this yet though, I want to finish the coredhcp parts that go with it first in case it reveals API issues with the updated behaviour

Are you still testing it?

Natolumin:
I'm pretty sure it's still broken for darwin, but no worse than it was before, so I think it's good to go in the current state

PR #317, opened 2019-09-18
Allow IsMessageType to accept more than one message type

rminnich:
For dhcp6 rapid solicit there is more than one possible message
type in a packet. Allow IsMessageType to match more than one
type of packet, by making at least one required argument
and a variadic for the rest.

### DIFF
--- a/dhcpv6/nclient6/client.go
+++ b/dhcpv6/nclient6/client.go
@@ -234,12 +234,21 @@ func WithBroadcastAddr(n *net.UDPAddr) ClientOpt {
 // Matcher matches DHCP packets.
 type Matcher func(*dhcpv6.Message) bool
 
-// IsMessageType returns a matcher that checks for the message type.
+// IsMessageType returns a matcher that checks for one of several message
+// types.  It is most efficient to put the most likely type first.
 //
 // If t is MessageTypeNone, all packets are matched.
-func IsMessageType(t dhcpv6.MessageType) Matcher {
+func IsMessageType(t dhcpv6.MessageType, tt ...dhcpv6.MessageType) Matcher {
 	return func(p *dhcpv6.Message) bool {
-		return p.MessageType == t || t == dhcpv6.MessageTypeNone
+		if p.MessageType == t || t == dhcpv6.MessageTypeNone {
+			return true
+		}
+		for _, mt := range tt {
+			if p.MessageType == mt {
+				return true
+			}
+		}
+		return false
 	}
 }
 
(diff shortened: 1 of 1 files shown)


insomniacslk:
Will merge once the tests finish

insomniacslk:
@rminnich there is a build failure in the client code, I'm going to sleep now but once fixed @hugelgupf could merge this

dhcpv6/nclient6/client.go
```diff
@@ -234,12 +234,21 @@ func WithBroadcastAddr(n *net.UDPAddr) ClientOpt {
 // Matcher matches DHCP packets.
 type Matcher func(*dhcpv6.Message) bool
 
-// IsMessageType returns a matcher that checks for the message type.
+// IsMessageType returns a matcher that checks for one of several message
+// types.  It is most efficient to put the most likely type first.
 //
 // If t is MessageTypeNone, all packets are matched.
-func IsMessageType(t dhcpv6.MessageType) Matcher {
+func IsMessageType(t dhcpv6.MessageType, tt ...dhcpv6.MessageType) Matcher {
```
hugelgupf: Why not just make it a vararg, instead of a separate first arg + vararg?

dhcpv6/nclient6/client.go
```diff
@@ -250,7 +259,7 @@ func (c *Client) RapidSolicit(ctx context.Context, modifiers ...dhcpv6.Modifier)
 	if err != nil {
 		return nil, err
 	}
-	msg, err := c.SendAndRead(ctx, c.serverAddr, solicit, IsMessageType(dhcpv6.MessageTypeReply))
+	msg, err := c.SendAndRead(ctx, c.serverAddr, solicit, IsMessageType(dhcpv6.MessageTypeReply, dhcpv6.MessageTypeAdvertise))
```
hugelgupf: I don't think this is valid.

If the response is a Reply, we can just return like we already do.

But if the response is an Advertise, then we must send a Request next: https://github.com/insomniacslk/dhcp/blob/834a63bade8aa8927af8b4c9842f0ad0345289d4/dhcpv6/nclient6/client.go#L275

dhcpv6/nclient6/client.go
insomniacslk: This syntax enforces the presence of at least one argument at compile time, without resorting to error handing which would conplicate that API

dhcpv6/nclient6/client.go
insomniacslk: The RFC is not specific about the behaviour in case of Advertise instead of Offer/Ack. If a server ignores the rapidcommit flag and treats it as a 4-way transaction, it's unspecified what the client should do. In this case I would just ignore the reply here, and let the logic around `RapidSolicit` to choose whether to retry with a 4-way transaction. What are your thoughts?

https://tools.ietf.org/html/rfc4039#section-3.1

hugelgupf: My thought is that's valid, but takes twice as long as you wait for RapidSolicit to time out.

insomniacslk: another option might be to let the function fall back to "relaxed" mode if the server doesn't respond. The reason I call it relaxed is that the answer is not valid (the server should not respond with advertise to a rapid-solicit, and should drop it if not able to handle that), but we can break the rules and make it work without waiting for timeout. But I'd like that the default stays "don't break the rule" and provide a flag to do otherwise. Thoughts?

PR #319, opened 2019-09-23
nclient6: copy & paste log infra to v6

hugelgupf:


insomniacslk:
LGTM! This would be nice to have across the other packages too

PR #322, opened 2019-09-25
server6: allow nil IP in addition to unspecified

hugelgupf:
This way, you can use `&net.UDPAddr{Port: 547}` without having to actually put an IP down


insomniacslk:
LGTM. I've re-run the flaky test, but once green I'll merge this

PR #323, opened 2019-10-01
server{4,6}: Return UDPConn from NewIPv*UDPConn

Natolumin:
The concrete type under the interface is known here since we create the connection in the same function. Since *net.UDPConn implements net.PacketConn anyway, returning the concrete type here is more powerful and less risky than having downstream users cast the value themselves

There should be no code change for downstream users as the concrete type implements the interface, with the exception of explicit casts (`udpc := conn.(*net.UDPConn)`), which can simply be removed

(I could not find any users of these functions outside the library on github anyway so this looks safe)

### DIFF
--- a/dhcpv4/server4/conn.go
+++ b/dhcpv4/server4/conn.go
@@ -1,6 +1,7 @@
 package server4
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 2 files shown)


Natolumin:
I think the test failure is transient but I don't understand why: https://travis-ci.org/insomniacslk/dhcp/builds/592047213 (same code, but also tests w/ go 1.13) passed in all versions

PR #325, opened 2019-10-07
dhcpv6: Basic fuzzing support and fixes for discovered bugs

Natolumin:
Since this library parses sometimes untrusted data, I thought it would be relevant to add fuzzing support to make sure we act reasonably whatever the data.

This adds a fairly simple fuzzing entrypoint for dhcpv6 (I have the same in the pipe for dhcpv4, but it needs a bit more work), which parses a message, then writes it back out, and compares them.
This is fairly limited obviously: 
 * Maybe not all messages don't need to be byte-for-byte identical after being deserialized-serialized, causing some false-positives. This hasn't caused issues yet though
 * It would only catch bugs in messages that come from deserializing, not from messages being crafted using the library

See the page for go-fuzz for usage instructions: https://github.com/dvyukov/go-fuzz; but basically:
```
go get github.com/dvyukov/go-fuzz/go-fuzz{,-build}
cd dhcpv6 && go-fuzz-build
go-fuzz -bin dhcpv6-fuzz.zip /path/to/working/directory
```

It has already caught a couple bugs when run for some time on my desktop without even a source corpus. With an initial corpus built from real traffic and some more processing power I think it could be useful.
If you do not want this support upstream, I can remove this first commit and only submit the bugfixes instead: 

There are 2 bugfixes from missing error checks, causing invalid (truncated) messages to be accepted (empty messages, and 1-byte messages).
There is one bugfix for the iana.HWType encoding: it is declared as 8bits, but both the field in the packet, and the [IANA table](https://www.iana.org/assignments/arp-parameters/arp-parameters.xml#arp-parameters-2) are 16bits; and there exists one value >255 in that table. This changes a type definition, so downstream users may need to apply some changes if they were relying on iana.HWType being 8bits

### DIFF
--- a/iana/hwtypes.go
+++ b/iana/hwtypes.go
@@ -1,7 +1,7 @@
 package iana
 
 // HWType is a hardware type as per RFC 2132 and defined by the IANA.
-type HWType uint8
+type HWType uint16
 
 // See IANA for values.
 const (
--- a/dhcpv6/dhcpv6.go
+++ b/dhcpv6/dhcpv6.go
@@ -46,6 +46,9 @@ func MessageFromBytes(data []byte) (*Message, error) {
 		MessageType: messageType,
 	}
 	buf.ReadBytes(d.TransactionID[:])
+	if buf.Error() != nil {
+		return nil, fmt.Errorf("Error parsing DHCPv6 header: %v", buf.Error())
+	}
 	if err := d.Options.FromBytes(buf.Data()); err != nil {
 		return nil, err
 	}
@@ -68,6 +71,9 @@ func RelayMessageFromBytes(data []byte) (*RelayMessage, error) {
 	d.LinkAddr = net.IP(buf.CopyN(net.IPv6len))
 	d.PeerAddr = net.IP(buf.CopyN(net.IPv6len))
 
+	if buf.Error() != nil {
+		return nil, fmt.Errorf("Error parsing RelayMessage header: %v", buf.Error())
+	}
 	// TODO: fail if no OptRelayMessage is present.
 	if err := d.Options.FromBytes(buf.Data()); err != nil {
 		return nil, err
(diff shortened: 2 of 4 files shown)


pmazzini:
> There is one bugfix for the iana.HWType encoding: it is declared as 8bits, but both the field in the packet, and the IANA table are 16bits

The dhcpv4 field in the packet is 8bits, isn't it?

iana/hwtypes.go
```diff
@@ -1,7 +1,7 @@
 package iana
 
 // HWType is a hardware type as per RFC 2132 and defined by the IANA.
-type HWType uint8
+type HWType uint16
```
hugelgupf: ? I don't think this is right? And it might break ipv4?

Natolumin: This is definitely right for dhcpv6 (and ARP, for that matter, which also has 16bits for this field), see this table (referenced in RFC8415 in https://tools.ietf.org/html/rfc8415#section-11.2): https://www.iana.org/assignments/arp-parameters/arp-parameters.xml#arp-parameters-2, there are assigned hardware types for values 256 and 257

However yes it seems it would break bootp/v4, we need to add an explicit conversion to uint8 there I guess

Natolumin:
v2: added regression tests for each fix

Natolumin:
I think we shouldn't. The DHCPv4 struct is a logical representation of what there is in a packet, it doesn't have to match the wire format exactly. For example ServerHostName and BootFileName are strings instead of fixed-size arrays of bytes, ClientHWAddr is `net.HardwareAddr` (which has arbitrary size) instead of fixed-size array of bytes.
I think this is the same idea as those. Additionally keeping it as a separate type lets us add methods on the type later, eg to map a HWType to a hardware address length or explicitly check for overflow

Issue #326, opened 2019-10-15
`server4.NewServer()` does not respect `addr.IP`


lewisthompson:
In https://github.com/insomniacslk/dhcp/blob/3b29cbd2a47fb7ea9a028ca3bc150f855ae2c5ec/dhcpv4/server4/conn.go#L39-L42:

```go
	// Bind to the port.
	if err := unix.Bind(fd, &unix.SockaddrInet4{Port: port}); err != nil {
		return nil, fmt.Errorf("cannot bind to port %d: %v", port, err)
	}
```

The `Bind` is made only against the specified `port` meaning that the `net.UDPAddr` passed to `NewServer` is partially ignored.

I am attempting to build a DHCP server that supports multiple ranges of IPs.  The only way I understand that I can make this work is to run multiple instances of `server4` where each is bound to a specific interface.  For this to work I need to be able to bind to the configured host IP of each of those interfaces (to avoid `address already in use`).

e.g. if I have `if0 / 192.168.1.1/24` and `if1 / 192.168.2.1/24` I want to launch one `server4` listening on `192.168.1.1:67` on `if0` and another on `192.168.2.1:67` on `if1`.

Am I missing something obvious here or would having `NewIPv4UDPConn` respect the `IP` allow things to work for me?

Natolumin:
> Am I missing something obvious here or would having NewIPv4UDPConn respect the IP allow things to work for me?

There is at least one issue with not listening on the wildcard address: you won't receive broadcasts. Without relays, all the DHCPDISCOVER messages should be broadcast, so basically you won't receive anything.

> I am attempting to build a DHCP server that supports multiple ranges of IPs. The only way I understand that I can make this work is to run multiple instances of server4 where each is bound to a specific interface. For this to work I need to be able to bind to the configured host IP of each of those interfaces (to avoid address already in use).

I do not think that is necessary. The BindToInterface call should ensure you can listen on the wildcard address on multiple sockets on multiple interfaces at the same time, as long as each of the sockets binds to an interface, so it should work to bind on wildcard on each of these interfaces
Here's an excerpt of the sockets on my system which shows that working:
```
$ sudo ip netns exec dhcp_relay ss -lnpu
Local Address:Port      Peer Address:Port
0.0.0.0%cli_u:67              0.0.0.0:*            users:(("coredhcp",pid=20848,fd=9))
0.0.0.0%srv_d:67              0.0.0.0:*            users:(("coredhcp",pid=20848,fd=8))
```
[...]

PR #327, opened 2019-10-15
server4: Allow listening on a non-wildcard IP address on v4

Natolumin:
It was mentioned in #326 that NewIPv4UDPConn doesn't support listening on a specific address, only on the wildcard address. This extends it to allow listening on an address, and at the same time homogenizes the function signature with the NewIPv6UDPConn server6 equivalent. It modifies NewServer() to pass the full address given to it instead of just the port as well

Note that listening on a non-wildcard interface is seldom useful as the socket won't receive broadcasts, so it is useless in a direct-attached server. It can be useful in a server only used behind relays

This breaks API compatibility for NewIPv4UDPConn, which as far as I know/on github nobody uses (yet)

### DIFF
--- a/dhcpv4/server4/conn.go
+++ b/dhcpv4/server4/conn.go
@@ -14,7 +14,7 @@ import (
 // given based on a IPv4 DGRAM socket. The UDP connection allows broadcasting.
 //
 // The interface must already be configured.
-func NewIPv4UDPConn(iface string, port int) (*net.UDPConn, error) {
+func NewIPv4UDPConn(iface string, addr *net.UDPAddr) (*net.UDPConn, error) {
 	fd, err := unix.Socket(unix.AF_INET, unix.SOCK_DGRAM, unix.IPPROTO_UDP)
 	if err != nil {
 		return nil, fmt.Errorf("cannot get a UDP socket: %v", err)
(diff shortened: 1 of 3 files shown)


dhcpv4/server4/conn.go
```diff
@@ -38,8 +38,10 @@ func NewIPv4UDPConn(iface string, port int) (*net.UDPConn, error) {
 		}
 	}
 	// Bind to the port.
-	if err := unix.Bind(fd, &unix.SockaddrInet4{Port: port}); err != nil {
-		return nil, fmt.Errorf("cannot bind to port %d: %v", port, err)
+	saddr := unix.SockaddrInet4{Port: addr.Port}
+	copy(saddr.Addr[:], addr.IP.To4())
```
hugelgupf: If addr.IP != nil?

dhcpv4/server4/conn.go
```diff
@@ -14,7 +14,7 @@ import (
 // given based on a IPv4 DGRAM socket. The UDP connection allows broadcasting.
 //
 // The interface must already be configured.
-func NewIPv4UDPConn(iface string, port int) (*net.UDPConn, error) {
+func NewIPv4UDPConn(iface string, addr *net.UDPAddr) (*net.UDPConn, error) {
```
hugelgupf: Make addr net.UDPAddr so it can't be nil, or check it for nil?

dhcpv4/server4/conn.go
Natolumin: That's fine, addr.IP.To4() will be nil in that case, and nothing will be copied; which is what we want anyway: https://play.golang.org/p/0RRxwSbL1WL

dhcpv4/server4/conn.go
Natolumin: Good point. NewServer() also doesn't check for nil and takes a pointer, I'll add in both

Natolumin:
v2: enforce address family is v4 and handle addr==nil gracefully

dhcpv4/server4/conn.go
Natolumin: In the end I didn't add a check in NewServer, since it now just passes the value through it won't crash with addr=nil anymore (but would have before this PR). NewIPv4UDPConn now checks and will do the right thing

PR #329, opened 2019-11-03
dhcpv6: 4RD option support

Natolumin:
This PR includes option parsing for 4RD options defined in RFC7600

The first commit is a preparation commit which adds the necessary Option* constants; I took the opportunity to update the table up to the current version of the [IANA table](https://www.iana.org/assignments/dhcpv6-parameters/dhcpv6-parameters.xhtml#dhcpv6-parameters-2)

The second commit includes struct definitions, parsing and serializing for the 3 option types defined in [RFC7600§4.9](https://tools.ietf.org/html/rfc7600#section-4.9)

Additional comments on specific lines of the code inline

### DIFF
--- a/dhcpv6/option_4rd.go
+++ b/dhcpv6/option_4rd.go
@@ -1,0 +1,178 @@
+package dhcpv6
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/u-root/u-root/pkg/uio"
+)
+
+// Opt4RD represents a 4RD option. It is only a container for 4RD_*_RULE options
+type Opt4RD Options
+
+// Code returns the Option Code for this option
+func (op *Opt4RD) Code() OptionCode {
+	return Option4RD
+}
+
+// ToBytes serializes this option
+func (op *Opt4RD) ToBytes() []byte {
+	return (*Options)(op).ToBytes()
+}
+
+// String returns a human-readable representation of the option
+func (op *Opt4RD) String() string {
+	return fmt.Sprintf("Opt4RD{%v}", (*Options)(op))
+}
+
+// ParseOpt4RD builds an Opt4RD structure from a sequence of bytes.
+// The input data does not include option code and length bytes
+func ParseOpt4RD(data []byte) (*Opt4RD, error) {
+	var opt Options
+	err := opt.FromBytes(data)
+	return (*Opt4RD)(&opt), err
+}
+
+// Opt4RDMapRule represents a 4RD Mapping Rule option
+// The option is described in https://tools.ietf.org/html/rfc7600#section-4.9
+// The 4RD mapping rules are described in https://tools.ietf.org/html/rfc7600#section-4.2
+type Opt4RDMapRule struct {
+	// Prefix4 is the IPv4 prefix mapped by this rule
+	Prefix4 net.IPNet
+	// Prefix6 is the IPv6 prefix mapped by this rule
+	Prefix6 net.IPNet
+	// EABitsLength is the number of bits of an address used in constructing the mapped address
+	EABitsLength uint8
+	// WKPAuthorized determines if well-known ports are assigned to addresses in an A+P mapping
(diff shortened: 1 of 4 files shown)


dhcpv6/option_4rd.go
```diff
@@ -0,0 +1,178 @@
...
+	buf.Write8(uint8(p4Len))
+	buf.Write8(uint8(p6Len))
+	buf.Write8(op.EABitsLength)
+	if op.WKPAuthorized {
+		buf.Write8(opt4RDWKPAuthorizedMask)
+	} else {
+		buf.Write8(0)
+	}
+	if op.Prefix4.IP.To4() == nil {
+		// The API prevents us from returning an error here
```
Natolumin: I hate this, but I can't imagine another solution without changing the whole API and returning `([]byte, error)` for ToBytes.

dhcpv6/types.go
```diff
@@ -154,79 +154,214 @@ const (
...
+	OptionBootfileParam:                         "OPT_BOOTFILE_PARAM",
+	OptionClientArchType:                        "OPTION_CLIENT_ARCH_TYPE",
+	OptionNII:                                   "OPTION_NII",
+	OptionGeolocation:                           "OPTION_GEOLOCATION",
+	OptionAFTRName:                              "OPTION_AFTR_NAME",
+	OptionERPLocalDomainName:                    "OPTION_ERP_LOCAL_DOMAIN_NAME",
+	OptionRSOO:                                  "OPTION_RSOO",
+	OptionPDExclude:                             "OPTION_PD_EXCLUDE",
+	OptionVirtualSubnetSelection:                "Virtual Subnet Selection",
+	OptionMIPv6IdentifiedHomeNetworkInformation: "MIPv6 Identified Home Network Information",
```
Natolumin: No change in values here, just gofmt moving everything around to be *less* aligned for some reason

dhcpv6/option_4rd.go
```diff
@@ -0,0 +1,178 @@
...
+	opt.Prefix6.IP = net.IP(buf.CopyN(net.IPv6len))
+	return &opt, buf.FinError()
+}
+
+// Opt4RDNonMapRule represents 4RD parameters other than mapping rules
+type Opt4RDNonMapRule struct {
+	// HubAndSpoke is whether the network topology is hub-and-spoke or meshed
+	HubAndSpoke bool
+	// TrafficClass is an optional 8-bit tunnel traffic class identifier
+	TrafficClass *uint8
```
Natolumin: This isn't great, but there is no Option<> type in go obviously. Does anyone have a better idea for this ?

dhcpv6/option_4rd.go
hugelgupf: I see this a little differently.

I think nil is as good as net.IP{0, 0, 0, 0}. I don't see you checking for that?

And then there's probably a lot of other invalid values for this, like net.IP{255, 255, 255, 255}.

Are we able to enumerate the exact list of invalid things? Or does it depend on your specific network?

I don't think this library should concern itself with validation of the values. My take is - the user is responsible. Especially because in a bunch of other places in this library, nil = net.IP{0, 0, ...}, too.

Natolumin: I wasn't worried about invalid packets, I definitely agree (at least this API of) the library should allow you to craft packets that don't respect the RFC requirements. Also agree that at least treating nil = net.IP{0...} is fine and common

My worry was about unrepresentable packets. Say you put an (non-ipv4-mapped) ipv6 address in the Prefix4 field, or a 5-byte slice. Those aren't valid net.IP packets and you can't represent them in the wire format, so I think they still should raise some kind of serialization error, instead of silently serializing to zeros.

dhcpv6/option_4rd.go
pmazzini: I don't.

pmazzini:
Thanks for adding the additional options and IANA codes.

PR #331, opened 2019-11-19
Add support for copying relay options

kworm83:
Here is an initial PR for adding support to copy relay options if they exist in a request to dhcpv4.NewReplyFromRequest.  RFC3046 section 2.2 specifies "DHCP servers claiming to support the Relay Agent Information option SHALL echo the entire contents of the Relay Agent Information option
 in all replies." so I think it is safe to add to the NewReplyFromRequest function.

Perhaps the new function should be name CopyRelayOptions instead of WithRelayOptions?

### DIFF
--- a/dhcpv4/modifiers.go
+++ b/dhcpv4/modifiers.go
@@ -43,14 +43,12 @@ func WithGatewayIP(ip net.IP) Modifier {
 	}
 }
 
-//WithRelayOptions copies the relay options from the request to the reply
-func WithRelayOptions(request *DHCPv4) Modifier {
+// WithRelayAgentInfo copies the relay options from the request to the reply.
+func WithRelayAgentInfo(request *DHCPv4) Modifier {
 	return func(d *DHCPv4) {
 		// If request has Relay Agent Info copy it to the reply
-		if request.Options.Has(OptionRelayAgentInformation) {
-			relayopt := request.Options.Get(OptionRelayAgentInformation)
-			opt := OptGeneric(OptionRelayAgentInformation, relayopt)
-			d.Options.Update(opt)
+		if relayOpt := request.RelayAgentInfo(); relayOpt != nil {
+			d.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionRelayAgentInformation, Value: relayOpt})
 		}
 	}
 }
--- a/dhcpv4/modifiers.go
+++ b/dhcpv4/modifiers.go
@@ -43,6 +43,18 @@ func WithGatewayIP(ip net.IP) Modifier {
 	}
 }
 
+//WithRelayOptions copies the relay options from the request to the reply
+func WithRelayOptions(request *DHCPv4) Modifier {
+	return func(d *DHCPv4) {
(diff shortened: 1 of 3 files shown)


dhcpv4/modifiers.go
```diff
@@ -43,6 +43,18 @@ func WithGatewayIP(ip net.IP) Modifier {
 	}
 }
 
+//WithRelayOptions copies the relay options from the request to the reply
```
hugelgupf: super nit: Add a space at the beginning and a period at the end `// WithRelay...`

dhcpv4/modifiers.go
```diff
@@ -43,6 +43,18 @@ func WithGatewayIP(ip net.IP) Modifier {
...
+func WithRelayOptions(request *DHCPv4) Modifier {
```
hugelgupf: `WithRelayAgent`? There's other relay options, right?

dhcpv4/modifiers.go
```diff
@@ -43,6 +43,18 @@ func WithGatewayIP(ip net.IP) Modifier {
...
+	return func(d *DHCPv4) {
+		// If request has Relay Agent Info copy it to the reply
+		if request.Options.Has(OptionRelayAgentInformation) {
```
pmazzini: I think this can be simplified:

```
	if relayOpt := request.RelayAgentInfo(); relayOpt != nil {
		d.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionRelayAgentInformation, Value: relayOpt})
	}
```

dhcpv4/modifiers.go
kworm83: WithRelayAgentInfo might be best?

dhcpv4/modifiers.go
kworm83: Looks good to me.  What do you think about the func name WithRelayAgentInfo?

dhcpv4/modifiers.go
```diff
@@ -43,6 +43,16 @@ func WithGatewayIP(ip net.IP) Modifier {
 	}
 }
 
+// WithRelayAgentInfo copies the relay options from the request to the reply.
+func WithRelayAgentInfo(request *DHCPv4) Modifier {
+	return func(d *DHCPv4) {
+		// If request has Relay Agent Info copy it to the reply
+		if relayOpt := request.RelayAgentInfo(); relayOpt != nil {
+			d.UpdateOption(dhcpv4.Option{Code: dhcpv4.OptionRelayAgentInformation, Value: relayOpt})
```
pmazzini: I know I added it it my previous comment but the dhcpv4 namespace is not needed.

kworm83: Fixed that and the function name in dhcpv4.go - have something funny going on with my test build environment bc it should not have built cleanly on my system.

pmazzini:
lgtm. Can you add a test?

pmazzini:
> Can you add a test?

A unit test for the new modifier in `modifiers_test.go`.

kworm83:
Yes I will try to add a test.

PR #332, opened 2019-11-20
Proposed removal of some logging from Serve() function

kworm83:
I propose that the stdout logging at startup, before error return, and on each connection should be removed from the Serve() function.  Since it is a library logging does not seem appropriate in those areas.  I left the logging in the two areas that continue instead of error return for discussion on whether it would be appropriate to remove those and just silently drop packets/connections with errors.


pmazzini:
Go’s log package doesn’t have leveled logs which would help disabling it. It makes sense not to log in the library. I don't have a strong opinion with regard to this change. @insomniacslk?

hugelgupf:
we could do something like in nclient4?

pmazzini:
> we could do something like in nclient4?

Yes, that is probably the best approach.

pmazzini:
Closing in favor of #371

PR #333, opened 2019-11-20
Respect RFC 6842

hugelgupf:


dhcpv4/modifiers.go
```diff
@@ -43,16 +43,20 @@ func WithGatewayIP(ip net.IP) Modifier {
...
-		if relayOpt := request.RelayAgentInfo(); relayOpt != nil {
-			d.UpdateOption(Option{Code: OptionRelayAgentInformation, Value: relayOpt})
+		if val := request.Options.Get(opt); val != nil {
+			d.UpdateOption(OptGeneric(opt, val))
 		}
 	}
 }
 
+// WithRelayAgentInfo copies the relay options from the request to the reply.
+func WithRelayAgentInfo(request *DHCPv4) Modifier {
```
pmazzini: I would remove this and keep the generic one only :)

dhcpv4/option_ip_address_lease_time.go
```diff
@@ -1,11 +1,15 @@
...
 
 import (
+	"math"
 	"time"
 
 	"github.com/u-root/u-root/pkg/uio"
 )
 
+// MaxLeaseTime is the maximum lease time that can be encoded.
+var MaxLeaseTime = math.MaxUint32 * time.Second
```
pmazzini: what is this for?

hugelgupf: Convenience for when I want to give a long lease

Issue #334, opened 2019-11-22
nclient4 API change proposal: (*Client).Request(...) (discover, offer, request, ack *dhcpv4.DHCPv4, err error)


xaionaro:
Hello!

The old v4 client (`client4`) has method [Exchange](https://github.com/insomniacslk/dhcp/blob/ec0e0154d15c429b3dcc56af60a2d7b62eb3d5e7/dhcpv4/client4/client.go) which returns 4 messages: `Discover`, `Offer`, `Request` and `Ack`.

While the closest analog in the new v4 client (`nclient4`), method [Request](https://github.com/insomniacslk/dhcp/blob/ec0e0154d15c429b3dcc56af60a2d7b62eb3d5e7/dhcpv4/nclient4/client.go#L331) returns only `Offer` and `Ack`.

And we want to migrate our DHCP dumper from `client4` to `nclient4` and we need all 4 messages. So I have 3 choices:

* To re-implement `Exchange` based on `nclient4` inside our dumper.
* To implement an analog of `Exchange` inside `nclient4`.
* To fix `Request` in `nclient4` to return all 4 messages.

I'd prefer to follow the last variant:

* This is not our specific that it's required to dump all 4 messages. It seems to be quite common task among other companies. So this function should be in the upstream (in `nclient4`).
* Writing another function that does almost the same as `Request` is a duplication of code. We could just fix `Request` and ignore `Discover` and `Request` when they are not needed.

@insomniacslk: What do you think? Should I prepare a PR with the fix for `Request`?

UPD: It will also require to modify `DiscoverOffer` to return both: `Discover` and `Offer`.

insomniacslk:
CC @hugelgupf regarding the changes to nclient4

insomniacslk:
we are also staying on the old client4 for similar reasons

pmazzini:
If this is only for logging I think that the logger interface can be used instead since it prints each packet.

https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/nclient4/client.go#L424

xaionaro:
> If this is only for logging I think that the logger interface can be used instead since it prints each packet.

I wouldn't trust the logging system as to a persistent API.

xaionaro:
@pmazzini : Your way with logger has two advantages:

* It prints messages in real time (instead of printing them all in the ending).
* It prints re-sends of a message.

So I tried to implement both ways here:

https://github.com/insomniacslk/dhcp/pull/335

[...]

PR #335, opened 2019-11-26
Simplified porting from client4 to nclient4

xaionaro:
Related: https://github.com/insomniacslk/dhcp/issues/334

The PR is supposed to reduce costs of migrating from client4 to nclient4.

Changes in [nclient4](https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/nclient4/client.go):

* Published `Logger`. So now somebody can use option `WithLogger` and define own logger or/and to use a builtin logger. And added comment `print _all_ DHCP messages` to guarantee (to an external user) and to remind (to an internal developer) that PrintMessage always prints all messages. So in task https://github.com/insomniacslk/dhcp/issues/334 it's possible to use this loggers instead of changing `Request`, now.
* Fixed the problem with unicast frames. Previously `nclient4` was always sending broadcast frames. But our utility (related to https://github.com/insomniacslk/dhcp/issues/334) requires also support of unicast frames (which are supported in the old package [client4](https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/client4/client.go)).
* Removed function `isErrClosing(...)`. It's not universal through different implementations of [net.PacketConn](https://golang.org/pkg/net/#PacketConn).
* ~~`DiscoverOffer` and `Request` now returns all messages (not only replies, but also requests). So `Request` is now more logically compatible with the old method [`Exchange`](https://github.com/insomniacslk/dhcp/blob/793c0f67c0a9a2b178bfe98223fd2bfc92a115f3/dhcpv4/client4/client.go#L183).~~ (it's requested to move this change to a separate PR)
* Moved errors to `var`-s and `type`-s. Used `%w` instead of `%v` while wrapping an error with `fmt.Errorf`. Wrapped more received errors with `fmt.Errorf`.

### DIFF
--- a/.travis/tests.sh
+++ b/.travis/tests.sh
@@ -10,14 +10,23 @@ echo "" > coverage.txt
 # tests.
 ip a
 
+GO_TEST_OPTS=()
+if [[ "$TRAVIS_GO_VERSION" =~ ^1.(9|10|11|12)$ ]]
+then
+    # We use fmt.Errorf with verb "%w" which appeared only in Go1.13.
+    # So the code compiles and works with Go1.12, but error descriptions
+    # looks uglier and it does not pass "vet" tests on Go<1.13.
+    GO_TEST_OPTS+='-vet=off'
+fi
+
 for d in $(go list ./... | grep -v vendor); do
-    go test -race -coverprofile=profile.out -covermode=atomic $d
+    go test -race -coverprofile=profile.out -covermode=atomic ${GO_TEST_OPTS[@]} $d
     if [ -f profile.out ]; then
         cat profile.out >> coverage.txt
         rm profile.out
     fi
     # integration tests
-    go test -c -cover -tags=integration -race -covermode=atomic $d
+    go test -c -cover -tags=integration -race -covermode=atomic ${GO_TEST_OPTS[@]} $d
     testbin="./$(basename $d).test"
     # only run it if it was built - i.e. if there are integ tests
     test -x "${testbin}" && sudo "./${testbin}" -test.coverprofile=profile.out
--- a/dhcpv4/client4/client.go
+++ b/dhcpv4/client4/client.go
@@ -1,3 +1,4 @@
+// Package client4 is deprecated. Use "nclient4" instead.
 package client4
 
 import (
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -18,7 +18,6 @@ import (
 	"log"
 	"net"
 	"os"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -27,10 +26,16 @@ import (
 )
 
 const (
-	defaultTimeout   = 5 * time.Second
-	defaultRetries   = 3
 	defaultBufferCap = 5
-	maxMessageSize   = 1500
+
+	// DefaultTimeout is the default value for read-timeout if option WithTimeout is not set
+	DefaultTimeout = 5 * time.Second
+
+	// DefaultRetries is amount of retries will be done if no answer was received within read-timeout amount of time
+	DefaultRetries = 3
+
+	// MaxMessageSize is the value to be used for DHCP option "MaxMessageSize".
+	MaxMessageSize = 1500
 
 	// ClientPort is the port that DHCP clients listen on.
 	ClientPort = 68
@@ -51,6 +56,12 @@ var (
 var (
 	// ErrNoResponse is returned when no response packet is received.
 	ErrNoResponse = errors.New("no matching response packet received")
+
+	// ErrNoConn is returned when NewWithConn is called with nil-value as conn.
+	ErrNoConn = errors.New("conn is nil")
+
+	// ErrNoIfaceHWAddr is returned when NewWithConn is called with nil-value as ifaceHWAddr
+	ErrNoIfaceHWAddr = errors.New("ifaceHWAddr is nil")
 )
 
 // pendingCh is a channel associated with a pending TransactionID.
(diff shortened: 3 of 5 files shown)


dhcpv4/client4/client.go
```diff
@@ -1,3 +1,5 @@
+// This package is deprecated. Use "nclient4" instead.
+
```
hugelgupf: Remove empty line, so this becomes a godoc.

dhcpv4/nclient4/client.go
```diff
@@ -249,54 +303,57 @@ func (c *Client) receiveLoop() {
...
 func WithTimeout(d time.Duration) ClientOpt {
-	return func(c *Client) {
+	return func(c *Client) (err error) {
 		c.timeout = d
+		return
 	}
 }
 
-// WithSummaryLogger logs one-line DHCPv4 message summarys when sent & received.
-func WithSummaryLogger() ClientOpt {
```
hugelgupf: Please keep the WithSummaryLogger and WithDebugLogger for people already using nclient4.

dhcpv4/nclient4/client.go
```diff
@@ -193,21 +249,19 @@ func (c *Client) Close() error {
 	return err
 }
 
-func isErrClosing(err error) bool {
-	// Unfortunately, the epoll-connection-closed error is internal to the
-	// net library.
-	return strings.Contains(err.Error(), "use of closed network connection")
+func (c *Client) isClosed() bool {
+	return atomic.LoadUint32(&c.closed) != 0
```
hugelgupf: closed is never set? Also, why the change?

dhcpv4/nclient4/client.go
```diff
@@ -249,54 +303,57 @@ func (c *Client) receiveLoop() {
...
 
-// WithDebugLogger logs multi-line full DHCPv4 messages when sent & received.
-func WithDebugLogger() ClientOpt {
-	return func(c *Client) {
-		c.logger = debugLogger{
-			Logger: log.New(os.Stderr, "[dhcpv4] ", log.LstdFlags),
-		}
+// WithUnicast forces client to send messages as unicast frames.
+// By default client sends messages as broadcast frames even if server address is defined.
+func WithUnicast() ClientOpt {
```
hugelgupf: Can you allow users to specify a *net.UDPAddr here?

If they don't specify one, choose the default like you did here?

dhcpv4/nclient4/client.go
```diff
@@ -441,7 +515,8 @@ func (c *Client) retryFn(fn func(timeout time.Duration) error) error {
 	timeout := c.timeout
 
 	// Each retry takes the amount of timeout at worst.
-	for i := 0; i < c.retry || c.retry < 0; i++ {
+	for i := 0; i < c.retry || c.retry < 0; i++ { // TODO: why is this called "retry" if this is "tries" ("retries"+1)?
+		c.logger.Printf("read-timeout is %v", timeout)
```
hugelgupf: revert this debug print?

dhcpv4/nclient4/client.go
xaionaro: `c.closed` is set in `c.Close()`. The change because the previous way is:

* Hacky. Search a substring in the error description is an unstrict/risky way to handle errors.
* Wrong. I can implement `net.PacketConn` which returns other error. And this package accepts arbitrary `net.PacketConn`-s via `NewWithConn(...)`.

xaionaro:
Thank you for your time and review :)

I fixed almost everything. Only one thing left for discussion (`isClosed`).

pmazzini:
Please sign the commit:

```
git rebase HEAD~2 --signoff
git push --force-with-lease origin master
```

> DiscoverOffer and Request now returns all messages (not only replies, but also requests). So Request is now more logically compatible with the old method Exchange.

Can this be done on a separate PR?

> Used xerrors.Errorf instead of fmt.Errorf

This will introduce a new dependency to `u-root`. Can't we use something from the standard library instead?

xaionaro:
> Can't we use something from the standard library instead?

Yeah. It appears they already added flag `%w` to `fmt.Errorf`, so I just could change `%v` to `%w`. Will be fixed.

> If the format specifier includes a %w verb with an error operand, the returned error will implement an Unwrap method returning the operand. It is invalid to include more than one %w verb or to supply it with an operand that does not implement the error interface. The %w verb is otherwise a synonym for %v.

xaionaro:
@pmazzini : removed an extra dependency, squashed commits, signed-off :)

xaionaro:
@hugelgupf : I used `%w` for `fmt.Errorf`. So it will print neat error description only on Go>1.13.

For comparison:

An error description on Go1.13
```
could not get local IPv4 from fbwhoami: host doesn't have an IPv4 address
```

An error description on Go1.12:
```
could not get local IPv4 from fbwhoami: &{%!w(string=host doesn't have an IPv4 address)}
```

I hope it's not a problem.

.travis/tests.sh
```diff
@@ -10,14 +10,20 @@ echo "" > coverage.txt
 # tests.
 ip a
 
+GO_TEST_OPTS=()
+if [[ "$TRAVIS_GO_VERSION" =~ ^1.(9|10|11|12)$ ]]
+then
+    GO_TEST_OPTS+='-vet=off'
```
pmazzini: Maybe add a comment that this is because we are using the `%w` verb.

dhcpv4/nclient4/client.go
```diff
@@ -63,35 +74,61 @@ type pendingCh struct {
...
+	// PrintMessage print _all_ DHCP messages
 	PrintMessage(prefix string, message *dhcpv4.DHCPv4)
+
+	// Printf is use to print the rest debugging information
+	Printf(format string, v ...interface{})
 }
 
-type emptyLogger struct{}
+// EmptyLogger prints nothing
+type EmptyLogger struct{}
```
pmazzini: Do we need to make this public?

dhcpv4/nclient4/client.go
```diff
@@ -193,21 +249,19 @@ func (c *Client) Close() error {
...
+	return atomic.LoadUint32(&c.closed) != 0
 }
 
 func (c *Client) receiveLoop() {
 	defer c.wg.Done()
 	for {
 		// TODO: Clients can send a "max packet size" option in their
-		// packets, IIRC. Choose a reasonable size and set it.
-		b := make([]byte, maxMessageSize)
+		//       packets, IIRC. Choose a reasonable size and set it.
```
pmazzini: Is this indentation on purpose?

dhcpv4/nclient4/client.go
xaionaro: I don't see any reason to hide it. Somebody can do now something like:
```
var logger nclient4.Logger
switch loggerType {
case someValue0:
    logger = &nclient4.EmptyLogger{}
case someValue1:
    logger = &nclient4.DebugLogger{Printfer: log.New(...)}
case someValue2:
    logger = &myLogger{}
}
opts = append(opts, nclient4.WithLogger(logger))
```

But I can hide it if you want :)

dhcpv4/nclient4/client.go
```diff
@@ -129,39 +166,58 @@ type Client struct {
...
-	i, err := net.InterfaceByName(iface)
-	if err != nil {
-		return nil, err
-	}
-	pc, err := NewRawUDPConn(iface, ClientPort)
-	if err != nil {
-		return nil, err
-	}
-	return NewWithConn(pc, i.HardwareAddr, opts...)
+	return new(iface, nil, nil, opts...)
```
pmazzini: Just out of curiosity, is there a reason for reorganizing this?

dhcpv4/nclient4/client.go
pmazzini: No, it is fine, if you think it may be useful lets leave it.

dhcpv4/nclient4/client.go
```diff
@@ -63,35 +74,61 @@ type pendingCh struct {
...
 
-func (e emptyLogger) Printf(format string, v ...interface{})             {}
-func (e emptyLogger) PrintMessage(prefix string, message *dhcpv4.DHCPv4) {}
+// PrintMessage is just a dummy function that does nothing
+func (e EmptyLogger) PrintMessage(prefix string, message *dhcpv4.DHCPv4) {}
 
-type shortSummaryLogger struct {
-	*log.Logger
+// Printfer is used for actual output of the logger. For example *log.Logger is a Printfer.
+type Printfer interface {
```
pmazzini: Do we actually need this new interface?

dhcpv4/nclient4/client.go
xaionaro: Yes. The `Conn` should be initialized after processing options to be able to initialize an unicast `Conn` via these options.

dhcpv4/nclient4/client.go
xaionaro: I use the same interface in multiple `struct`-s, so it's just DRY. And it's public, because `DebugLogger` and `ShortSummaryLogger` are public.

dhcpv4/nclient4/client.go
xaionaro: Yeah (my IDE did not recognize the second line as a "TODO"). But I'll revert it.

dhcpv4/nclient4/client.go
pmazzini: Thanks for the clarification!

.travis/tests.sh
xaionaro: Fixed.

dhcpv4/nclient4/client.go
xaionaro: Fixed.

.travis/tests.sh
xaionaro: Fixed.

dhcpv4/nclient4/client.go
pmazzini: Why not use `*log.Logger` inside `DebugLogger` and `ShortSummaryLogger`?

xaionaro: "Require less, provide more"

We don't need exactly `*log.Logger`. We need just some `Printf`-er here (not more). Somebody else could use `logrus` or whatever else instead of `*log.Logger`. If you require only the thing you really need you make this library much more flexible. There's only one real disadvantage I know: handling through interfaces is slower than handling directly through an implementation.

pmazzini:
Lets move forward, we can revisit how the close check is being done.

PR #336, opened 2019-11-29
Added option nclient4.WithHWAddr

xaionaro:
This option allows to receive answers DHCP with arbitrary client addresses.


insomniacslk:
Cc @pmazzini

pmazzini:
> I think there's a modifier called WithHwAddr... This is gonna get confusing

Right. Any suggestion?

If we use the `WithHwAddr` packet modifier alone it will be not enough because the client will wait for packets with a different MAC, that is the reason for adding a client modifier.

We can change it if you think there is a better alternative.

PR #338, opened 2019-12-03
nclient4: use ulog instead of Printfer

hugelgupf:
We already have a Printfer-like thing in u-root via pkg/ulog. Use that instead: https://godoc.org/github.com/u-root/u-root/pkg/ulog#Logger

### DIFF
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/u-root/u-root/pkg/ulog"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pmazzini:
cc @xaionaro.

dhcpv4/nclient4/client.go
```diff
@@ -76,58 +77,39 @@ type pendingCh struct {
 
 // Logger is a handler which will be used to output logging messages
 type Logger interface {
-	// PrintMessage print _all_ DHCP messages
-	PrintMessage(prefix string, message *dhcpv4.DHCPv4)
+	ulog.Logger
```
xaionaro: In my extremely humble opinion it is wrong to require `Print(v ...interface{})` if you do not use it and do not plan to use it.

Natolumin:
I don't understand the point of this. Does it make anything better ?
Interfaces are automatically implemented when when all their functions are implemented by a type, so adding this external interface here adds nothing of value; and now you have to go to a completely separate project just to find its (trivial!) definition.

PR #340, opened 2019-12-10
Add option BootfileParam

xaionaro:
See RFC5970 section 3.2.

* Added support of option OptBootFileParam
* Added BootfileParam to netboot results

### DIFF
--- a/dhcpv6/client6/client.go
+++ b/dhcpv6/client6/client.go
@@ -177,6 +177,10 @@ func (c *Client) sendReceive(ifname string, packet dhcpv6.DHCPv6, expectedType d
 		adv, err = dhcpv6.FromBytes(buf[:n])
 		if err != nil {
 			// skip non-DHCP packets
+			//
+			// TODO: It also skips DHCP packets with any errors (for example
+			// if bootfile params are encoded incorrectly). We need to
+			// log such cases instead of silently skip them.
 			continue
 		}
 		if recvMsg, ok := adv.(*dhcpv6.Message); ok && isMessage {
--- a/dhcpv6/option_bootfileparam.go
+++ b/dhcpv6/option_bootfileparam.go
@@ -1,0 +1,60 @@
+package dhcpv6
+
+import (
+	"fmt"
+
+	"github.com/u-root/u-root/pkg/uio"
+)
+
+// OptBootFileParam implements the OptionBootfileParam option
+//
+// This module defines the OPT_BOOTFILE_PARAM structure.
+// https://www.ietf.org/rfc/rfc5970.txt (section 3.2)
+type OptBootFileParam []string
+
+var _ Option = OptBootFileParam(nil)
+
+// Code returns the option code
+func (op OptBootFileParam) Code() OptionCode {
+	return OptionBootfileParam
+}
+
+// ToBytes serializes the option and returns it as a sequence of bytes
+func (op OptBootFileParam) ToBytes() []byte {
+	buf := uio.NewBigEndianBuffer(nil)
+	for _, param := range op {
+		if len(param) >= 1<<16 {
+			// TODO: say something here instead of silently ignoring a parameter
+			continue
+		}
+		buf.Write16(uint16(len(param)))
+		buf.WriteBytes([]byte(param))
+		/*if err := buf.Error(); err != nil {
+			// TODO: description of `WriteBytes` says it could return
+			// an error via `buf.Error()`. But a quick look into implementation of
+			// `WriteBytes` at the moment of this comment showed it does not set any
+			// errors to `Error()` output. It's required to make a decision:
+			// to fix `WriteBytes` or it's description or
+			// to find a way to handle an error here.
+		}*/
+	}
+	return buf.Data()
+}
+
+func (op OptBootFileParam) String() string {
+	return fmt.Sprintf("OptBootFileParam(%v)", ([]string)(op))
+}
+
+// ParseOptBootFileParam builds an OptBootFileParam structure from a sequence
+// of bytes. The input data does not include option code and length bytes.
+func ParseOptBootFileParam(data []byte) (result OptBootFileParam, err error) {
+	buf := uio.NewBigEndianBuffer(data)
+	for buf.Has(2) {
+		length := buf.Read16()
+		result = append(result, string(buf.CopyN(int(length))))
+	}
+	if err := buf.FinError(); err != nil {
+		return nil, err
+	}
+	return
+}
--- a/dhcpv6/option_bootfileparam_test.go
+++ b/dhcpv6/option_bootfileparam_test.go
@@ -1,0 +1,71 @@
+package dhcpv6
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	testBootfileParams0Compiled = "\x00\x0eroot=/dev/sda1\x00\x00\x00\x02rw"
+	testBootfileParams1 = []string{
+		"initrd=http://myserver.mycompany.local/initrd.xz",
+		"",
+		"root=/dev/sda1",
+		"rw",
+		"netconsole=..:\000:.something\000here.::..",
+		string(make([]byte, (1<<16) - 1)),
+	}
+)
+
+// compileTestBootfileParams is an independent implementation of bootfile param encoder
+func compileTestBootfileParams(t *testing.T, params []string) []byte {
+	var length [2]byte
+	buf := bytes.Buffer{}
+	for _, param := range params {
+		if len(param) >= 1<<16 {
+			panic("a too long parameter")
+		}
+		binary.BigEndian.PutUint16(length[:], uint16(len(param)))
+		_, err := buf.Write(length[:])
+		require.NoError(t, err)
+		_, err = buf.WriteString(param)
+		require.NoError(t, err)
+	}
+
+	return buf.Bytes()
+}
+
+func TestOptBootFileParam(t *testing.T) {
+	expected := string(compileTestBootfileParams(t, testBootfileParams1))
+	opt, err := ParseOptBootFileParam([]byte(expected))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(opt.ToBytes()) != expected {
+		t.Fatalf("Invalid boot file parameter. Expected %v, got %v", expected, opt)
+	}
+}
+
+func TestParsedTypeOptBootFileParam(t *testing.T) {
+	tryParse := func(compiled []byte, expected []string) {
+		opt, err := ParseOption(OptionBootfileParam, compiled)
+		require.NoError(t, err)
+		bootfileParamOpt, ok := opt.(OptBootFileParam)
+		require.True(t, ok, fmt.Sprintf("invalid type: %T instead of %T", opt, bootfileParamOpt))
+		require.Equal(t, compiled, bootfileParamOpt.ToBytes())
+		require.Equal(t, expected, ([]string)(bootfileParamOpt))
+	}
+
+	tryParse(
+		[]byte(testBootfileParams0Compiled),
+		[]string{"root=/dev/sda1", "", "rw"},
+	)
+	tryParse(
+		compileTestBootfileParams(t, testBootfileParams1),
+		testBootfileParams1,
+	)
+}
--- a/dhcpv6/options.go
+++ b/dhcpv6/options.go
@@ -75,6 +75,8 @@ func ParseOption(code OptionCode, optData []byte) (Option, error) {
 		opt, err = ParseOptRemoteId(optData)
 	case OptionBootfileURL:
 		opt, err = ParseOptBootFileURL(optData)
+	case OptionBootfileParam:
+		opt, err = ParseOptBootFileParam(optData)
 	case OptionClientArchType:
 		opt, err = ParseOptClientArchType(optData)
 	case OptionNII:
--- a/netboot/netboot.go
+++ b/netboot/netboot.go
@@ -16,6 +16,20 @@ var sleeper = func(d time.Duration) {
 	time.Sleep(d)
 }
 
+// BootConf is a structure describes everything a host needs to know to boot over network
+type BootConf struct {
+	// NetConf is the network configuration of the client
+	NetConf
+
+	// BootfileURL is "where is the image (kernel)".
+	// See RFC5970 section 3.1 for IPv6 and RFC2132 section 9.5 ("Bootfile name") for IPv4
+	BootfileURL string
+
+	// BootfileParam is "what arguments should we pass (cmdline)".
+	// See RFC5970 section 3.2 for IPv6.
+	BootfileParam []string
+}
+
 // RequestNetbootv6 sends a netboot request via DHCPv6 and returns the exchanged packets. Additional modifiers
 // can be passed to manipulate both solicit and advertise packets.
 func RequestNetbootv6(ifname string, timeout time.Duration, retries int, modifiers ...dhcpv6.Modifier) ([]dhcpv6.DHCPv6, error) {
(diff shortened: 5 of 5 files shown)


dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+)
+
+// OptBootFileParam implements the OptionBootfileParam option
+//
+// This module defines the OPT_BOOTFILE_PARAM structure.
+// https://www.ietf.org/rfc/rfc5970.txt (section 3.2)
+type OptBootFileParam string
```
Natolumin: If I read the RFC properly, this should be []string instead ? You can have multiple strings in a single OPT_BOOTFILE_PARAM option, they're encoded as 16-bit length fields followed by utf-8 strings

xaionaro: Khm, yeah, you right. OK, I'll fix it. Thank you.

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,76 @@
+package dhcpv6
+
+import (
+	"bytes"
+	"encoding/binary"
```
Natolumin: This library has been using `github.com/u-root/u-root/pkg/uio` pretty extensively for parsing/writing, it has a couple useful features esp. around error handling that you end up reimplementing in this PR. I think it would be a good idea to also use it here

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,76 @@
...
+	return OptionBootfileParam
+}
+
+// ToBytes serializes the option and returns it as a sequence of bytes
+func (op OptBootFileParam) ToBytes() []byte {
+	var length [2]byte
+	var buf bytes.Buffer
+	for _, param := range op {
+		bootpEndian.PutUint16(length[:], uint16(len(param)))
+		buf.Write(length[:])
```
Natolumin: For example, I think this little dance would be clearer/simpler with uio

dhcpv6/client6/client.go
```diff
@@ -177,6 +177,10 @@ func (c *Client) sendReceive(ifname string, packet dhcpv6.DHCPv6, expectedType d
 		adv, err = dhcpv6.FromBytes(buf[:n])
 		if err != nil {
 			// skip non-DHCP packets
+			//
+			// TODO: It also skips DHCP packets with any errors (for example
```
Natolumin: is this TODO for this PR or a future one ? It would be useful to make an issue for this if it won't be addressed in this PR

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,76 @@
...
+	return fmt.Sprintf("[bootfile param] unexpected end (expected: %d; left %d)", err.Expected, err.Left)
+}
+
+// OptBootFileParam implements the OptionBootfileParam option
+//
+// This module defines the OPT_BOOTFILE_PARAM structure.
+// https://www.ietf.org/rfc/rfc5970.txt (section 3.2)
+type OptBootFileParam []string
+
+var _ Option = OptBootFileParam(nil)
```
Natolumin: What is this for ?

dhcpv6/option_bootfileparam_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testBootfileParam = "\x00\x30initrd=http://myserver.mycompany.local/initrd.xz\x00\x0eroot=/dev/sda1\x00\x02rw\x00\x24netconsole=..:\000:.something\000here.::.."
+)
+
+func TestOptBootFileParam(t *testing.T) {
+	expected := testBootfileParam
+	opt, err := ParseOptBootFileParam([]byte(expected))
```
Natolumin: This doesn't really test the multi-option parsing code, it only fills in a single string even though your example is clearly intended to expand into multiple strings

dhcpv6/option_bootfileparam_test.go
```diff
@@ -0,0 +1,31 @@
...
+		t.Fatalf("Invalid boot file parameter. Expected %v, got %v", expected, opt)
+	}
+}
+
+func TestParsedTypeOptBootFileParam(t *testing.T) {
+	opt, err := ParseOption(OptionBootfileParam, []byte(testBootfileParam))
+	require.NoError(t, err)
+	bootfileParamOpt, ok := opt.(OptBootFileParam)
+	require.True(t, ok, fmt.Sprintf("invalid type: %T instead of %T", opt, bootfileParamOpt))
+	require.Equal(t, testBootfileParam, string(bootfileParamOpt.ToBytes()))
```
Natolumin: Testing that the round-trip is unchanged is good, but it would also be nice to test that the parsed value is correct. If for example the option were parsing as a single string (as the original implementation did) instead of 4, this test would pass when it should fail.

dhcpv6/option_bootfileparam.go
xaionaro: To check if `OptBootFileParam` implements `Option` on build time. In this case it was actually copied from [`option_bootfileurl.go`](https://github.com/insomniacslk/dhcp/blob/3997b8a58c66233275ad9b26612ef733a63bd9d3/dhcpv6/option_bootfileurl.go#L13).

xaionaro:
> The changes in netboot/ should probably be a separate commit, if not a separate PR. I haven't looked at them yet as I don't really know that code

I'll make a separate commit if it's good enough :)

dhcpv6/option_bootfileparam.go
xaionaro: Created an issue: https://github.com/insomniacslk/dhcp/issues/341

dhcpv6/client6/client.go
xaionaro: Created an issue: https://github.com/insomniacslk/dhcp/issues/341

dhcpv6/option_bootfileparam.go
xaionaro: Well. I don't understand why `uio` was used over `binary`. The error handling did not become easier (I still have to write `if`s in the same places). Moreover with `uio` errors became ["anonymous"](https://github.com/u-root/u-root/blob/801a65c9a22bcc5dba4d02208b0c94b074dc9bae/pkg/uio/buffer.go#L100), so they cannot be compared/checked in any way in future (except substring-search which is not reliable). But I agree that code consistency is important. So OK and thanks :)

dhcpv6/option_bootfileparam_test.go
xaionaro: Yeah, a good thought.

dhcpv6/option_bootfileparam_test.go
xaionaro: Added multi-option test to `TestParsedTypeOptBootFileParam`

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,68 @@
...
+func (op OptBootFileParam) Code() OptionCode {
+	return OptionBootfileParam
+}
+
+// ToBytes serializes the option and returns it as a sequence of bytes
+func (op OptBootFileParam) ToBytes() []byte {
+	buf := uio.NewBigEndianBuffer(nil)
+	for _, param := range op {
+		if len(param) >= 1<<16 {
+			panic("a too long parameter")
```
Natolumin: I don't think panicking is OK here. this would be the only `panic()` outside of a test in this entire library. (this is the same issue as the previous comment, that the ToBytes() api really should return an error). 
This might end up being acceptable, but we need to make a conscious decision that ToBytes() can panic and clearly communicate that to users

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,68 @@
...
+
+		param, err := buf.ReadN(int(length))
+		if err != nil {
+			return nil, fmt.Errorf("[bootfile param] unable to read the parameter value: %w", err)
+		}
+
+		result = append(result, string(param))
+
+	}
+	if err := buf.FinError(); err != nil {
```
Natolumin: Part of the idea of uio is that you only need to check this error here at the end, and any previously encountered error would be folded into this. So you don't need the previous 2 error checks, which should make the loop much simpler

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,68 @@
...
+// of bytes. The input data does not include option code and length bytes.
+func ParseOptBootFileParam(data []byte) (result OptBootFileParam, err error) {
+	buf := uio.NewBigEndianBuffer(data)
+	for buf.Has(2) {
+		length := buf.Read16()
+		if err := buf.Error(); err != nil {
+			return nil, fmt.Errorf("[bootfile param] unable to read the parameter length: %w", err)
+		}
+
+		param, err := buf.ReadN(int(length))
```
Natolumin: I think you want `buf.CopyN` here instead, which populates `Error()` like the other `Read*` functions. `ReadN` is more of a low-level functions and gives you a direct slice into the underlying buffer (I do think the name ambiguity between the uio.Buffer and uio.Lexer functions is misleading)

This doesn't matter too much since you convert to string anyway, but it would allow you to drop the error check as well.

dhcpv6/option_bootfileparam.go
pmazzini: Good catch. I totally missed this one.

xaionaro: @Natolumin I agree that a `panic()` is bad. But what do you suggest? Just ignore an error? Because there's no `error` in returnings of `ToBytes()`.

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		buf.Write16(uint16(len(param)))
+		buf.WriteBytes([]byte(param))
+		/*if err := buf.Error(); err != nil {
+			// TODO: description of `WriteBytes` says it could return
+			// an error via `buf.Error()`. But a quick look into implementation of
+			// `WriteBytes` at the moment of this comment showed it does not set any
+			// errors to `Error()` output. It's required to make a decision:
+			// to fix `WriteBytes`'s description or to find a way to handle an error here.
+		}*/
```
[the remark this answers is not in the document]
xaionaro: Should I leave the comment as is. Or you want me to remove it?

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// ParseOptBootFileParam builds an OptBootFileParam structure from a sequence
+// of bytes. The input data does not include option code and length bytes.
+func ParseOptBootFileParam(data []byte) (result OptBootFileParam, err error) {
+	buf := uio.NewBigEndianBuffer(data)
+	for buf.Has(2) {
+		length := buf.Read16()
+		if err := buf.Error(); err != nil {
+			return nil, fmt.Errorf("[bootfile param] unable to read the parameter length: %w", err)
```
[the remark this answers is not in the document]
xaionaro: We will get an error message, but it will look a dirty, like in case with `Sprintf` here:
https://play.golang.org/p/ZZyAHz0qq2S

```
test: %!w(*errors.errorString=&{some error})
test: some error
```
(the first line is for go<1.13; the second is for go>=1.13)

dhcpv6/option_bootfileparam.go
xaionaro: > I think you want `buf.CopyN` here instead

Yeah, such a method was intended initially. And yes, I mistook because of this:

> I do think the name ambiguity between the uio.Buffer and uio.Lexer functions is misleading

However ReadN isn't worse is this case, because (as you already said) I convert it to string, anyway. And ReadN even allows to avoid an extra memory allocation and copy. And the error check will be required anyway -- I should not add `string([]byte(nil))` to `result`. On the other hand a next programmer may think it returns a copy (because the name looks like other `Read*` functions) and make a mistake. I can change it if required, should I?

dhcpv6/option_bootfileparam.go
xaionaro: How it will exit the loop in case of error (if it's unable to consume the last byte)? I still have to add at least one error check into the loop. Also I would like to add explanation text to the error if possible (like `[bootfile param]`). However if you think I should as much drop error checks as possible anyway, then I will do it. Should drop one check?

dhcpv6/option_bootfileparam.go
pmazzini: As per `hugelgupf` in https://github.com/insomniacslk/dhcp/pull/329:

```
I don't think this library should concern itself with validation of the values. My take is - the user is responsible.
```

I would keep it consistent with other options and don't panic in here.

dhcpv6/option_bootfileparam_test.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+		binary.BigEndian.PutUint16(length[:], uint16(len(param)))
+		_, err := buf.Write(length[:])
+		require.NoError(t, err)
+		_, err = buf.WriteString(param)
+		require.NoError(t, err)
+	}
+
+	return buf.Bytes()
+}
```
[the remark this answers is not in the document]
xaionaro: Well, in my understanding the pre-serialized case has:

Advantages:

* Less lines. So the it's required less time to read the code.

Disadvantages:

* Harder to read the test (to verify by eyes that the code works correctly).
* Harder to support the test (for example to add a test case).
* It is very inconvenient to prepare the parameter `string(make([]byte, (1<<16) - 1))` as pre-serialized value.
* For me it was just easier/faster to write a function than to calculate lengths, translate it to hex and compile it all manually.

But of course I may be wrong here and I can change it to a pre-serialized implementation of the test if required. Should I?

dhcpv6/option_bootfileparam.go
xaionaro: In my very-very humble opinion, it sounds terrible. Even standard Golang packages panics on unhandled errors. But OK. I will just drop huge parameters (use `continue` instead of `panic()`).

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,68 @@
...
+			panic("a too long parameter")
+		}
+		buf.Write16(uint16(len(param)))
+		buf.WriteBytes([]byte(param))
+		/*if err := buf.Error(); err != nil {
+			// TODO: description of `WriteBytes` says it could return
+			// an error via `buf.Error()`. But a quick look into implementation of
+			// `WriteBytes` at the moment of this comment showed it does not set any
+			// errors to `Error()` output. It's required to make a decision:
+			// to fix `WriteBytes`'s description or to find a way to handle an error here.
```
hugelgupf: or just fix WriteBytes.

dhcpv6/option_bootfileparam.go
```diff
@@ -0,0 +1,68 @@
...
+	return fmt.Sprintf("OptBootFileParam(%v)", ([]string)(op))
+}
+
+// ParseOptBootFileParam builds an OptBootFileParam structure from a sequence
+// of bytes. The input data does not include option code and length bytes.
+func ParseOptBootFileParam(data []byte) (result OptBootFileParam, err error) {
+	buf := uio.NewBigEndianBuffer(data)
+	for buf.Has(2) {
+		length := buf.Read16()
+		if err := buf.Error(); err != nil {
```
hugelgupf: the only reason it wouldn't be able to read 2 bytes is if there weren't 2 bytes left. length would return 0, and buf.Has(2) on the next iteration will be false (it should have been false to begin with...)

either way -- checking error here is unnecessary. FinError will take care of it.

dhcpv6/option_bootfileparam.go
hugelgupf: you're right about the confusion, it's weird. do you have a suggestion?

I think here you should use CopyN -- it makes a copy, but it avoids the unnecessary error check.

hugelgupf: (there should probably be a ReadN that doesn't return an error and folds it up to FinError)

dhcpv6/option_bootfileparam.go
hugelgupf: The only reason that Read16 or ReadN/CopyN would fail is because there aren't enough bytes left in the buffer.

buf.Has(2) should cover that. there should be no error check in the loop.

which check do you think still has to be there?

dhcpv6/option_bootfileparam_test.go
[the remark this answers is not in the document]
hugelgupf: copying & pasting the implementation to compare against is kind of against the spirit of unit tests like this.. you should be manually calculating some values, as eye-bleeding as it is, and comparing against those serializations.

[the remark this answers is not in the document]
xaionaro: @hugelgupf this is not a copy&paste. Logic is the same, but the code is reimplemented (using different package). When I will compile a string by myself I will use the same logic (as I implemented in the test).

dhcpv6/option_bootfileparam.go
xaionaro: @hugelgupf well, OK, but there's no description in the functions `Read*` that it's the only reason why it could fail, so there's no guarantee that it won't be changed in future. Anyway OK, I will fix it.

xaionaro:
I've fixed everything :)

dhcpv6/option_bootfileparam_test.go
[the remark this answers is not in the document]
hugelgupf: You used the same logic .. but you are testing that logic.

If you made a mistake in the original, stands to reason you'd make the same mistake here.

PR #343, opened 2019-12-28
v6: introduce Getters for Options

hugelgupf:
Allow the Options type to have getters for each specific options, in
order to avoid users having to cast options to their specific type.

This commit introduces a getter for exactly one option: the
ClientArchType.

i.e. users can replace

```go
archTypes := msg.GetOneOption(OptionClientArchType).(*OptClientArchType)
```

with

```go
archTypes := msg.Options.ArchTypes()
```

Or more specifically,
```go
opt = d.GetOneOption(dhcpv6.OptionDNSRecursiveNameServer)
if opt == nil {		
  return nil, errors.New("No option DNS Recursive Name Servers found ")
}
odnsserv := opt.(*dhcpv6.OptDNSRecursiveNameServer)
netconf.DNSServers = odnsserv.NameServers
```
can be replaced with...
```go
dns := d.Options.DNS()
if len(dns) == 0 {
  return nil, errors.New("no option DNS Recursive Name Servers found")
}
netconf.DNSServers = dns
```

Because a few message types and options embed options (normal message,
relay message, IANA/IATA option) and each have a restricted set of
options that can be used inside them, we'll introduce at least 3 or more
Options subtypes:

- MessageOptions
- RelayOptions
- IdentityOptions

Perhaps others will join at a later time, such as VendorOptions or
AddressOptions for the IAAddress options field.


hugelgupf:
cc @stapelberg

dhcpv6/option_clientid_test.go
```diff
@@ -8,27 +8,27 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestParseOptClientId(t *testing.T) {
+func TestparseOptClientID(t *testing.T) {
```
golangcibot: tests: TestparseOptClientID has malformed name: first letter after 'Test' must not be lowercase (from `govet`)

dhcpv6/option_serverid_test.go
```diff
@@ -8,27 +8,27 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestParseOptServerId(t *testing.T) {
+func TestparseOptServerID(t *testing.T) {
```
golangcibot: tests: TestparseOptServerID has malformed name: first letter after 'Test' must not be lowercase (from `govet`)

pmazzini:
This is not consistent with the DHCPv4 library where the getters are applied directly to the DHCP struct instead of the Options.

Should `archTypes := msg.Options.ArchTypes()` be `archTypes := msg.ArchTypes()` instead to make it consistent with the v4 library?

hugelgupf:
> This is not consistent with the DHCPv4 library where the getters are applied directly to the DHCP struct instead of the Options.
> 
> Should `archTypes := msg.Options.ArchTypes()` be `archTypes := msg.ArchTypes()` instead to make it consistent with the v4 library?

Hey, sorry, finally picking this up again. It's a fair question. I did it this way so IdentityOptions could be shared by IA_NA and IA_TA (when it exists some day). I think it's okay for it to not be quite exactly the same.

pmazzini:
Ok, cc @Natolumin to see if he has any opinion on this after https://github.com/insomniacslk/dhcp/pull/346#pullrequestreview-341714747.

hugelgupf:
ping

insomniacslk:
apologies for the delay, I am catching up with a queue of ~20 PRs. Looking at this now

insomniacslk:
I agree we don't have to be necessarily consistent with the `dhcpv4` package API (even if desirable), since the two protocols share part of the name and not much else

Natolumin:
My opinion is the same as in #346, but it's not a big deal, it just seems simpler (to me) to have it on the option rather than the suboptions list, in a way that I don't think sharing IdentityOptions (which will be a like 5-line function) compensates

Feel free to disagree, that's just bikeshedding

dhcpv6/dhcpv6message.go
```diff
@@ -13,11 +13,140 @@ import (
 
 const MessageHeaderSize = 4
 
+// MessageOptions are the options that may appear in a normal DHCPv6 message.
+//
+// RFC 3315 Appendix B lists the valid options that can be used.
+type MessageOptions struct {
```
insomniacslk: well, this is much better than what we have now :)

dhcpv6/option_clientid.go
```diff
@@ -4,36 +4,31 @@ import (
...
-	Cid Duid
+// OptClientID represents a Client Identifier option as defined by RFC 3315
+// Section 22.2.
+func OptClientID(d Duid) Option {
+	return &optClientID{d}
 }
 
-func (op *OptClientId) Code() OptionCode {
-	return OptionClientID
+type optClientID struct {
```
insomniacslk: do we want an unexported typed with an exported field? And in general I'm not sure this should become private

(same below)

hugelgupf: I think to the user these types are mostly noise in the godoc. The user wants two things: A way to add the type to options (`OptClientID` function), and a way to get the type from a message/option (`ClientID()` getter)

Where I do think the types should be exported is where it makes multi-member options more readable, e.g.

```go
msg.Options.Update(&OptIANA{
  IAID: id,
  T1: x,
  T2: y,
  Options: IdentityOptions(opts),
})
```
I think is clearer than
```go
msg.Options.Update(OptIANA(id, x, y, opts))
```

hugelgupf: And yeah, the member should probably be unexported. Though since the Type is unexported, it doesn't make that much of a difference to anything

hugelgupf:
ping

dhcpv6/dhcpv6.go
```diff
@@ -204,13 +203,9 @@ func IsUsingUEFI(msg *Message) bool {
...
 	//               8    EFI Xscale
 	//               9    EFI x86-64
-	if opt := msg.GetOneOption(OptionClientArchType); opt != nil {
-		optat := opt.(*OptClientArchType)
-		for _, at := range optat.ArchTypes {
-			// TODO investigate if other types are appropriate
-			if at == iana.EFI_BC || at == iana.EFI_X86_64 {
-				return true
-			}
+	if archTypes := msg.Options.ArchTypes(); archTypes != nil {
```
pmazzini: This is changing the current behaviour of the function. If a packet has multiple arch type options it will check only one. I have seen clients sending packets with more than one option type and different info in each of them. I think that in v4 we merge them in the map.

dhcpv6/dhcpv6message.go
```diff
@@ -237,20 +361,13 @@ func (m *Message) IsNetboot() bool {
...
 func (m *Message) IsOptionRequested(requested OptionCode) bool {
-	for _, optoro := range m.GetOption(OptionORO) {
-		for _, o := range optoro.(*OptRequestedOption).RequestedOptions() {
-			if o == requested {
-				return true
-			}
-		}
-	}
-	return false
+	return m.Options.RequestedOptions().Contains(requested)
```
pmazzini: Same as above. If there are multiple `OptionORO` now you check only one.

hugelgupf: In v4, merging them is part of the RFC because of length. That's not part of v6 afaict.

RFC 8415:
>   Unless otherwise noted, each option may appear only in the options area of a DHCP message and may appear only once. 

Chapter 21.7, Option Request Option, does not appear to mention anything.

Why is the client not merging OROs?

dhcpv6/dhcpv6.go
hugelgupf: No, it doesn't. Previously, this used `GetOneOption`, which only returns one option. The option itself contains multiple arch types in it, though, and that behavior is preserved with the for loop here.

hugelgupf: `iana.Archs` is a `[]iana.Arch` with a `Contains` function that has the for loop.

dhcpv6/dhcpv6message.go
hugelgupf: As above, this is an array of option codes. `Contains` is a function that checks them all.

pmazzini: > Why is the client not merging OROs?

I have seen it in proprietary clients. I would still support multiple OROs.

pmazzini: Yes, but `RequestedOptions` only returns the first `ORO`.

dhcpv6/dhcpv6.go
pmazzini: You are right, so my concern is with `OROs` then.

dhcpv6/dhcpv6message.go
hugelgupf: `RequestedOptions` now merges them.

pmazzini: Thanks!

pmazzini:
> ping

Sorry for the delay. In general it looks ok, I'll give it a deeper look tomorrow morning unless somebody wants to stamp it before that.

pmazzini:
We need stretchr/testify#881 to get travis build working again.

hugelgupf:
Thanks. That PR is merged. @insomniacslk, merge when ready?

hugelgupf:
opened https://github.com/rtr7/router7/pull/37 for rtr7.

dhcpv6/option_clientid.go
insomniacslk: > Though since the Type is unexported, it doesn't make that much of a difference to anything

The thing is that you can return an unexported object, and it can have exported attributes. For example
```
package lib

type unexported struct {
	Name string
}

func Get() *unexported {
	return &unexported{
		Name: "unexported",
	}
}
```

and then calling `lib.Get().Name` from another package will just work.

I remember some linter warning complaining about this type of declaration, but I guess we can live with it

pmazzini:
Triggering a new travis build after rtr7/router7#37 was merged.

stapelberg:
> Triggering a new travis build after [rtr7/router7#37](https://github.com/rtr7/router7/pull/37) was merged.

It wasn’t merged yet. I’d prefer it if you could submit first, and then I merge.

insomniacslk:
I'm re-reading this PR regarding whether the methods should be in the options or in the top level object, hopefully before Chris wakes up :)

pmazzini:
> > Triggering a new travis build after [rtr7/router7#37](https://github.com/rtr7/router7/pull/37) was merged.
> 
> It wasn’t merged yet. I’d prefer it if you could submit first, and then I merge.

Yes, sorry, I meant to mention stretchr/testify#881.

hugelgupf:
The testify change was merged already. Travis will be broken until both this and rtr7 are merged, and this change needs to go in first.

insomniacslk:
I'm also not too convinced on whether the getters should live, but since I am struggling to find time for a proper review, and the change is technically good and reviewed, I don't want to block this. @pmazzini since you did most of the review, are you OK to merge this?

pmazzini:
> I'm also not too convinced on whether the getters should live
> the change is technically good 

I am on the same boat. I think the change is ok but I am not convinced if it is the right place to put the getters.

@hugelgupf How about putting the getters directly on top of the [options](https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/options.go#L102) without having a dedicated group of options such as MessageOptions/RelayOptions/IdentityOptions. I think it will make things simpler.

hugelgupf:
> > I'm also not too convinced on whether the getters should live
> > the change is technically good
> 
> I am on the same boat. I think the change is ok but I am not convinced if it is the right place to put the getters.
> 
> @hugelgupf How about putting the getters directly on top of the [options](https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/options.go#L102) without having a dedicated group of options such as MessageOptions/RelayOptions/IdentityOptions. I think it will make things simpler.

Shared options - yeah, I'm okay with doing that for StatusCode(). However, the option groups exist in order to tell users which things are useful on which message in line with the RFC and usage in the wild.

When using the library, I don't know if for example the FQDN option is supposed to be in a DHCPv6 message, in a relay message, in an IANA or an IATA. Where can it appear, which places do I have to check? The option groups are supposed to answer that question.

hugelgupf:
I added an FQDN getter, and I found that I hadn't made a DomainSearchList thing either yet.

pmazzini:
> the option groups exist in order to tell users which things are useful on which message

The library doesn't prevent a user from adding an option in the wrong place, so some knowledge is still needed from the user side.

10000TB:
How is option parsing error addressed  ?

hugelgupf:
> > the option groups exist in order to tell users which things are useful on which message
> 
> The library doesn't prevent a user from adding an option in the wrong place, so some knowledge is still needed from the user side.

Solving one problem at a time...

pmazzini:
> are you OK to merge this?

Fine with me but it would be nice if someone else can take a look also.

hugelgupf:
ping.

PR #345, opened 2019-12-28
v6: remote IAPrefix option

### LINKED ISSUE #344: DHCPv6: IAPrefix option incomplete
The v6 package has the IAPrefix option implemented, but not the IA_PD option. However, RFC 3633 states that IAPrefix must be encapsulated in the Options field of the IA_PD option. So the current IAPrefix cannot be used the correct way.

Is anyone using it incorrectly somehow? If not, I'll just remove it since it cannot be used correctly at the moment.

hugelgupf:
As implemented, it cannot be used correctly.

RFC 3633 states the IAPrefix option must be encap'd in an IA_PD option,
but this library doesn't support IA_PD. It's impossible to use the
option correctly as implemented.

Closes #344.


hugelgupf:
Aaaah, I found it. Wow.

PR #347, opened 2019-12-28
v6: use time.Duration for duration fields

hugelgupf:


hugelgupf:
also cc @stapelberg, I see you converting these manually with casts. Those casts may fail to be correct after this change.

stapelberg:
Thanks for the heads up. I can commit a fix once this is in, or you could send a PR, whichever you prefer :)

PR #349, opened 2019-12-28
v6: introduce IdentityOptions and getters

hugelgupf:
IA_NA and IA_TA options can contain two fields:

- IAAddress
- Status Code

IAAddress may contain one field:

 - Status Code

(See RFC 8415 Appendix C for this info.)

This adds two types, IdentityOptions and AddressOptions with Getters for each of the allowed option types.


hugelgupf:
Updated.

PR #351, opened 2020-01-02
Fix "conn" leaking

martelli:
On machines with many NICs, it can cause the golang runtime to run out of threads.

### DIFF
--- a/netboot/rtnetlink_linux.go
+++ b/netboot/rtnetlink_linux.go
@@ -1,6 +1,10 @@
 package netboot
 
-import "github.com/jsimonetti/rtnetlink"
+import (
+	"log"
+
+	"github.com/jsimonetti/rtnetlink"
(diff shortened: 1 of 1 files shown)


pmazzini:
Go tip linter check is failing but it is unrelated to this PR

https://travis-ci.org/insomniacslk/dhcp/builds/631842573

pmazzini:
This looks to be a golangci-lint issue: https://github.com/golangci/golangci-lint/issues/896

I would go ahead and merge this PR but I can only do so if travis run was successful. cc @insomniacslk

insomniacslk:
@pmazzini i can force the merge. However I'd like a check of the return value of Close() before, and logging a warning if it fails, cc @martelli

martelli:
I've refactored it to log any error messages.

PR #352, opened 2020-01-09
interfaces: add support for AIX

Helflym:


insomniacslk:
the build failure on go/tip is unrelated to this PR.

@Helflym also note that we don't build this library for AIX in the CI (just like we don't build it for freebsd/netbsd/etc), so that is not guaranteed either, and should be added to the Travis-CI configuration

PR #353, opened 2020-01-24
Fix a compare in Serve() of IP4 Zero to use an IP4 address

rminnich:
Fixes, among other things, u-root pxeserver


insomniacslk:
Broken go tip, we are aware of the issue, I'll force the merge (I guess we can temporarily disabled tip until it's working again)

PR #356, opened 2020-02-07
nclient6: allow non-rapid responses to RapidSolicit

hugelgupf:
Supersedes #317. cc @rminnich.

### DIFF
--- a/dhcpv6/nclient6/client.go
+++ b/dhcpv6/nclient6/client.go
@@ -288,11 +288,17 @@ func WithDebugLogger() ClientOpt {
 type Matcher func(*dhcpv6.Message) bool
 
 // IsMessageType returns a matcher that checks for the message type.
-//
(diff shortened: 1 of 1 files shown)


dhcpv6/nclient6/client.go
```diff
@@ -290,9 +290,17 @@ type Matcher func(*dhcpv6.Message) bool
 // IsMessageType returns a matcher that checks for the message type.
 //
 // If t is MessageTypeNone, all packets are matched.
```
pmazzini: Why does it match if MessageType is None?

pmazzini:
Other than the above comment, this looks good to me.

dhcpv6/nclient6/client.go
hugelgupf: yeah you're right i'm not sure that made sense

PR #365, opened 2020-03-07
dhcpv6: always write 16-byte IPs

hugelgupf:


hugelgupf:
Hey @insomniacslk there's some nftables thing again...

PR #366, opened 2020-03-07
dhcp6: fix VendorOpts getter

hugelgupf:
RFC 8415 Section 21.17 states it may appear more than once:

   Multiple instances of the Vendor-specific Information option may
   appear in a DHCP message.  Each instance of the option is interpreted
   according to the option codes defined by the vendor identified by the
   Enterprise Number in that option.  Servers and clients MUST NOT send
   more than one instance of the Vendor-specific Information option with
   the same Enterprise Number.

### DIFF
--- a/dhcpv6/dhcpv6message.go
+++ b/dhcpv6/dhcpv6message.go
@@ -179,16 +179,40 @@ func (mo MessageOptions) UserClasses() [][]byte {
 	return nil
 }
 
-// VendorOpts returns the enterprise number and a list of vendor options.
-func (mo MessageOptions) VendorOpts() (uint32, Options) {
-	opt := mo.Options.GetOne(OptionVendorOpts)
+// VendorOpts returns the all vendor-specific options.
+//
+// RFC 8415 Section 21.17:
+//
+//   Multiple instances of the Vendor-specific Information option may appear in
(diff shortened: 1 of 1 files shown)


dhcpv6/dhcpv6message.go
```diff
@@ -179,16 +179,40 @@ func (mo MessageOptions) UserClasses() [][]byte {
...
-// VendorOpts returns the enterprise number and a list of vendor options.
-func (mo MessageOptions) VendorOpts() (uint32, Options) {
-	opt := mo.Options.GetOne(OptionVendorOpts)
+// VendorOpts returns the all vendor-specific options.
+//
+// RFC 8415 Section 21.17:
+//
+//   Multiple instances of the Vendor-specific Information option may appear in
+//   a DHCP message.
+func (mo MessageOptions) VendorOpts() []*OptVendorOpts {
```
pmazzini: Why are you not returning the `EnterpriseNumber`?

hugelgupf: I am, though? OptVendorOpts includes the Enterprise num?

pmazzini: Oh, right.

PR #368, opened 2020-03-10
dhcpv6 fqdn: fix domain encoding

### LINKED ISSUE #360: dhcpv6: Problem with FQDN option encoding
There is a problem with the encoding of the FQDN option. According to RFC 4704 section 4.2:

> The data in the Domain Name field MUST be encoded as described in Section 8 of [5] (RFC 3315)

And according to RFC 3315 section 8:

>   So that domain names may be encoded uniformly, a domain name or a
>    list of domain names is encoded using the technique described in
>    section 3.1 of RFC 1035 [10].  A domain name, or list of domain
>    names, in DHCP MUST NOT be stored in compressed form, as described in
>    section 4.1.4 of RFC 1035.

(I have a fix but it doesn't use the rfc1035label package, which I guess it should.
https://github.com/mikma/insomniacslk-dhcp/commits/bugfix/fqdn-encoding)

pmazzini:
Fixes #360.

### DIFF
--- a/dhcpv6/modifiers.go
+++ b/dhcpv6/modifiers.go
@@ -32,8 +32,12 @@ func WithNetboot(d DHCPv6) {
 // WithFQDN adds a fully qualified domain name option to the packet
 func WithFQDN(flags uint8, domainname string) Modifier {
 	return func(d DHCPv6) {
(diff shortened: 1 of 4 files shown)


pmazzini:
cc @mikma @10000TB

PR #371, opened 2020-03-11
Support for custom logger in servers

tanis2000:
This adds support for a custom logger for both the server4 and server6 packages.

Fixes #370.

PR #373, opened 2020-03-28
dhcpv4: server4: Allow the use of an external logging function

purpleidea:
This allows you to specify a logging function to use when embedding the
server in an existing project that has logging facilities. Without this,
this server will log to the console directly, instead of logging to a
controlled function that can add some context around the message.

HTH and thanks for your work!


Natolumin:
This is a duplicate of #371, but I appreciate the simpler solution of this one. I think we're basically ready to merge #371, so feel free to chime in that PR if you have something to add

purpleidea:
@Natolumin Thanks for the pointer, I didn't see it before thanks!

I think this is all that is required, but it's all up to you. I left my reasoning there.

Cheers

pmazzini:
Closing after merging #371

PR #375, opened 2020-03-29
dhcpv4: server4: Add a simple logging implementation

purpleidea:
This adds a simple implementation of the logging struct so that you
don't need to implement the complex interface if you don't need it.


purpleidea:
I think that the solution in https://github.com/insomniacslk/dhcp/pull/371/files was over-engineered, but in any case, I expect most people like myself just want simple logging. This patch adds a simple implementation so that you can easily "get logging" by passing in your own printf function straight away.

This is meant as an alternative to https://github.com/insomniacslk/dhcp/pull/373 which wasn't accepted.

Cheers!

pmazzini:
Can you sign the commit?

purpleidea:
@pmazzini You want to merge this now?

PR #379, opened 2020-04-10
Enhance close mechnism for dhcpv4/6 server

### LINKED ISSUE #376: server4/server6 doesn't closed properly with Close() function
Hi,

I am using server4/6 in goroutine(thread) mode, and I may close server4/server6 before exit my main program and do some other task. The Serve() function will return error after Close().

Error reading from packet conn: read udp 0.0.0.0:67: use of closed network connection

It shall be caused by the for loop still there without exit after the server4/6 closed by Close() function, so when try to read the close conn, above error message get printed and returned.

I suggest to enhance the Serve() and Close() function, if Close() is called, the for loop can be stopped before conn get closed.

My code as the following logic:
thread 1:

	go func() {
		s.errors <- s.Server4.Serve()
	}()
	err := <-s.errors
	if s.Server6 != nil {
		s.Server6.Close()
	}
	if s.Server4 != nil {
		s.Server4.Close()
	}
	return err

And I may send signal to stop the thread, and the error message happened after the stop signal sent:

    func (s *Server) Stop() {
	    select {
	    case s.errors <- nil:
	    default:
	    }
    }

guo1017138:
Close #376 

Signed-off-by: Guo, Larry (NSB - CN/Qingdao) <larry.guo@nokia-sbell.com>

### DIFF
--- a/dhcpv4/server4/server.go
+++ b/dhcpv4/server4/server.go
@@ -67,18 +67,22 @@ type Handler func(conn net.PacketConn, peer net.Addr, m *dhcpv4.DHCPv4)
 
 // Server represents a DHCPv4 server object
 type Server struct {
-	conn    net.PacketConn
-	Handler Handler
-	logger  Logger
+	conn      net.PacketConn
+	Handler   Handler
+	logger    Logger
+	closeFlag bool
 }
 
 // Serve serves requests.
-func (s *Server) Serve() error {
+func (s *Server) Serve() (err error) {
 	s.logger.Printf("Server listening on %s", s.conn.LocalAddr())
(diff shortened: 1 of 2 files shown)


guo1017138:
Thanks for the comment. Existing Close() implemention will close net.PacketConn directly. But, the for loop is still running there until it tries to ReadFrom a closed conn. then it will get an ERROR "read udp 0.0.0.0:67: use of closed network connection" and return it to caller. 

My main program now has workaround to handle it by filter it our. However, there shall be away to prevent this error after conn closed. 

I am not an expert on threading and maybe you can suggest how to avoid that.

PR #382, opened 2020-05-29
dhcpv4: Fix some panics found by fuzzing

Natolumin:
This PR adds a go-fuzz endpoint (similar to #325) and fixes a few panics that it discovered, as well as an RFC violation. 

See individual commits for further details, but for a quick summary:
* Patch 1 and 2 are preparation, they introduce the fuzzing entrypoint and refactor a test for ease-of-use
* Patch 3 fixes a panic with strings over their size limit in writing out packets
* Patch 4 removes a special case in ToBytes() that prevents from crafting some arbitrary packets with hlen=0. This warrants extra attention because it could break applications that do not use any of the `New()` functions to create a packet but instead craft them manually, **and** leave the `ClientHWAddr` field empty to rely on `ToBytes()` to initialize it to the `00:00:00:00:00:00` Ethernet address. Seems unlikely but worth a check
* Patch 5 fixes an oversight where some valid options (with a length of 0) would be omitted when writing out packets
* Patch 6 fixes a **panic when parsing malformed packets** which is a DOS vulnerability

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -492,13 +492,11 @@ func (d *DHCPv4) ToBytes() []byte {
 	copy(buf.WriteN(16), d.ClientHWAddr)
 
 	var sname [64]byte
-	copy(sname[:], []byte(d.ServerHostName))
-	sname[len(d.ServerHostName)] = 0
+	copy(sname[:63], []byte(d.ServerHostName))
 	buf.WriteBytes(sname[:])
 
 	var file [128]byte
-	copy(file[:], []byte(d.BootFileName))
-	file[len(d.BootFileName)] = 0
+	copy(file[:127], []byte(d.BootFileName))
 	buf.WriteBytes(file[:])
 
 	// The magic cookie.
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -144,6 +144,7 @@ func New(modifiers ...Modifier) (*DHCPv4, error) {
 	d := DHCPv4{
 		OpCode:        OpcodeBootRequest,
 		HWType:        iana.HWTypeEthernet,
+		ClientHWAddr:  make(net.HardwareAddr, 6),
 		HopCount:      0,
 		TransactionID: xid,
 		NumSeconds:    0,
@@ -476,9 +477,6 @@ func (d *DHCPv4) ToBytes() []byte {
 
 	// HwAddrLen
 	hlen := uint8(len(d.ClientHWAddr))
-	if hlen == 0 && d.HWType == iana.HWTypeEthernet {
-		hlen = 6
-	}
 	buf.Write8(hlen)
 	buf.Write8(d.HopCount)
 	buf.WriteBytes(d.TransactionID[:])
--- a/dhcpv4/options.go
+++ b/dhcpv4/options.go
@@ -185,12 +185,20 @@ func (o Options) Marshal(b *uio.Lexer) {
 		code := uint8(c)
 		// Even if the End option is in there, don't marshal it until
 		// the end.
-		if code == optEnd {
+		// Don't write padding either, since the options are sorted
+		// it would always be written first which isn't useful
+		if code == optEnd || code == optPad {
 			continue
 		}
 
 		data := o[code]
 
+		// Ensure even 0-length options are written out
+		if len(data) == 0 {
+			b.Write8(code)
+			b.Write8(0)
+			continue
+		}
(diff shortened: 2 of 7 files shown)


dhcpv4/options.go
```diff
@@ -191,6 +191,12 @@ func (o Options) Marshal(b *uio.Lexer) {
 
 		data := o[code]
 
+		// Ensure even 0-length options are written out
+		if len(data) == 0 {
+			b.Write8(code)
+			b.Write8(0)
```
hugelgupf, line 197: 0 length options like Pad do not get a length field.

hugelgupf: What is option 224?

Natolumin: RFC 2132 has:
> Any options defined subsequent to this document MUST contain a length octet even if the length is fixed or zero.

Padding is a special case (also with End, the only 0-length option defined *before* RFC2132), I don't remember if it is handled differently in the code. I'll check

For option 224, it's in the private use area, just used here for the test. Another possibility would be option 80 rapid commit, which *does* have a length field per [RFC4039§4](https://tools.ietf.org/html/rfc4039#section-4)

Natolumin: Handled Pad similarly to End, by ignoring it; since we don't store the order of options and sort them by code anyway, padding doesn't make sense as it could be moved around anywhere.

dhcpv4/dhcpv4.go
```diff
@@ -493,12 +491,20 @@ func (d *DHCPv4) ToBytes() []byte {
 
 	var sname [64]byte
 	copy(sname[:], []byte(d.ServerHostName))
-	sname[len(d.ServerHostName)] = 0
+	if len(d.ServerHostName) < len(sname) {
+		sname[len(d.ServerHostName)] = 0
+	} else {
+		sname[len(sname)-1] = 0
```
hugelgupf, line 497: actually, you could simplify this whole thing to

sname[len(sname)-1] = 0

sname is pre-initialized to 0 values. the only case where you have to append a 0 to null-terminate a string is if d.ServerHostName > 64.

dhcpv4/dhcpv4.go
```diff
@@ -493,12 +491,20 @@ func (d *DHCPv4) ToBytes() []byte {
...
+	}
 	buf.WriteBytes(sname[:])
 
 	var file [128]byte
 	copy(file[:], []byte(d.BootFileName))
-	file[len(d.BootFileName)] = 0
+	if len(d.BootFileName) < len(file) {
+		file[len(d.BootFileName)] = 0
+	} else {
+		file[len(file)-1] = 0
```
hugelgupf, line 506: ditto

dhcpv4/options.go
hugelgupf: ok makes sense. thanks

dhcpv4/dhcpv4.go
```diff
@@ -144,6 +144,7 @@ func New(modifiers ...Modifier) (*DHCPv4, error) {
 	d := DHCPv4{
 		OpCode:        OpcodeBootRequest,
 		HWType:        iana.HWTypeEthernet,
+		ClientHWAddr:  make(net.HardwareAddr, 6),
```
hugelgupf, line 147: not all hwaddrs are going to have a length of 6. what is this solving?

Natolumin: This is the whole commit, which explains the issue in a bit more detail; basically I think that this is a better tradeoff than special-casing it in ToBytes: https://github.com/insomniacslk/dhcp/pull/382/commits/6ce6e224fca409ec3ad6ae8e66795ae3ef2a3e98

New() initializes HWType to Ethernet, for which the length is 6, so at this line, the hwaddr does have a length of 6.
The core issue here is that there are 2 fields that only make sense together, but they can be defined separately (and by default, one of them is undefined), so we resort to hacky special case to detect the default value of the packet and workaround the fact that the default value is broken. 

The best fix would be a generic solution for figuring out the length of the ClientHWAddr based on HWType, but I couldn't find one. I'm not even sure all the public ones are documented or fixed length, and it'd mean we still have no solution for undeclared/non-public HWTypes

So the next idea I had was to at least initialize to a reasonable value and let the user handle the rest if they want to change the HWType and the length.
If you have a better solution to handle this, I also don't love this current solution so I'd be happy to receive some ideas there.

dhcpv4/dhcpv4.go
Natolumin: I think I got a slightly simpler idea, relying on the fact that the buffer is 0-initialized. Thanks for reminding me :)

PR #386, opened 2020-06-18
add lease&release support for nclient4

hujun-open:
This PR adds lease & release support for nclient4, main contents are:

- added DHCPv4ClientLease struct, it could be used to apply the lease (Client.ApplyLease()) or release the lease (Client.Release())
- Leased is obtained after a full DORA exchange via Client.RequestSavingLease()
- Obtained lease could be applied or unapplied via Client.ApplyLease(), which calls a handler function; a default handler is used if a custom one is not provided via WithApplyLeaseHandler(); the default handler just add/remove assigned address to/from the binding interface
- add Client.clientIdOptions, which is a dhcpv4.OptionCodeList, includes all DHCP option that server uses to identify a client, which is common practice that DHCP server uses certain option (along with MAC address) to identify the client, e.g. option82/remote-id; Client.SendAndRead() now checks if the DHCP request packet contains all options required by clientIdOptions
- Client.Release () unicast a DHCPv4 release message to server


dhcpv4/nclient4/client.go
```diff
@@ -162,6 +162,18 @@ type Client struct {
...
 	pending map[dhcpv4.TransactionID]*pendingCh
+
+	//clientIdOptions is a list of DHCPv4 option code that DHCP server used to
+	//identify client other than the HWAddress,
+	//like client-id, option82/remote-id..etc
+	clientIDOptions dhcpv4.OptionCodeList
+
+	//lease info after DORA, nil before DORA
+	lease *DHCPv4ClientLease
+	//the handnler function for applying lease
```
pmazzini, line 173: s/handnler/handler/

pmazzini:
Thanks for the PR!

dhcpv4/nclient4/client.go
```diff
@@ -162,6 +162,18 @@ type Client struct {
 	// TransactionID. receiveLoop uses this map to determine which channel
 	// to send a new DHCP message to.
 	pending map[dhcpv4.TransactionID]*pendingCh
+
+	//clientIdOptions is a list of DHCPv4 option code that DHCP server used to
+	//identify client other than the HWAddress,
+	//like client-id, option82/remote-id..etc
+	clientIDOptions dhcpv4.OptionCodeList
```
pmazzini, line 169: The only option that would make sense to have here is the client-id one (61). I don't think it makes sense to have the relay agent info here (82). Also, the proper parsing of option 61 is not implemented in the library. Should it be implemented before using it here?

dhcpv4/nclient4/example_lease_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/insomniacslk/dhcp/dhcpv4/nclient4"
+)
+
+func Example_dHCPv4ClientLease() {
+	ifname := "eth0"
+	remoteID := "client-1"
+	var idoptlist dhcpv4.OptionCodeList
+	//specify option82 is part of client identification used by DHCPv4 server
```
pmazzini, line 17: As above, I wouldn't use the the relay agent info in the client code.

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,227 @@
...
+package nclient4
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/vishvananda/netlink"
```
pmazzini, line 12: We moved away from `vishvananda/netlink`, check #271. Can you use `jsimonetti/rtnetlink`?

dhcpv4/nclient4/client.go
hujun-open: ok, will fix the typo

dhcpv4/nclient4/client.go
hujun-open: using option82 (circuit-id/remote-id) to identify client is a common practice by broadband service provider (as described RFC3046 section 1); the idea of clientIDOptions is to have a flexible way that could work with many different DHCP server requirements.

I could also take a look on option61 parsing

dhcpv4/nclient4/example_lease_test.go
hujun-open: see my comment above

dhcpv4/nclient4/lease.go
hujun-open: ok, will take a look

dhcpv4/nclient4/client.go
pmazzini: Sure, it can be used by the server but it shouldn't be inserted by the client.

As per the RFC:

> The new option is called the Relay Agent Information option and is inserted by the DHCP relay agent

So this is not a normal client behaviour.

hujun-open: yes, a DHCP client on end user device will not insert option82; however if someone want to use this library to write a testing client to test DHCP server that requires option82 without any access node, then it is useful in such case (actually that's my use case)
plus clientIDOptions is not hard-coded to option82 or any specific option; it could be empty  if it is not needed;

pmazzini: This feels like you are tuning the client to your particular needs more than implementing standard and RFC compliant things.

Can you not use simply use a modifier?

dhcpv4/nclient4/client.go
```diff
@@ -521,6 +537,15 @@ var errDeadlineExceeded = errors.New("INTERNAL ERROR: deadline exceeded")
 // ClientHWAddr is returned.
 func (c *Client) SendAndRead(ctx context.Context, dest *net.UDPAddr, p *dhcpv4.DHCPv4, match Matcher) (*dhcpv4.DHCPv4, error) {
 	var response *dhcpv4.DHCPv4
+	//check if the request packet has all options required by c.clientIdOptions
```
pmazzini, line 540: I would remove this check

dhcpv4/nclient4/client.go
hujun-open: the problem of using a modifier is when client need to send DHCP release, it needs to use same ID option as used during DORA, otherwise server won't be able to identify the lease to release; and in case when persistent lease storage is needed, e.g. save lease to a file so that it could be read later on to create the client and send the release, the ID option need to be saved as part of lease; 
maybe it is not a good idea of using option82/remote-id in my example, but clientIDOptions is not really only for option82;

dhcpv4/nclient4/client.go
hujun-open: This check is more for preventing user to shoot itself in the foot, I agree it is not mandatory, I could remove it

dhcpv4/nclient4/client.go
```diff
@@ -162,6 +162,18 @@ type Client struct {
...
 	// to send a new DHCP message to.
 	pending map[dhcpv4.TransactionID]*pendingCh
+
+	//clientIdOptions is a list of DHCPv4 option code that DHCP server used to
+	//identify client other than the HWAddress,
+	//like client-id, option82/remote-id..etc
+	clientIDOptions dhcpv4.OptionCodeList
+
+	//lease info after DORA, nil before DORA
+	lease *DHCPv4ClientLease
```
hugelgupf, line 172: I don't think the current lease should be stored in this client, and I don't think this client should be concerned with applying the lease information to the host system.

I'd rather a user be able to make the choice of what to store and where for themselves, and that a user be able to make the choice of what configuration to apply to the host system and how, because I have a feeling it's going to be different for some.

I think FB tends to use the netboot package in this library, and we tend to use https://github.com/u-root/u-root/tree/master/pkg/dhclient, but either way, I don't think it belongs in nclient4.

reasoning being I'd rather not support this for all users if we cannot get every edge case correct, which neither netboot nor u-root's pkg/dhclient nor this currently do. and if we do, then it'd definitely be in a separate package from nclient4, just to decouple it -- i don't want u-root, for example, to be saddled with this dependency when it already has its own implementation of this.

My overall recommendation would be to keep the RenewalTime and RebindingTime -- that's really useful -- and the Release function. But let users pass in the lease they want released, rather than using one stored in this client:

```go
func (c *Client) Release(ack *DHCPv4) error {

}
```

That `ack` should have all the information needed to assemble a `Release`. This would be similar to the way that `Advertise` and `Request` interact in this client already.

dhcpv4/nclient4/client.go
hugelgupf: I think this is something for callers to do, not for this library, too.

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,227 @@
...
+		return
+	}
+
+	ack, err = c.SendAndRead(ctx, c.serverAddr, request, nil)
+	if err != nil {
+		err = fmt.Errorf("got an error while processing the request: %w", err)
+		return
+	}
+	//save lease
+	c.lease = &DHCPv4ClientLease{}
```
hugelgupf, line 135: I'm not sure I understand entirely why DHCPv4ClientLease exists. I would think that the ACK packet *is* the lease, and we end up with 

```go
type DHCPv4ClientLease struct {
  ACK *DHCPv4
  CreationTime time.Time
}
```

because I don't think the information you pulled out is all that users are going to want to use from the ACK packet.

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,227 @@
...
+	//update server address
+	c.serverAddr = &(net.UDPAddr{IP: ack.ServerIdentifier(), Port: 67})
+	err = c.ApplyLease(true)
+	return
+}
+
+//Release send DHCPv4 release messsage to server.
+//release is sent as unicast per RFC2131, section 4.4.4.
+//The lease need to be applied with c.ApplyLease(true) first before calling Release.
+func (c *Client) Release() error {
```
hugelgupf, line 174: I think if this library is to remain relatively neutral about the things I said before, I'd let the user supply the lease they want to release here:

```go
func (c *Client) Release(ack *DHCPv4) error {
  if ack == nil {
    return fmt.Errorf("no lease to release")
  }
  req, err := dhcpv4.New()
  if err != nil {
    ...
  }
  uniconn, err := net.DialUDP("udp4", &net.UDPAddr{
    IP: ack.YourIPAddr,
    Port: 68,
  }, &net.UDPAddr{
    IP: ack.ServerIdentifier(),
    Port: 67,
  })
  ...
}
```

hugelgupf: Shouldn't we also wait for an ACK response from the server?

dhcpv4/nclient4/client.go
hujun-open: ack does not necessarily have all the information needed to build release, as my comments above, in case DHCP sever uses DHCP option to identify a DHCP client, for example, client-id option, ack will not have the client-id option;
I agree that this library shouldn't force how user apply the lease, that's why there is WithApplyLeaseHandler(h func(DHCPv4ClientLease, bool) error) for user to provide their own handler to however applying lease as they like;  the default handler (defaultLeaseApplyHandler)  only add assigned address to the interface, so that release message could use it as the source address, and if user doesn't want that behavoir, it could simply add and empty handler via  WithApplyLeaseHandler

dhcpv4/nclient4/lease.go
hujun-open: according to RFC2131, there is no response for DHCP release:
- section 3.1 the flowchart doesn't show any response to release
- section 4.3.4, no mention of response

dhcpv4/nclient4/lease.go
hujun-open: yes, most of part of lease come from ACK, beside the the ID options used during DORA, so maybe I could change it to sth like following:

```
type DHCPv4ClientLease struct {
  ACK *DHCPv4
  CreationTime time.Time
  IDOptions      dhcpv4.Options
}
```

dhcpv4/nclient4/client.go
hugelgupf: I think part of not forcing it, though, means not forcing other users of the library to take on the netlink dependency. E.g. the two different users of this library use different netlink libraries to achieve their goals, and both users care a lot about binary size requirements. We don't want both libraries in our dependencies, we want to choose. Unfortunately I don't trust Go's dead code elimination enough to be sure it'll be eliminated when it's not used.

hugelgupf: I'm open to committing something like u-root's pkg/dhclient in this library as well, for convenience, if it's maintainable. Some package that's higher-level than nclient4, but not in nclient4 imho.

dhcpv4/nclient4/lease.go
hugelgupf: I'd appreciate that, yeah! I'm not sure I fully get the IDOptions thing yet. What ID options are these, specifically, that aren't in the ack and aren't known to this library already?

dhcpv4/nclient4/lease.go
hugelgupf: you're right, sorry about that. I'll modify my request to be something like:

```go
func (c *Client) Release (lease *DHCPv4ClientLease) error {
  ...
}
```

I'd still rather the client be able to release any lease, not just the one stored in it (which is why I'd rather it not be stored in the Client object either).

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		return err
+	}
+	c.logger.PrintMessage("sent message:", req)
+	return c.ApplyLease(false)
+}
+
+//NewWithLease return a Client with populated lease.
+//this function could be used to release a saved lease.
+func NewWithLease(clease DHCPv4ClientLease, opts ...ClientOpt) (*Client, error) {
```
hugelgupf, line 210: This actually sort of illustrates why I'd rather the lease not be stored in the *Client object -- now you have to construct a Client with a lease to be able to release it.

dhcpv4/nclient4/lease.go
hujun-open: rfc2131, section 4.2, specifies that client-id option61 could be used to identify client, quote below
`
The client MAY choose to explicitly provide
   the identifier through the 'client identifier' option.  If the client
   supplies a 'client identifier', the client MUST use the same 'client
   identifier' in all subsequent messages, and the server MUST use that
   identifier to identify the client.`
then in section 4.3.1's table, client-id is marked as "MUST NOT" for DHCPACK, meaning ACK will not include client-id; so in such case, client need to remember the client-id used during DORA, and use the same value in release, otherwise server won't be able to identify the lease that release targets.  

So the idea of IDOptions is meant to be a generic way for option like client-id, for example a DHCP server might require using class-id(option60) + client-id to identify client, then IDOptions will include option 60 and 61;

dhcpv4/nclient4/client.go
hujun-open: for netlink library, I could switch to  jsimonetti/rtnetlink, it shouldn't be difficult if that's your only concern;
however if you think nclient4 should only be low-level (e.g. remain stateless, no need for lease/release), then maybe then I just have to create another client ...

dhcpv4/nclient4/lease.go
hujun-open: no, NewWithLease is only for use case like client did DORA, save the lease to a file, and quit running, then after some time, start again, read lease from the file, create a client with saved lease, and could do renew/release with the loaded lease;
if you don't need such use case, you could simply call Client.Release() after DORA, no need to create a new client.

dhcpv4/nclient4/client.go
hugelgupf: I don't have a preference for one or the other, but we do need to satisfy requirements for *both* users of this library, which includes not having unwanted dependency. vishv/netlink is unwanted for those that use rtnetlink, and rtnetlink is unwanted for those who still use vishv/netlink. A wrapper around this client for your use case would be preferred IMHO. @pmazzini, wdyt?

dhcpv4/nclient4/lease.go
hugelgupf: Makes sense.

dhcpv4/nclient4/lease.go
hugelgupf: What if I have a client with more than one lease to more than one DHCP server?

hujun-open: current design of this PR assume only one lease per client, for multiple leases, then multiple clients need to be created; which I think is acceptable design, because there could be significant differences between leases (e.g different binding interface), so there is not much benefit to share a single client instance

hugelgupf: I was thinking more of multiple leases per interface.

As I said elsewhere, I'm happy to have a more stateful DHCP client in a different package in this library, but I'd rather keep nclient4 a little less stateful. We use this DHCP client for netbooting mostly, and in that case we don't need and don't want this state to be around, so it'd be great to have a way to separate the use cases.

hujun-open:
Based on discussion we had so far, I propose following changes to this PR
- Client struct no longer include DHCPv4ClientLease reference
- IDOptions is still part of lease
- DHCPv4ClientLease will now include following fields:
```
type DHCPv4ClientLease struct {
  BindIfName string
  ACK *DHCPv4
  CreationTime time.Time
  IDOptions      dhcpv4.Options
}
```
- remove Client.RequestSavingLease, add a new Client.RequestLease, which is just Client.Request + return a DHCPv4ClientLease, like following:
```
func (c *Client) RequestLease(ctx context.Context, modifiers ...dhcpv4.Modifier) (offer, ack *dhcpv4.DHCPv4, lease DHCPv4ClientLease,  err error) {
```
- still keep Client.clientIDOptions and WithClientIDOptions function
- remove all Lease Applying code, GetLease, NewWithLease function
- change Client.Release() to Client.Release(DHCPv4ClientLease), however this will require user of library to "manually" add the assigned address to the binding interface, because DHCP release require using assigned IP as the source IP

@hugelgupf , @pmazzini what do you think?

hugelgupf:
> Based on discussion we had so far, I propose following changes to this PR
> 
> * Client struct no longer include DHCPv4ClientLease reference
> * IDOptions is still part of lease
> * DHCPv4ClientLease will now include following fields:
> 
> ```
> type DHCPv4ClientLease struct {
>   BindIfName string
>   ACK *DHCPv4
>   CreationTime time.Time
>   IDOptions      dhcpv4.Options
> }
> ```
> 
> * remove Client.RequestSavingLease, add a new Client.RequestLease, which is just Client.Request + return a DHCPv4ClientLease, like following:
> 
> ```
> func (c *Client) RequestLease(ctx context.Context, modifiers ...dhcpv4.Modifier) (offer, ack *dhcpv4.DHCPv4, lease DHCPv4ClientLease,  err error) {
> ```
> 
> * still keep Client.clientIDOptions and WithClientIDOptions function
> * remove all Lease Applying code, GetLease, NewWithLease function
> * change Client.Release() to Client.Release(DHCPv4ClientLease), however this will require user of library to "manually" add the assigned address to the binding interface, because DHCP release require using assigned IP as the source IP
> 
> @hugelgupf , @pmazzini what do you think?

This plan SGTM.

The regular Client.Request could return a Lease instead of the ACK, I think callers would be willing to update their usage of this, no need to have a separate RequestLease.

I'd call DHCPv4ClientLease just Lease, no need for the long name. 

Thank you!

hujun-open:
ok, I have made changes as discussed, also updated the example_lease_test, only extra thing is I decided to no longer include BindIfName in Lease struct, since that's not really needed for release to work

dhcpv4/nclient4/client.go
pmazzini: > A wrapper around this client for your use case would be preferred IMHO.

Yes, I agree. Maybe even a separate package inside this repo.

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,65 @@
...
+	//This is to make sure use same client identification options used during
+	//DORA, so that DHCP server could identify the required lease
+	req.Options = lease.IDOptions
+
+	req.UpdateOption(dhcpv4.OptMessageType(dhcpv4.MessageTypeRelease))
+	req.ClientHWAddr = lease.ACK.ClientHWAddr
+	req.ClientIPAddr = lease.ACK.YourIPAddr
+	req.UpdateOption(dhcpv4.OptServerIdentifier(lease.ACK.ServerIPAddr))
+	req.SetUnicast()
+	luaddr, err := net.ResolveUDPAddr("udp4", fmt.Sprintf("%v:%v", lease.ACK.YourIPAddr, 68))
```
hugelgupf, line 50: just curious, i haven't looked it up: does the Release have to be unicast? Also, the existing conn does have the ability to unicast, but would be assembled with the src address as 0.0.0.0.

dhcpv4/nclient4/example_lease_test.go
```diff
@@ -0,0 +1,65 @@
...
+	remoteidsubopt := dhcpv4.OptGeneric(dhcpv4.AgentRemoteIDSubOption, []byte(remoteid))
+	option82 := dhcpv4.OptRelayAgentInfo(remoteidsubopt)
+	_, lease, err := clnt.Request(context.Background(), dhcpv4.WithOption(option82))
+	if err != nil {
+		log.Fatal(err)
+	}
+	//print the lease
+	log.Printf("Got lease:\n%+v", lease)
+	//apply the lease
+	applyLease(lease, ifname)
```
hugelgupf, line 57: the code is looking great to me, but I'm not sure how we're gonna run this test automatically. This seems like it'll only work if we run this test somewhere where we're root and we have an unconfigured interface? That's pretty unlikely to run on something like CircleCI or Travis

dhcpv4/nclient4/lease.go
hujun-open: according to RFC2131 section 4.4.4:`The client unicasts DHCPRELEASE messages to the server.`
regarding the source IP address, there is no specific requirement in the RFC, however the DHCP server (a commercial one) I have been testing rejected an unicast release with 0.0.0.0 as the source IP; which kind of make sense, because after DORA, the client suppose to use the assigned address as source IP;
then today, I tried with ISC Kea DHCP server, it actually accept both 0.0.0.0 and assigned IP as source, so now I decide to change the Release() to use client's conn directly, if the server require using the assigned ip as source, then user need to use nclient4.WithUnicast to create client;

dhcpv4/nclient4/example_lease_test.go
hujun-open: since now I have decided no longer mandate user to add assigned address to the interface, and Release() use Client.conn directly, I could try to write some test cases using socketpair, just like the cases in client_test.go;

hujun-open:
changes committed, I have also added a lease_test.go, which contains some test cases;

dhcpv4/nclient4/client.go
```diff
@@ -447,12 +453,19 @@ func (c *Client) Request(ctx context.Context, modifiers ...dhcpv4.Modifier) (off
...
 	}
 
-	ack, err = c.SendAndRead(ctx, c.serverAddr, request, nil)
+	ack, err := c.SendAndRead(ctx, c.serverAddr, request, nil)
 	if err != nil {
 		err = fmt.Errorf("got an error while processing the request: %w", err)
 		return
 	}
-
+	lease = &Lease{}
```
hugelgupf, line 461: put the offer in the Lease too? might as well

dhcpv4/nclient4/client.go
pmazzini: How about renaming this to `copyOptions`? That is what this is for, isn't it?And it is not specific to ID options.

dhcpv4/nclient4/client.go
hujun-open: ok, I have added offer in the Lease

hujun-open:
> 
> 
> this is looking really good! sorry, i totally failed to check in on this PR

I have added offer to Lease, could you take a look? @hugelgupf

dhcpv4/nclient4/client.go
```diff
@@ -432,7 +438,8 @@ func (c *Client) DiscoverOffer(ctx context.Context, modifiers ...dhcpv4.Modifier
 // Request completes the 4-way Discover-Offer-Request-Ack handshake.
 //
 // Note that modifiers will be applied *both* to Discover and Request packets.
-func (c *Client) Request(ctx context.Context, modifiers ...dhcpv4.Modifier) (offer, ack *dhcpv4.DHCPv4, err error) {
+func (c *Client) Request(ctx context.Context, modifiers ...dhcpv4.Modifier) (lease *Lease, err error) {
+	var offer *dhcpv4.DHCPv4
```
pmazzini, line 442: NIT: can you replace this var with `:=` in the next line?

dhcpv4/nclient4/example_lease_test.go
```diff
@@ -0,0 +1,37 @@
+package nclient4_test
```
pmazzini, line 1: Maye `package nclient4` to keep it consistent with the rest of the tests?

pmazzini:
Things looks good to me except for the code added to copy option82 around. I'll give it a deeper look.

dhcpv4/nclient4/client.go
```diff
@@ -447,12 +454,20 @@ func (c *Client) Request(ctx context.Context, modifiers ...dhcpv4.Modifier) (off
...
 	}
-
+	lease = &Lease{}
+	lease.ACK = ack
+	lease.Offer = offer
+	lease.CreationTime = time.Now()
+	lease.IDOptions = dhcpv4.Options{}
+	for _, optioncode := range c.clientIDOptions {
+		v := request.Options.Get(optioncode)
+		lease.IDOptions.Update(dhcpv4.OptGeneric(optioncode, v))
```
pmazzini, line 469: Are you copying this so that it can later be applied to the release? How about copying the modifiers instead?

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,60 @@
...
+
+//Release send DHCPv4 release messsage to server, based on specified lease.
+//release is sent as unicast per RFC2131, section 4.4.4.
+//Note: some DHCP server requries of using assigned IP address as source IP,
+//use nclient4.WithUnicast to create client for such case.
+func (c *Client) Release(lease *Lease) error {
+	if lease == nil {
+		return fmt.Errorf("lease is nil")
+	}
+	req, err := dhcpv4.New()
```
pmazzini, line 39: How about copying the modifiers and applying them here instead of defining a list of IDOptions?

dhcpv4/nclient4/example_lease_test.go
hujun-open: example_lease_test.go is intended to be an example in go doc, that's why the package has "_test" suffix, see  ["whole file example" of this blog](https://blog.golang.org/examples)

dhcpv4/nclient4/client.go
hujun-open: will update it

dhcpv4/nclient4/example_lease_test.go
pmazzini: Got it. Thanks for the info.

dhcpv4/nclient4/client.go
hujun-open: yes, that's purpose; however in order to make this more generic, for example if user wants to save lease to a file, it is better to copy the option itself, since modifier is just a function;

dhcpv4/nclient4/lease.go
hujun-open: see my comment above

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return err
+	}
+	//This is to make sure use same client identification options used during
+	//DORA, so that DHCP server could identify the required lease
+	req.Options = lease.IDOptions
+
+	req.UpdateOption(dhcpv4.OptMessageType(dhcpv4.MessageTypeRelease))
+	req.ClientHWAddr = lease.ACK.ClientHWAddr
```
pmazzini, line 48: Can the modifiers `WithYourIP`, `WithHwAddr`, `WithBroadcast(false)` and `WithOptionCopied(OptionServerIdentifier)` be used when building the request to make more consistent with how other requests are generated?

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,60 @@
...
+	req.ClientIPAddr = lease.ACK.YourIPAddr
+	req.UpdateOption(dhcpv4.OptGeneric(dhcpv4.OptionServerIdentifier, lease.ACK.Options.Get(dhcpv4.OptionServerIdentifier)))
+	req.SetUnicast()
+	if err != nil {
+		return err
+	}
+	_, err = c.conn.WriteTo(req.ToBytes(), &net.UDPAddr{IP: lease.ACK.Options.Get(dhcpv4.OptionServerIdentifier), Port: 67})
```
pmazzini, line 55: Can 67 be replaced with ServerPort constant?

dhcpv4/nclient4/lease.go
hujun-open: I could use `WithYourIP`, `WithHwAddr`, `WithBroadcast(false)` and `WithMessageType`; however I can't use `WithOptionCopied(OptionServerIdentifier)` since Lease doesn't contain any request packet, and I want Release() work just with Lease, so that is could be used to e.g. read a lease from file and use to send release message

dhcpv4/nclient4/lease.go
hujun-open: ok, will do

hujun-open:
ok, the new commit should address most of your concern beside saving modifers in Lease for the reason I gave above, pls take a look, @pmazzini

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,64 @@
...
+
+//Release send DHCPv4 release messsage to server, based on specified lease.
+//release is sent as unicast per RFC2131, section 4.4.4.
+//Note: some DHCP server requries of using assigned IP address as source IP,
+//use nclient4.WithUnicast to create client for such case.
+func (c *Client) Release(lease *Lease) error {
+	if lease == nil {
+		return fmt.Errorf("lease is nil")
+	}
+	modList := []dhcpv4.Modifier{
```
pmazzini, line 39: I think part of this should be moved so that it can be reused.

I think we should have a `NewReleaseFromACK` in `dhcpv4.go`, similar to [NewRequestFromOffer](https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/dhcpv4.go#L235)

Maybe at another time?

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,64 @@
...
+// WithClientIDOptions configures a list of DHCPv4 option code that DHCP server
+// uses to identify client, beside the MAC address.
+func WithClientIDOptions(cidl dhcpv4.OptionCodeList) ClientOpt {
+	return func(c *Client) (err error) {
+		c.clientIDOptions = cidl
+		return
+	}
+}
+
+//Release send DHCPv4 release messsage to server, based on specified lease.
```
pmazzini, line 31: NIT: add a space after //

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,64 @@
...
+//release is sent as unicast per RFC2131, section 4.4.4.
+//Note: some DHCP server requries of using assigned IP address as source IP,
+//use nclient4.WithUnicast to create client for such case.
+func (c *Client) Release(lease *Lease) error {
```
pmazzini, line 35: Should `Release` be able to take `Modifiers` similar to `Request`?

dhcpv4/nclient4/example_lease_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/insomniacslk/dhcp/dhcpv4/nclient4"
+)
+
+func Example() {
+	ifname := "enp0s10"
+	clntid := "client-1"
+	var idoptlist dhcpv4.OptionCodeList
+	//specify option61 is part of client identification used by DHCPv4 server
+	idoptlist.Add(dhcpv4.OptionClientIdentifier)
+	clntOptions := []nclient4.ClientOpt{nclient4.WithClientIDOptions(idoptlist), nclient4.WithDebugLogger()}
```
pmazzini, line 17: I do not really like this API and how some options are copied around.

This is how it looks like now:

```
// it tells the client to copy some options to the lease struct
clntOptions := []nclient4.ClientOpt{nclient4.WithClientIDOptions(idoptlist)}

// create the option you are interested in
option61 := dhcpv4.OptClientIdentifier([]byte(clntid))

// tell the Request to add the option you are interested in
lease, err := clnt.Request(context.Background(), dhcpv4.WithOption(option61))
```

Can it be changed to:

```
// create the option you are interested in
option61 := dhcpv4.OptClientIdentifier([]byte(clntid))

// tell the Request to add the option you are interested in
lease, err := clnt.Request(context.Background(), dhcpv4.WithOption(option61))

// tell the Release to add the option you are interested in
clnt.Release(lease, dhcpv4.WithOption(option61))
```

I think it will be more consistent.

I am not going to block it though if @hugelgupf was ok with the previous design

hugelgupf: I had a similar concern, but tbh not the time to follow up. Either is fine with me, this could be nicer.

I haven't fully followed where this ID option first comes into play. Perhaps that could also be stored in the Lease?

hugelgupf: (I'm guessing the expectation is that it be the same as in the original Request or something?)

dhcpv4/nclient4/lease.go
hujun-open: ok, I could move this to NewReleaseFromACK

dhcpv4/nclient4/lease.go
hujun-open: will do

dhcpv4/nclient4/lease.go
hujun-open: I could do that

dhcpv4/nclient4/example_lease_test.go
hujun-open: The reason behind Lease.IDOptions is in some cases DHCP server uses certain DHCP options to identify a lease, for example, RFC2131 allows DHCP server use "Client Identifier" option61 to identify a lease (just like in example_lease_test.go), so when client need to release lease, it needs to use same identifier so that server could locate the lease, but option61 is not included in ACK/Offer, which means it needs to be part of lease, that's why Lease has this IDOptions;  I know there is DHCP server in some other use cases use different option like vendor class + client id;
hopefully this clarify the use case

pmazzini: As a user of the client library you need to create any additional option and pass it with a modifier to `Request`. What is them problem of passing it the same way to `Release`?

hujun-open: but in following case, the client program might not aware the identification options if they are not part of lease:
1. client program run 1st time, doing DORA for 1000 with 1000 different option61, get 1000 leases, save them to a file, and then quit
2. later on, client program need to run again to release these 1000 leases, so it will read leases from file, but at this point client  program doesn't know what or even if option61 need to be included in Release unless option61 is part of lease

pmazzini: I think an abstraction on top of this client should handle the persistence of the lease in the way you want.

You will be creating the additional options and you can persist them.

This is not about option61, which is probably not what you are using because the library does not implement it.

Or even have a configuration file with all the additional options that are needed. I would leave all of this to an abstraction on top of this client.

Other than how these options copied around, the diff looks good to me.

hujun-open:
ok, I have now removed IDOptions, along some other updates to address latest comments from @pmazzini

dhcpv4/nclient4/lease.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// NewReleaseFromLease creates a DHCPv4 Release message from the lease.
+// default Release message without any Modifer is created as following:
+//  - option Message Type is Release
+//  - ClientIP is set to lease.ACK.YourIPAddr
+//  - ClientHWAddr is set to lease.ACK.ClientHWAddr
+//  - Unicast
+//  - option Server Identifier is set to ServerIdentifier of lease.ACK
+func NewReleaseFromLease(lease *Lease, modifiers ...dhcpv4.Modifier) (*dhcpv4.DHCPv4, error) {
```
pmazzini, line 28: I was suggesting to add a `NewReleaseFromACK` in the main dhcp lib similar to `NewRequestFromOffer`.
Anyway, this lgtm

hujun-open: I knew, and then I realized that if I put `NewReleaseFromACK` in dhcpv4 pkg, then dhcpv4 need to import nclient4 pkg since Lease is defined in nclient4, then it is import loop since nclient4 also import dhcpv4; to solve that, I need to move Lease to dhcpv4, then I feel it is a bit odd for dhcpv4 to include Lease since dhcpv4 is supposed to be protocol level only pkg while Lease is really a client thing

pmazzini: You do not need to import the `lease` struct to have a base `NewReleaseFromACK` or `NewReleaseFromRequest`.

```
func NewReleaseFromACK(ack *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) {
```

And then you can use that function here.

hujun-open: ah, sorry, I just realized you were talking abut ACK not Lease, I don't know what happened, but somehow my mind automatically translated NewReleaseFromACK to NewReleaseFromLease ... 
I could actually change it to NewReleaseFromACK and put it in dhcpv4

dhcpv4/dhcpv4.go
```diff
@@ -758,3 +790,20 @@ func (d *DHCPv4) VIVC() VIVCIdentifiers {
...
 }
+
+// NewReleaseFromACK creates a DHCPv4 Release message from ACK.
+// default Release message without any Modifer is created as following:
+//  - option Message Type is Release
+//  - ClientIP is set to ack.YourIPAddr
+//  - ClientHWAddr is set to ack.ClientHWAddr
+//  - Unicast
+//  - option Server Identifier is set to ack's ServerIdentifier
+func NewReleaseFromACK(ack *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) {
```
pmazzini, line 801: NIT: I would move it near the other New*From* functions.

dhcpv4/dhcpv4.go
```diff
@@ -271,6 +271,23 @@ func NewReplyFromRequest(request *DHCPv4, modifiers ...Modifier) (*DHCPv4, error
...
+	return New(PrependModifiers(modifiers,
+		WithMessageType(MessageTypeRelease),
+		WithClientIP(ack.YourIPAddr),
+		WithHwAddr(ack.ClientHWAddr),
+		WithBroadcast(false),
+		WithOption(OptServerIdentifier(ack.ServerIdentifier())),
```
pmazzini, line 287: NIT: maybe WithOptionCopied instead of WithOption

PR #388, opened 2020-07-11
nclient6: optional dropped packet logging

hugelgupf:
We're chasing a Heisenbug. 

cc @ganshun (this was Ian's idea)


GanShun:
That's cool! hope it helps

PR #390, opened 2020-07-22
Go modules

ameshkov:
In order to do this, I've added main.go and main_linux.go that import all other packages. This way "go build" or "go test ./..." trigger go.mod/sum update.

Supporting Go modules would make it much easier to contribute for someone like me who does not like to store all code in GOPATH.


AkihiroSuda:
@insomniacslk PTAL? Also I'd like to see tagged releases

hugelgupf:
I did some of this in #399, with some other cleanups. The main.go is, imho, unnecessary. `go mod tidy` can be used to trigger a go mod update from the main directory without having a main.go.

hugelgupf:
That and #400 should allow us to deprecate the `dhcpv4/async` and `dhcpv6/async` clients, which I'd like to get rid of before tagging a v1.0.0 -- I don't think we should guarantee compatibility with them.

PR #391, opened 2020-08-05
dhcpv6: Handle IA_TA options

Natolumin:
This creates support for IA_TA options, based on and reusing the blocks from IA_NA, to which it is extremely similar (It's the same, but with T1 and T2 removed)

### DIFF
--- a/dhcpv6/option_temporaryaddress.go
+++ b/dhcpv6/option_temporaryaddress.go
@@ -1,0 +1,48 @@
+package dhcpv6
+
+import (
+	"fmt"
+
+	"github.com/u-root/u-root/pkg/uio"
+)
+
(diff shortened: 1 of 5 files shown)


dhcpv6/option_temporaryaddress.go
```diff
@@ -0,0 +1,47 @@
...
+// option.
+//
+// This module defines the OptIATA structure.
+// https://www.ietf.org/rfc/rfc8415.txt
+type OptIATA struct {
+	IaId    [4]byte
+	Options IdentityOptions
+}
+
+func (op *OptIATA) Code() OptionCode {
```
pmazzini, line 19: Maybe add a comment to make sticker-ci happy

PR #394, opened 2020-08-11
Option: Information Refresh Time https://tools.ietf.org/html/rfc8415

mcphailtom:
Added structure, parsing and convenience fuctions for OptInformationRefreshTime (32) https://tools.ietf.org/html/rfc8415

### DIFF
--- a/dhcpv6/dhcpv6message.go
+++ b/dhcpv6/dhcpv6message.go
@@ -248,6 +248,21 @@ func (mo MessageOptions) ElapsedTime() time.Duration {
 	return 0
 }
 
+// InformationRefreshTime returns the Information Refresh Time option
+// as defined by RFC 815 Section 21.23.
+//
+// InformationRefreshTime returns the provided default if no option is present.
+func (mo MessageOptions) InformationRefreshTime(def time.Duration) time.Duration {
+	opt := mo.Options.GetOne(OptionInformationRefreshTime)
+	if opt == nil {
+		return def
+	}
+	if t, ok := opt.(*optInformationRefreshTime); ok {
+		return t.InformationRefreshtime
+	}
+	return def
+}
+
 // FQDN returns the FQDN option as defined by RFC 4704.
 func (mo MessageOptions) FQDN() *OptFQDN {
 	opt := mo.Options.GetOne(OptionFQDN)
--- a/dhcpv6/option_informationrefreshtime.go
+++ b/dhcpv6/option_informationrefreshtime.go
@@ -1,0 +1,48 @@
+package dhcpv6
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/u-root/u-root/pkg/uio"
+)
+
+// OptInformationRefreshTime implements OptionInformationRefreshTime option.
+// https://tools.ietf.org/html/rfc8415#section-21.23
+func OptInformationRefreshTime(irt time.Duration) *optInformationRefreshTime {
+	return &optInformationRefreshTime{irt}
+}
+
+// optInformationRefreshTime represents an OptionInformationRefreshTime.
+type optInformationRefreshTime struct {
+	InformationRefreshtime time.Duration
+}
+
+// Code returns the option's code
+func (op *optInformationRefreshTime) Code() OptionCode {
+	return OptionInformationRefreshTime
+}
+
+// ToBytes serializes the option and returns it as a sequence of bytes
+func (op *optInformationRefreshTime) ToBytes() []byte {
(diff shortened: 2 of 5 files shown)


dhcpv6/option_informationrefreshtime.go
```diff
@@ -0,0 +1,46 @@
+package dhcpv6
+
+import (
+	"fmt"
+	"github.com/u-root/u-root/pkg/uio"
```
pmazzini, line 5: NIT: use a separate blocks for external dependency

https://github.com/golang/go/wiki/CodeReviewComments#imports

dhcpv6/dhcpv6message.go
```diff
@@ -248,6 +248,21 @@ func (mo MessageOptions) ElapsedTime() time.Duration {
 	return 0
 }
 
+// InformationRefreshTime returns the Information Refresh Time option
+// as defined by RFC 815 Section 21.23.
+//
+// ElapsedTime returns a duration of 0 if the option is not present.
```
Natolumin, line 254: This makes it impossible to differentiate packets with the explicit "0" value (that's not valid per the rfc, but it's still representable on the wire so we should be able to handle it) from packets where the option isn't supplied.

Can you make it either a pointer or return `(Duration, error)` or `(Duration, bool)` where the boolean indicates if it was found, similar to the interface of maps ?

Additionally, there's a copy-paster error, `ElapsedTime` -> `InformationRefreshTime`

dhcpv6/option_informationrefreshtime.go
```diff
@@ -0,0 +1,46 @@
...
+	return fmt.Sprintf("InformationRefreshTime: %v", op.InformationRefreshtime)
+}
+
+// parseOptInformationRefreshTime builds an optInformationRefreshTime structure from a sequence
+// of bytes. The input data does not include option code and length bytes.
+func parseOptInformationRefreshTime(data []byte) (*optInformationRefreshTime, error) {
+	var opt optInformationRefreshTime
+	buf := uio.NewBigEndianBuffer(data)
+
+	var irt Duration
```
Natolumin, line 42: offtopic note: Duration doesn't handle the special value of `0xffffffff` to mean infinity, which we should fix (but not in this PR)

dhcpv6/option_informationrefreshtime.go
```diff
@@ -0,0 +1,46 @@
+package dhcpv6
+
+import (
+	"fmt"
+	"github.com/u-root/u-root/pkg/uio"
+	"time"
+)
+// OptInformationRefreshTime implements OptionInformationRefreshTime option.
+// https://tools.ietf.org/html/rfc8415
```
Natolumin, line 9: nit: for a bit more accuracy
```suggestion
// https://tools.ietf.org/html/rfc8415#section-21.23
```

dhcpv6/dhcpv6message.go
mcphailtom: Great idea, apologies I got a bit lazy here (evidenced by the copy paste error) :)

pmazzini: I wonder if we should make it consistent with the v4 one in this case:

```
func (d *DHCPv4) IPAddressRenewalTime(def time.Duration) time.Duration {
```

https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/dhcpv4.go#L690

I am not sure what the best option is, just asking for opinion.

Natolumin: I don't think there is a lot of value in trying to keep the interfaces the same; the protocols are different and I still haven't seen any code that really manages to handle both transparently. The current APIs are already very different

So the question is, do we have users for this api, and what do they care about ?
 * They want to get a (valid) value, anything: then the version with a default is good
 * They want to know if the option was passed, and if yes what the value is : then we want the double-return

As far as I can tell, there are no users on github of either [IPAddressRenewalTime](https://github.com/search?q=IPAddressRenewalTime&type=Code)(no hits except this lib) or [ElapsedTime](https://github.com/search?l=Go&p=3&q=dhcpv6+ElapsedTime&type=Code)(None of the hits are users of this lib). 
So maybe the simplest solution here is to not have a getter at all, until we have a client that wants it and knows what they want ?

mcphailtom:
So I'm happy to provide a single data point for consumption... I consume (privately). Personally I don't really mind which way it goes, though the default option basically just shortens the path to what I would ultimately do as a consumer anyway.

pmazzini:
I'll merge it tomorrow if there are no other comments.

PR #397, opened 2020-09-12
Add Additional IANA Architecture Types

bensallen:
Adds the additional architecture types as defined by IANA
http://www.iana.org/assignments/dhcpv6-parameters/dhcpv6-parameters.xhtml#processor-architecture.

Note, RFC4578 errata, https://www.rfc-editor.org/errata_search.php?rfc=4578

"The values for EFI BC and EFI x86-64 should be swapped. UEFI implementations use value 7 to report EFI x86-64, not value 9."

IANA's list also agrees with this swap. Thus I've swapped these values here to match.

I took some liberty with the new architecture's mnemonic names compared to the descriptions from IANA. This was an attempt to standardize wording, capitalization, and match the existing style.

### DIFF
--- a/iana/archtype.go
+++ b/iana/archtype.go
@@ -10,7 +10,7 @@ import (
 // Arch encodes an architecture type per RFC 4578, Section 2.1.
 type Arch uint16
 
-// See RFC 4578.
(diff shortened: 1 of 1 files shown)


pmazzini (approved):
Thanks

hugelgupf:
Thanks Ben

PR #400, opened 2020-09-16
Bender compat for nclients

hugelgupf:
Trying to move github.com/pinterest/bender to nclients, so we can delete async before we release a v1.0.0.

Also add support for NAKs in DHCPv4.


pmazzini (approved):
lgtm

PR #404, opened 2020-12-11
Abort parsing options with 0 length

mcphailtom:
Found a case where DNSmasq sends an option with length set to zero which causes the parser to error on the packet. Does seem redundant to send a zero length option, however it seemed pragmatic to add a guard against parsing them which avoids  failing on the packet.

Cannot find anything in the RFC that explicitly says an option with length set to zero is invalid

### DIFF
--- a/dhcpv6/options.go
+++ b/dhcpv6/options.go
@@ -201,6 +201,11 @@ func (o *Options) FromBytesWithParser(data []byte, parser OptionParser) error {
 		code := OptionCode(buf.Read16())
 		length := int(buf.Read16())
 
+		// Abort trying to parse options with zero length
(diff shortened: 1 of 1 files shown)


Natolumin:
The rapid-commit option is always 0-length: https://tools.ietf.org/html/rfc8415#section-21.14 so this won't work

mcphailtom:
> The rapid-commit option is always 0-length: https://tools.ietf.org/html/rfc8415#section-21.14 so this won't work

Well that is unfortunate. I see there are some other related PRs i.e. #396 that are trying to deal with parsing failures. Is there a train of thought as to the direction of error handling, or is the preference to leave as is?

PR #407, opened 2021-02-19
Pad packets after the 'end' option to be more conformant

xcllnt:
RFC 2131 gives two examples of padding in section 4.1. For both the
examples, the 'end' option preceeds the padding with the 'pad' option.
These examples relate to overloading pre-defined fields for options,
not padding the packet for a minimal size.

Older versions of the U-Boot firmware support DHCP, but do not expect
the 'pad' option. The firmware logs the following warning for each
occurence:
    *** Unhandled DHCP Option in OFFER/ACK: 0
This warning disappears when the 'end' option preceeds the 'pad'
option.

### DIFF
--- a/dhcpv4/async/client.go
+++ b/dhcpv4/async/client.go
@@ -9,8 +9,8 @@ import (
 	"time"
 
 	promise "github.com/fanliao/go-promise"
-	"github.com/insomniacslk/dhcp/dhcpv4"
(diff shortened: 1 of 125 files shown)


pmazzini:
The diff has a lot of unwanted stuff like changing the imports. Do you mind cleaning it?

xcllnt:
Ugh. I didn't intended to have additional commits added to the pull request...

PR #409, opened 2021-03-10
Improve compatibility with some devices

EugeneOne1:
We, the _AdGuard_ team, have received an [issue](https://github.com/AdguardTeam/AdGuardHome/issues/2582) some time ago.  After researching a little we've found that some devices may include trailing zero-byte in [DHCP option 12](https://tools.ietf.org/html/rfc2132#section-3.14) length calculation.  We've also found that similar issues with some other options were already solved in fcbde74dab23754aa33fbcd409327c0a1409f2a7.

For now, we've come up with a [little workaround](https://github.com/AdguardTeam/AdGuardHome/commit/5aa0ca9319919f35eb2e1dc4c9e6d8f0385c394e) in _AdGuard Home_, but it's a temporary solution.  Please review our little fix, we hope to get rid of our crutch as soon as possible :)

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -604,7 +604,8 @@ func (d *DHCPv4) DomainName() string {
 //
 // The Host Name option is described by RFC 2132, Section 3.14.
 func (d *DHCPv4) HostName() string {
-	return GetString(OptionHostName, d.Options)
+	name := GetString(OptionHostName, d.Options)
+	return strings.TrimRight(name, "\x00")
 }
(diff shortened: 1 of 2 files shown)


pmazzini:
As RFC2132 states "Options containing NVT ASCII data SHOULD NOT include a trailing NULL; however, the receiver of such options MUST be prepared to delete trailing nulls if they exist."

This is consistent with #286. LGTM

pmazzini (approved):

hugelgupf (approved):

PR #410, opened 2021-03-12
nclient4: Ensure server identifier matches

mfischer:
Ensure server identifier matches the offer server identifier when
looking for ACKs or NAKs.

This protects against rogue DHCP servers NAKing requests the shouldn't.

Signed-off-by: Moritz Fischer <moritzf@google.com>

### DIFF
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -414,6 +414,25 @@ func IsMessageType(t dhcpv4.MessageType, tt ...dhcpv4.MessageType) Matcher {
 	}
 }
 
+// IsCorrectServer returns a matcher that checks for the correct ServerAddress.
+func IsCorrectServer(s net.IP) Matcher {
+	return func(p *dhcpv4.DHCPv4) bool {
+		return p.ServerIdentifier().Equal(s)
+	}
+}
+
+// IsAll returns a matcher that checks for all given matchers to be true.
+func IsAll(ms ...Matcher) Matcher {
+	return func(p *dhcpv4.DHCPv4) bool {
+		for _, m := range matchers {
+			if !m(p) {
+				return false
+			}
+		}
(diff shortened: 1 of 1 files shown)


dhcpv4/nclient4/client.go
```diff
@@ -482,7 +501,9 @@ func (c *Client) RequestFromOffer(ctx context.Context, offer *dhcpv4.DHCPv4, mod
 		return nil, fmt.Errorf("unable to create a request: %w", err)
 	}
 
-	response, err := c.SendAndRead(ctx, c.serverAddr, request, IsMessageType(dhcpv4.MessageTypeAck, dhcpv4.MessageTypeNak))
+	response, err := c.SendAndRead(ctx, c.serverAddr, request, IsAll(
+		IsCorrectServer(offer.ServerIdentifier),
```
hugelgupf, line 505: offer.ServerIdentifier()?

hugelgupf: I was actually just looking for a mention in the RFC for validating this field. It doesn't require it. I'd add a comment here saying something like

// Servers are supposed to only respond to Requests containing their server identifier, but sometimes non-compliant servers respond anyway. Clients are not required to validate this field, but servers are required to include the server identifier in their Offer per RFC 2131 Section 4.3.1 Table 3.

mfischer: Yeah, makes sense :) Unfortunately I know at least one misbehaving server :-)

hugelgupf (approved):

hugelgupf:
i'm gonna wait till monday to see if @insomniacslk or @pmazzini have comments

pmazzini (approved):
lgtm

hugelgupf:
lol it doesn't build, but travis didn't catch it? looking into it.

PR #414, opened 2021-03-14
Switch to GitHub Actions

insomniacslk:
TravisCI not supported anymore.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,71 @@
+name: Tests
+
+on: [push, pull_request]
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go: ['1.14', '1.15', '1.16']
```
hugelgupf, line 10: mind leaving 1.13 as long as u-root is on 1.13?

insomniacslk: sure

insomniacslk: done

insomniacslk:
Linter issue fixed (outdated version, now using latest). Unit tests are now green. Fixed interface name (different from Travis-CI). Still one failure on integ tests:
```
sudo: unable to resolve host fv-az206-747: Temporary failure in name resolution
```

insomniacslk:
`EAI_AGAIN` from libc means SERVFAIL from the DNS server. Smells like an outbound filter, but this is odd, the same setup works on other projects

insomniacslk:
ooh it could be that integ tests mess up the network configuration. codecov runs fine for unit tests, so it's not a github network policy issue

insomniacslk:
I bet that the issue is in `TestConfigureInterfaceWithRouteAndDNS`, where the DNS is configured to be `8.8.8.8`. My guess is that Travis-CI didn't block 8.8.8.8 while GitHub Actions does

insomniacslk:
also - this means that `netboot.ConfigureInterface` only works on Linux but not macOS/Windows, as it writes to "/etc/resolv.conf"

insomniacslk:
need to sign off now. @hugelgupf feel free to commit on top of this PR if you have a solution for the DNS failure in the integ tests (see my previous comments for what's my guess on the cause).

Re. codecov's failure, it's complaining that we cover less than 70%. We can merge despite that, as it's the same as than before, and fix it later

insomniacslk:
yeah, two tests that call `netboot.ConfigureInterface` were messing up with /etc/resolv.conf. Saving/restoring its content, and enforcing sequential testing with subtests seems to address the issue

pmazzini (approved):

insomniacslk:
CC @hugelgupf this is merged. We can work out the testing coverage separately

PR #418, opened 2021-04-20
Adding xe interface format for Juniper qfx to list of circuit regexes

hhkumar:
This adds a new regex format of circuit ids for the Juniper qfx xe interface to the list of circuit regexes to compare against.


dhcpv4/ztpv4/parse_circuitid.go
```diff
@@ -19,6 +19,8 @@ type CircuitID struct {
 var circuitRegexs = []*regexp.Regexp{
 	// Juniper QFX et-0/0/0:0.0
 	regexp.MustCompile("^et-(?P<slot>[0-9]+)/(?P<mod>[0-9]+)/(?P<port>[0-9]+):(?P<subport>[0-9]+).*$"),
+	// Juniper QFX xe-0/0/0:0.0
+	regexp.MustCompile("^xe-(?P<slot>[0-9]+)/(?P<mod>[0-9]+)/(?P<port>[0-9]+):(?P<subport>[0-9]+).*$"),
```
pmazzini, line 23: I wonder if it should be consolidated with the other Juniper QTX regexp.

pmazzini (approved):
other than the comment I left, it looks good to me

pmazzini (approved):
lgtm, thanks

dhcpv4/ztpv4/parse_circuitid_test.go
```diff
@@ -18,7 +18,8 @@ func TestMatchCircuitID(t *testing.T) {
 		{name: "juniperQFX pattern", circuit: "et-0/0/0:0.0", want: &CircuitID{Slot: "0", Module: "0", Port: "0", SubPort: "0"}},
```
pmazzini, line 18: I realize now that the test should be grouped with this pattern instead.

hhkumar: You mean since we have "juniperQFX pattern",  we dont need the other two tests "Juniper QFX pattern et" and ""Juniper QFX pattern xe"?

pmazzini: There are 2 Juniper QFX patterns:

1) https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/ztpv4/parse_circuitid.go#L21
2) https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/ztpv4/parse_circuitid.go#L30

This grouping is mixing both patterns and making this a bit more confusing.

I think the intention was to add the test close to the other one.

PR #419, opened 2021-04-20
rearrange parse_circuitid test cases

hhkumar:
Incorrect test case was modified in https://github.com/insomniacslk/dhcp/pull/418.

Reverting modified test case for "Juniper QFX pattern" and moving test case closer to "juniperQFX pattern".


pmazzini (approved):
thanks

PR #420, opened 2021-04-23
Add dhcp Opt 124(vivc) parsing to ztp

hhkumar:
This adds ability to derive Serial Number and Model from Opt 124 which is the OptionVendorIdentifyingVendorClass. This option encodes the Vendor Enterprise ID, Serial number and model in it.


dhcpv4/ztpv4/ztp.go
```diff
@@ -15,14 +16,39 @@ type VendorData struct {
 
 var errVendorOptionMalformed = errors.New("malformed vendor option")
 
+const entIDCiscoSystems = 0x9
```
pmazzini, line 19: Please move this to the iana package.

https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,14 +16,39 @@ type VendorData struct {
...
+
 // ParseVendorData will try to parse dhcp4 options looking for more
 // specific vendor data (like model, serial number, etc).
 func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
+	vd := &VendorData{}
+	
+	vivc := packet.VIVC()
+        for _, id := range vivc {
```
pmazzini, line 27: Indentation looks wrong here. Can you try go fmt?

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,14 +16,39 @@ type VendorData struct {
...
 var errVendorOptionMalformed = errors.New("malformed vendor option")
 
+const entIDCiscoSystems = 0x9
+
 // ParseVendorData will try to parse dhcp4 options looking for more
 // specific vendor data (like model, serial number, etc).
 func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
+	vd := &VendorData{}
+	
+	vivc := packet.VIVC()
```
pmazzini, line 26: This function is getting overloaded. Can you split things in different functions?

I think we would benefit from some code separation:
* A function for looking at the `VIVC` option
* A separate function for looking at the `ClassIdentifier` option
* And `ParseVendorData` calling both of them

pmazzini:
Why did you close it?

pmazzini:
Continuation in #425

PR #421, opened 2021-04-24
dhcpv4,6: Relay Port options (RFC8357)

Natolumin:
Implements the options specified in RFC8357 for both v4 and v6, intended to communicate across relays which server port should be used to contact the relay

For DHCPv4 this is a simple 0-length suboption of option 82(relay options) that tells the end server to use the same port for the return path that the relay used for the direct path. 
Since this is 0-length it only needs the option code constant and can be otherwise managed with the generic functions

For DHCPv6 this is a 1-value top-level option containing just the port, that is sent next to the RelayMsg option to indicate the port to use for the next relay in the chain


dhcpv6/options.go
```diff
@@ -57,6 +57,8 @@ func ParseOption(code OptionCode, optData []byte) (Option, error) {
 		opt, err = parseOptElapsedTime(optData)
 	case OptionRelayMsg:
 		opt, err = parseOptRelayMsg(optData)
+	case OptionRelayPort:
```
pmazzini, line 60: NIT: I think this is sorted by option number

pmazzini:
lgtm

dhcpv6/options.go
Natolumin: :facepalm: yes it is

pmazzini: Shouldn't it go almost to the end?

https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/types.go#L224

Natolumin: yeah sorry that's what I meant, I hadn't realized it was sorted, I'll update

pmazzini (approved):

PR #423, opened 2021-04-25
[interfaces] darwin

### LINKED ISSUE #378: [BUG] Binding to an interface on MacOS does not function correctly
When picking an interface on MacOS, the server will correctly listen for packets, but when doing an offer by replying to broadcast address, the system will route that to the Gateway interface only.

This is likely due to: 
```
// BindToInterface emulates linux's SO_BINDTODEVICE option for a socket by using
// IP_RECVIF.
func BindToInterface(fd int, ifname string) error {
	iface, err := net.InterfaceByName(ifname)
	if err != nil {
		return err
	}
	return syscall.SetsockoptInt(fd, syscall.IPPROTO_IP, syscall.IP_RECVIF, iface.Index)
}
```

I am hardly an expert in stuff this low in the stack, though from the looks of it this is just filtering the received traffic, but does nothing to direct the outbound traffic.

I have also found this, perhaps that might be the solution: 🤷‍♂ :

https://stackoverflow.com/a/57013928

> There is IP_BOUND_IF socket option
> `int idx = if_nametoindex("en0");`
> `setsockopt(sockfd, IPPROTO_IP, IP_BOUND_IF, &idx, sizeof(idx))`

Spent quite a few hours with Wireshark trying to figure out why the server was not sending an offer to the client, all the while not raising any alarms.

Will add a minimum viable example to illustrate the issue soon.

pmazzini:
Fixes #378


pmazzini:
ping

insomniacslk (approved):

insomniacslk:
The branch needs to be update. Otherwise looks good to me

insomniacslk (approved):

insomniacslk:
Screenshot_20210809-212203

Huh

pmazzini:
Did I break something?

nadoo:
@pmazzini `GOOS=freebsd GOARCH=amd64 go build`
```
bindtodevice_bsd.go:8:2: imported and not used: "golang.org/x/sys/unix"
bindtodevice_bsd.go:18:9: undefined: syscall
```

pmazzini:
> can't build for freebsd after this commit.

Sorry, https://github.com/insomniacslk/dhcp/pull/438 should fix it.

PR #424, opened 2021-04-25
[netboot] DNS not mandatory

### LINKED ISSUE #359: Why are dns mandatory in GetNetConfFromPacketv4
I noticied that the function GetNetConfFromPacketv4 return an error if it dosen't find any dns server.
This cause the dhcp client to fail if the server dosen't return any dns server.

is there a reason or is this a bug?

i tried reading the dhcp specs and for what i understand dns are NOT mandatory.

pmazzini:
Fixes #359

Similar to #197


pmazzini:
ping

insomniacslk (approved):

PR #425, opened 2021-04-26
Add dhcp Opt 124(vivc) parsing to ztp

hhkumar:
This adds ability to derive Serial Number and Model from Opt 124 which is the OptionVendorIdentifyingVendorClass. This option encodes the Vendor Enterprise ID, Serial number and model in it.


iana/entID.go
```diff
@@ -0,0 +1,4 @@
+package iana
```
pmazzini, line 1: You created a file with the name `entID.go`.

All the rest of the file names are lower case:

```
archtype.go
hwtypes.go
iana.go
statuscodes.go
```

 Can it be lower case for consistency?

hhkumar: sure will do

iana/entID.go
```diff
@@ -0,0 +1,4 @@
...
+
+// EntIDCiscoSystems is IANA Enterprise ID for Cisco Systems
+const EntIDCiscoSystems = 0x9
```
pmazzini, line 4: Can you define a type like in all the rest of the definitions on the IANA package?

pmazzini: This comment still stands.

iana/entid.go
```diff
@@ -0,0 +1,6 @@
+package iana
+
+type EntID uint32
+
+// EntIDCiscoSystems is IANA Enterprise ID for Cisco Systems
+const EntIDCiscoSystems EntID = 0x9
```
pmazzini, line 6: Does it need to be hexa?

I would use decimal like in the IANA definitions:

https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers

iana/entid.go
```diff
@@ -0,0 +1,6 @@
+package iana
+
+type EntID uint32
+
```
pmazzini, line 4: Can you add a `EntIDToString` map like in other IANA types and a String method?

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,39 +17,60 @@ type VendorData struct {
...
 var errVendorOptionMalformed = errors.New("malformed vendor option")
 
-// ParseVendorData will try to parse dhcp4 options looking for more
-// specific vendor data (like model, serial number, etc).
-func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
+func parseV4VIVC(vd *VendorData, packet *dhcpv4.DHCPv4) error {
+	vivc := packet.VIVC()
+	for _, id := range vivc {
+		if id.EntID == iana.EntIDCiscoSystems {
+			vd.VendorName = "Cisco Systems"
```
pmazzini, line 24: Can the "Cisco Systems" string come from the IANA package?

Reference comment below.

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,39 +17,60 @@ type VendorData struct {
 
 var errVendorOptionMalformed = errors.New("malformed vendor option")
 
-// ParseVendorData will try to parse dhcp4 options looking for more
-// specific vendor data (like model, serial number, etc).
-func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
+func parseV4VIVC(vd *VendorData, packet *dhcpv4.DHCPv4) error {
+	vivc := packet.VIVC()
+	for _, id := range vivc {
```
pmazzini, line 22:
```
	vivc := packet.VIVC()
	for _, id := range vivc {
```

This can be simplified to:

```
	for _, id := range packet.VIVC() {
```

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,39 +17,60 @@ type VendorData struct {
 
 var errVendorOptionMalformed = errors.New("malformed vendor option")
 
-// ParseVendorData will try to parse dhcp4 options looking for more
-// specific vendor data (like model, serial number, etc).
-func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
+func parseV4VIVC(vd *VendorData, packet *dhcpv4.DHCPv4) error {
```
pmazzini, line 20: `parseV4VIVC`

Do you need the `V4` in the name? I think it is redundant because it is already in the package name.

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,39 +17,60 @@ type VendorData struct {
...
+				case "PID":
+					vd.Model = string(p[1])
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func parseV4VendorClass(vd *VendorData, packet *dhcpv4.DHCPv4) error {
```
pmazzini, line 44: Same, I think the `V4` in the name is redundant, it is already in the package name.

iana/entid.go
hhkumar: Yes, because the enterprise ID in https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/option_vivc.go#L14 is a uint32. 

Not sure if we should maintain consistency or convert it to uint32 where it is being used. Any thoughts?

pmazzini: > Yes, because the enterprise ID s a uint32

What does that have to do with using decimal vs hexa numbers?

pmazzini (changes requested):
tests

iana/entid.go
hhkumar: Could go either way, if it makes more sense, I can have it as a decimal here and convert to uint32 where its being used

pmazzini: With regards to the `uint32` conversion: The ideal way would be to make the `OptVendorClass` struct to use a `iana.EntID` type instead of uint32.

Similar to how it is done here:
https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/option_clientlinklayeraddress.go#L18

hhkumar: You mean change VIVCIdentifier EntID  type from uint32 to iana.EntID right in https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/option_vivc.go#L14?

https://github.com/insomniacslk/dhcp/search?q=OptVendorClass does not seem to be related in this case.

pmazzini: Can you use the decimal "9" instead of "0x9"? I think we would benefit from this in case more enterprise IDs are added.

pmazzini: > You mean change VIVCIdentifier EntID type from uint32 to iana.EntID right in https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/option_vivc.go#L14?

Yes but it may be out of the scope of this PR, up to you.

> https://github.com/insomniacslk/dhcp/search?q=OptVendorClass does not seem to be related in this case.

That one should be changed as well but it is for v6, so yes, not related to this case.

iana/entid.go
```diff
@@ -0,0 +1,20 @@
+package iana
+
+// EntID represents the Enterprise IDs as set by IANA
+// https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers
+type EntID int
+
+// EntIDCiscoSystems is IANA Enterprise ID for Cisco Systems
+const EntIDCiscoSystems EntID = 9
```
pmazzini, line 8: NIT: I think that if you have a const scope like the following, you don't need to have a comment for each of the elements:

```
// See IANA for values.
const (
	EntIDCiscoSystems EntID = 9
)
```

This will leave it in better shape for extending it with other values.

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,39 +17,59 @@ type VendorData struct {
 
 var errVendorOptionMalformed = errors.New("malformed vendor option")
 
-// ParseVendorData will try to parse dhcp4 options looking for more
-// specific vendor data (like model, serial number, etc).
-func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
+func parseVIVC(vd *VendorData, packet *dhcpv4.DHCPv4) error {
```
pmazzini, line 20: I am not a fan of modifying the parameter, can it return a `*VendorData` instead and make it consistent with the `ParseVendorData` function definition?

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,39 +17,59 @@ type VendorData struct {
...
+				case "PID":
+					vd.Model = string(p[1])
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func parseVendorClass(vd *VendorData, packet *dhcpv4.DHCPv4) error {
```
pmazzini, line 43: I am not a fan of modifying the parameter, can it return a `*VendorData` instead and make it consistent with the `ParseVendorData` function definition?

dhcpv4/ztpv4/ztp.go
hhkumar: So, return vd explicitly instead of modifying it there?

pmazzini: Yes, and not take a `*VendorData` as parameter. Is it needed as parameter?

hhkumar: If we want to modify it there, then just pass a pointer (the way its implemented now). I honestly like that approach since its cleaner. 

Otherwise, we need to initialize a `vd := &VendorData{}` in each of the functions and modify the local version of vd and return it. I can go either way :D

dhcpv4/ztpv4/ztp_test.go
```diff
@@ -4,45 +4,60 @@ import (
...
 	"github.com/stretchr/testify/require"
 )
 
 func TestParseV4VendorClass(t *testing.T) {
 	tt := []struct {
-		name         string
-		vc, hostname string
-		want         *VendorData
-		fail         bool
+		name     string
```
pmazzini, line 13: Can this test be left unmodified and have a new one for `parseVIVC`?

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,14 +17,36 @@ type VendorData struct {
...
+				case "PID":
+					vd.Model = string(p[1])
+				}
+			}
+		}
+	}
+	return vd, nil
+}
+
+func parseVendorClass(packet *dhcpv4.DHCPv4) (*VendorData, error) {
```
pmazzini, line 44: maybe call this `parseClassIdentifier` to match the packet method and to be consistent with the `parseVIVC` function.

dhcpv4/ztpv4/ztp.go
```diff
@@ -71,6 +95,32 @@ func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
...
 
-	// We didn't match anything.
-	return nil, errors.New("no known ZTP vendor found")
+	return vd, nil
+}
+
+// ParseVendorData will try to parse dhcp4 options looking for more
+// specific vendor data (like model, serial number, etc).
+func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
+	vd, err := parseVIVC(packet)
```
pmazzini, line 104: NIT: can you call first the function that parses the lower option number?

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,14 +17,36 @@ type VendorData struct {
 
 var errVendorOptionMalformed = errors.New("malformed vendor option")
 
-// ParseVendorData will try to parse dhcp4 options looking for more
-// specific vendor data (like model, serial number, etc).
-func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
+func parseVIVC(packet *dhcpv4.DHCPv4) (*VendorData, error) {
```
pmazzini, line 20: the `parseVIVC` and `parseVendorClass` functions are not consistent, one returns an error if the option is not found and the other does not.

Maybe non of them should return an error in such case and only return an error if there is a parsing issue?

hhkumar: Modified parseClassIdentifier to not return an error when vc is empty.

dhcpv4/ztpv4/ztp.go
hhkumar: Called parseClassIdentifier before parseVIVC

dhcpv4/ztpv4/ztp_test.go
```diff
@@ -70,3 +71,48 @@ func TestParseV4VendorClass(t *testing.T) {
 		})
 	}
 }
+
+func TestParseV4VIVC(t *testing.T) {
```
pmazzini, line 75: Can you remove the V4 from the test name? I think it is redundant.

I know the previous test had it.

dhcpv4/ztpv4/ztp_test.go
```diff
@@ -70,3 +71,48 @@ func TestParseV4VendorClass(t *testing.T) {
...
+	tt := []struct {
+		name         string
+		vivc		 string
```
pmazzini, line 78: Indentation does not look good. Can you use go fmt?

dhcpv4/ztpv4/ztp_test.go
```diff
@@ -4,10 +4,11 @@ import (
 	"testing"
 
 	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/insomniacslk/dhcp/iana"
 	"github.com/stretchr/testify/require"
 )
 
-func TestParseV4VendorClass(t *testing.T) {
+func TestParseV4ClassIdentifier(t *testing.T) {
```
pmazzini, line 11: Since you are changing the name, would you mind removing the V4? I think it is redindant.

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,62 +17,107 @@ type VendorData struct {
...
 var errVendorOptionMalformed = errors.New("malformed vendor option")
 
-// ParseVendorData will try to parse dhcp4 options looking for more
-// specific vendor data (like model, serial number, etc).
-func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
-	vc := packet.ClassIdentifier()
-	if len(vc) == 0 {
-		return nil, errors.New("vendor options not found")
-	}
+func parseVIVC(packet *dhcpv4.DHCPv4) (*VendorData, error) {
```
pmazzini, line 20: NIT: Can you move `parseVIVC` below `parseClassIdentifier`?

Github diff will look cleaner and would be nice to have them sorted as option number as well.

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,62 +17,107 @@ type VendorData struct {
...
-		} else {
-			vd.Model = strings.Join(p[1:len(p)-1], "-")
-			vd.Serial = p[len(p)-1]
+
+			vd.VendorName = p[0]
+			return vd, nil
 		}
 
-		vd.VendorName = p[0]
+	return vd, nil
```
pmazzini, line 95: Should it return nil `VendorData` here and check the pointer is not null afterwards instead of checking for an empty struct?

pmazzini (commented):
Some NITs. It is looking better in general. Thanks

dhcpv4/ztpv4/ztp.go
hhkumar: I will return a nil, nil when VendorData is not set in either functions and return if it is not nil in ParseVendorData

dhcpv4/ztpv4/ztp.go
```diff
@@ -15,14 +17,9 @@ type VendorData struct {
...
-// ParseVendorData will try to parse dhcp4 options looking for more
-// specific vendor data (like model, serial number, etc).
-func ParseVendorData(packet *dhcpv4.DHCPv4) (*VendorData, error) {
-	vc := packet.ClassIdentifier()
-	if len(vc) == 0 {
-		return nil, errors.New("vendor options not found")
-	}
+func parseClassIdentifier(packet *dhcpv4.DHCPv4) (*VendorData, error) {
 	vd := &VendorData{}
+	vc := packet.ClassIdentifier()
```
pmazzini, line 22: NIT: can this be called in the switch?

I think it will be something like `switch vc := packet.ClassIdentifier(), vc {`

pmazzini:
Last NIT. Otherwise looks good to me.

pmazzini (approved):
lgtm

PR #429, opened 2021-05-28
Break circular dependency on u-root

hugelgupf:
u-root -> dhcp -> u-root is the circular dependency. Which is fine,
except we messed up modules in u-root big time. We need to break the
circular dep for now.

One warning!

uio does not use "uroot.nohwrng" on the cmdline anymore. Now, UROOT_NOHWRNG=$anything has to be specified in the environment (and one can put that in the cmdline as well, actually)


JohnAZoidberg (approved):
I have no authority to approve this.

pmazzini (approved):
lgtm

PR #434, opened 2021-08-08
[dhcpv4/dhcpv6] Adding Selector function for dropping packets that do not meet user-defined conditions.

63818569:
user-defined conditions means selecting packets using VendorOpts structure


pmazzini:
Can you describe a bit more the use case for this?

Do packets need to be filtered at this level or can a user filter them afterwards instead?

63818569:
> Can you describe a bit more the use case for this?
> 
> Do packets need to be filtered at this level or can a user filter them afterwards instead?

image
RFC2132 described below: 

8.4. Vendor Specific Information

   This option is used by clients and servers to exchange vendor-
   specific information.  The information is an opaque object of n
   octets, presumably interpreted by vendor-specific code on the clients
   and servers.  The definition of this information is vendor specific.
   The vendor is indicated in the vendor class identifier option.
   Servers not equipped to interpret the vendor-specific information
   sent by a client MUST ignore it (although it may be reported).
   Clients which do not receive desired vendor-specific information
   SHOULD make an attempt to operate without it, although they may do so
   (and announce they are doing so) in a degraded mode.

PR #436, opened 2021-08-09
Added OptNTPServer

insomniacslk:

### DIFF
--- a/dhcpv6/dhcpv6message.go
+++ b/dhcpv6/dhcpv6message.go
@@ -288,6 +288,32 @@ func (mo MessageOptions) DHCP4oDHCP6Server() *OptDHCP4oDHCP6Server {
 	return nil
 }
 
+// NTPServers returns the NTP server addresses contained in the
+// NTP_SUBOPTION_SRV_ADDR of an OPTION_NTP_SERVER.
+// If multiple NTP server options exist, the function will return all the NTP
+// server addresses it finds, as defined by RFC 5908.
+func (mo MessageOptions) NTPServers() []net.IP {
+	opts := mo.Options.Get(OptionNTPServer)
+	if opts == nil {
+		return nil
+	}
+	addrs := make([]net.IP, 0)
+	for _, opt := range opts {
+		ntp, ok := opt.(*OptNTPServer)
+		if ok {
+			continue
+		}
+		for _, subopt := range ntp.Suboptions {
+			so, ok := subopt.(*NTPSuboptionSrvAddr)
+			if !ok {
+				continue
+			}
+			addrs = append(addrs, net.IP(*so))
+		}
+	}
+	return addrs
+}
+
 // Message represents a DHCPv6 Message as defined by RFC 3315 Section 6.
 type Message struct {
 	MessageType   MessageType
--- a/dhcpv6/option_ntp_server.go
+++ b/dhcpv6/option_ntp_server.go
@@ -1,0 +1,155 @@
+package dhcpv6
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/insomniacslk/dhcp/rfc1035label"
+	"github.com/u-root/uio/uio"
+)
+
+// NTPSuboptionSrvAddr is NTP_SUBOPTION_SRV_ADDR according to RFC 5908.
+type NTPSuboptionSrvAddr net.IP
+
+// Code returns the suboption code.
+func (n *NTPSuboptionSrvAddr) Code() OptionCode {
+	return NTPSuboptionSrvAddrCode
+}
+
+// ToBytes returns the byte serialization of the suboption.
+func (n *NTPSuboptionSrvAddr) ToBytes() []byte {
+	buf := uio.NewBigEndianBuffer(nil)
+	buf.Write16(uint16(NTPSuboptionSrvAddrCode))
+	buf.Write16(uint16(net.IPv6len))
+	buf.WriteBytes(net.IP(*n).To16())
+	return buf.Data()
+}
+
+func (n *NTPSuboptionSrvAddr) String() string {
+	return fmt.Sprintf("Server Address: %s", net.IP(*n).String())
+}
+
+// NTPSuboptionMCAddr is NTP_SUBOPTION_MC_ADDR according to RFC 5908.
+type NTPSuboptionMCAddr net.IP
+
+// Code returns the suboption code.
+func (n *NTPSuboptionMCAddr) Code() OptionCode {
+	return NTPSuboptionMCAddrCode
+}
+
+// ToBytes returns the byte serialization of the suboption.
+func (n *NTPSuboptionMCAddr) ToBytes() []byte {
+	buf := uio.NewBigEndianBuffer(nil)
+	buf.Write16(uint16(NTPSuboptionMCAddrCode))
+	buf.Write16(uint16(net.IPv6len))
+	buf.WriteBytes(net.IP(*n).To16())
+	return buf.Data()
+}
+
+func (n *NTPSuboptionMCAddr) String() string {
+	return fmt.Sprintf("Multicast Address: %s", net.IP(*n).String())
+}
+
+// NTPSuboptionSrvFQDN is NTP_SUBOPTION_SRV_FQDN according to RFC 5908.
+type NTPSuboptionSrvFQDN rfc1035label.Labels
+
+// Code returns the suboption code.
+func (n *NTPSuboptionSrvFQDN) Code() OptionCode {
+	return NTPSuboptionSrvFQDNCode
+}
+
+// ToBytes returns the byte serialization of the suboption.
+func (n *NTPSuboptionSrvFQDN) ToBytes() []byte {
+	buf := uio.NewBigEndianBuffer(nil)
+	buf.Write16(uint16(NTPSuboptionSrvFQDNCode))
+	l := rfc1035label.Labels(*n)
+	buf.Write16(uint16(l.Length()))
+	buf.WriteBytes(l.ToBytes())
+	return buf.Data()
+}
+
+func (n *NTPSuboptionSrvFQDN) String() string {
+	l := rfc1035label.Labels(*n)
+	return fmt.Sprintf("Server FQDN: %s", l.String())
+}
+
+// NTPSuboptionSrvAddr is the value of NTP_SUBOPTION_SRV_ADDR according to RFC 5908.
+const (
+	NTPSuboptionSrvAddrCode = OptionCode(1)
+	NTPSuboptionMCAddrCode  = OptionCode(2)
+	NTPSuboptionSrvFQDNCode = OptionCode(3)
+)
+
+// parseNTPSuboption implements the OptionParser interface.
+func parseNTPSuboption(code OptionCode, data []byte) (Option, error) {
+	//var o Options
+	buf := uio.NewBigEndianBuffer(data)
+	length := len(data)
+	data, err := buf.ReadN(length)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read %d bytes for suboption: %w", length, err)
+	}
+	switch code {
+	case NTPSuboptionSrvAddrCode, NTPSuboptionMCAddrCode:
+		if length != net.IPv6len {
+			return nil, fmt.Errorf("invalid suboption length, want %d, got %d", net.IPv6len, length)
+		}
+		var so Option
+		switch code {
+		case NTPSuboptionSrvAddrCode:
+			sos := NTPSuboptionSrvAddr(data)
+			so = &sos
+		case NTPSuboptionMCAddrCode:
+			som := NTPSuboptionMCAddr(data)
+			so = &som
+		}
+		return so, nil
+	case NTPSuboptionSrvFQDNCode:
+		l, err := rfc1035label.FromBytes(data)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse rfc1035 labels: %w", err)
+		}
+		// TODO according to rfc3315, this label must not be compressed.
+		// Need to add support for compression detection to the
+		// `rfc1035label` package in order to do that.
+		so := NTPSuboptionSrvFQDN(*l)
+		return &so, nil
+	default:
+		gopt := OptionGeneric{OptionCode: code, OptionData: data}
+		return &gopt, nil
+	}
+}
+
+// ParseOptNTPServer parses a sequence of bytes into an OptNTPServer object.
+func ParseOptNTPServer(data []byte) (*OptNTPServer, error) {
+	var so Options
+	if err := so.FromBytesWithParser(data, parseNTPSuboption); err != nil {
+		return nil, err
+	}
+	return &OptNTPServer{
+		Suboptions: so,
+	}, nil
+}
+
+// OptNTPServer is an option NTP server as defined by RFC 5908.
+type OptNTPServer struct {
+	Suboptions Options
+}
+
+// Code returns the option code
+func (op *OptNTPServer) Code() OptionCode {
+	return OptionNTPServer
+}
+
+// ToBytes returns the option serialized to bytes.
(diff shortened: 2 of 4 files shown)


dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,47 @@
+package dhcpv6
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/u-root/uio/uio"
+)
+
+// OptNTPServer returns an NTP Server option as defined by RFC 3646.
```
pmazzini, line 10: This is not the proper RFC

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (op *optNTPServer) String() string {
+	return fmt.Sprintf("NTP: %v", op.NTPServers)
+}
+
+// parseOptNTPServer builds an optNTPServer structure
+// from a sequence of bytes. The input data does not include option code and length
+// bytes.
+func parseOptNTPServer(data []byte) (*optNTPServer, error) {
```
pmazzini, line 40: Is this following the proper option structure as defined in the RFC?
https://datatracker.ietf.org/doc/html/rfc5908

pmazzini (dismissed):
left some comments

dhcpv6/option_ntp_server.go
insomniacslk: no it's not. I noticed right after putting the PR up and commented in the internal chat, but the wire format is wrong and I have to adjust these functions

insomniacslk:
Now it should be RFC-compliant, @pmazzini can you take another look please?

insomniacslk:
rebased

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,58 @@
...
+	var opt optNTPServer
+	buf := uio.NewBigEndianBuffer(data)
+	for buf.Has(2 + net.IPv6len) {
+		suboptionPrefix := buf.CopyN(2)
+		if suboptionPrefix[0] != NTPSuboptionSrvAddr {
```
pmazzini, line 49: I don't think this is good. It will fail to parse RFC compliant options.

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,58 @@
...
+// NTPSuboptionSrvAddr is the value of NTP_SUBOPTION_SRV_ADDR according to RFC 5908.
+const NTPSuboptionSrvAddr byte = 1
+
+// OptNTPServer returns an NTP Server option as defined by RFC 5908.
+func OptNTPServer(ntpServers ...net.IP) Option {
+	return &optNTPServer{NTPServers: ntpServers}
+}
+
+type optNTPServer struct {
+	NTPServers []net.IP
```
pmazzini, line 19: This is not enough information for the RFC defined option.

I think this should be a an option list like in: https://github.com/insomniacslk/dhcp/blob/d4b282fe8d4cf3f8023e01e0f17cb74c510ad228/dhcpv4/option_relay_agent_information.go#L10

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/u-root/uio/uio"
+)
+
+// NTPSuboptionSrvAddr is the value of NTP_SUBOPTION_SRV_ADDR according to RFC 5908.
+const NTPSuboptionSrvAddr byte = 1
```
pmazzini, line 11: For consistency, I think this should be a type like in https://github.com/insomniacslk/dhcp/blob/d4b282fe8d4cf3f8023e01e0f17cb74c510ad228/dhcpv4/option_relay_agent_information.go#L48

pmazzini (dismissed):
left some comments

rojer9-fb:
perhaps not in this PR, but should an accessor to dhcpv6.MessageOptions also be added?
smt. like `func (mo MessageOptions) NTP() []net.IP`

dhcpv6/option_ntp_server.go
insomniacslk: done

dhcpv6/option_ntp_server.go
insomniacslk: oh, damn, the two fields are 16 bits each, not 8. Is that what you mean?

dhcpv6/option_ntp_server.go
insomniacslk: I messed up reading the RFC. Fixing

dhcpv6/option_ntp_server.go
pmazzini: Not really, what I mean is that the sub option code can be != NTPSuboptionSrvAddr

insomniacslk:
almost there, now adapting tests to the new interface

insomniacslk:
all feedback addressed. @pmazzini @rojer9-fb please take a look

insomniacslk:
also added a TODO to add label compression detection in `rfc1035label`, which I will do after this PR

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/insomniacslk/dhcp/rfc1035label"
+	"github.com/u-root/uio/uio"
+)
+
+// NTPSuboptionCode defines the type for NTP sub-option codes.
+type NTPSuboptionCode uint16
+
+// NTPSuboption is the interface that every NTP suboption must implement.
+type NTPSuboption interface {
```
pmazzini, line 15: Do we need a new interface definition here?

I would use the same [Option interface](https://github.com/insomniacslk/dhcp/blob/d4b282fe8d4cf3f8023e01e0f17cb74c510ad228/dhcpv6/options.go#L10) we already have.

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,157 @@
...
+
+// NTPSuboptionSrvAddr is the value of NTP_SUBOPTION_SRV_ADDR according to RFC 5908.
+const (
+	NTPSuboptionSrvAddrCode = NTPSuboptionCode(1)
+	NTPSuboptionMCAddrCode  = NTPSuboptionCode(2)
+	NTPSuboptionSrvFQDNCode = NTPSuboptionCode(3)
+)
+
+// NTPSuboptions is a collection of NTPSuboption.
+type NTPSuboptions []NTPSuboption
```
pmazzini, line 94: Similar to the above comment, I would use the current [Options](https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/options.go#L116) type we already have.

As in the [vendor options](https://github.com/insomniacslk/dhcp/blob/d4b282fe8d4cf3f8023e01e0f17cb74c510ad228/dhcpv6/option_vendor_opts.go#L15).

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,157 @@
...
+
+// FromBytes parses a sequence of bytes into an NTPSuboptions object.
+func (o *NTPSuboptions) FromBytes(data []byte) error {
```
pmazzini, line 97: Where is this called?

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+	return buf.FinError()
+}
+
+// OptNTPServer is an option NTP server as defined by RFC 5908.
+type OptNTPServer struct {
+	Suboptions NTPSuboptions
+}
+
```
pmazzini, line 140: Are you missing a function to parse the `OptNTPServer`?

Also, I think this is a good place where to use the [FromBytesWithParser](https://github.com/insomniacslk/dhcp/blob/d4b282fe8d4cf3f8023e01e0f17cb74c510ad228/dhcpv6/options.go#L192) `Options` method.

dhcpv6/option_ntp_server.go
insomniacslk: good point. I started with a simpler interface, and then added pretty much the same methods. I'll remove it

dhcpv6/option_ntp_server.go
insomniacslk: added in options.go

insomniacslk: no, in this iteration, I just haven't pushed the changes yet because I'm busy with some other stuff :( will do later today

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,163 @@
...
+func (n *NTPSuboptionSrvFQDN) String() string {
+	l := rfc1035label.Labels(*n)
+	return fmt.Sprintf("Server FQDN: %s", l.String())
+}
+
+// NTPSuboptionSrvAddr is the value of NTP_SUBOPTION_SRV_ADDR according to RFC 5908.
+const (
+	NTPSuboptionSrvAddrCode = OptionCode(1)
```
pmazzini, line 78: I am wondering if we should convert the dhcpv6 [OptionCode](https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/types.go#L78) to an interface like we [currently have for dhcpv4](https://github.com/insomniacslk/dhcp/blob/d11e07c6523bdfd6c1c1cef0737d3096e2c6afcf/dhcpv4/types.go#L100).

In that way we will be able to have a different option code here like in the [dhcpv4 relay agent information](https://github.com/insomniacslk/dhcp/blob/d11e07c6523bdfd6c1c1cef0737d3096e2c6afcf/dhcpv4/option_relay_agent_information.go#L62).

If not, the `OptionCode.String()` method will be confusing.

Or maybe not worth at this point?

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,163 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse rfc1035 labels: %w", err)
+		}
+		// TODO according to rfc3315, this label must not be compressed.
+		// Need to add support for compression detection to the
+		// `rfc1035label` package in order to do that.
+		so := NTPSuboptionSrvFQDN(*l)
+		return &so, nil
+	default:
+		return nil, fmt.Errorf("unknown NTP sub-option code %d", code)
```
pmazzini, line 121: Maybe we should parse the rest as generic options here instead of returning an error?

dhcpv6/option_ntp_server.go
```diff
@@ -0,0 +1,163 @@
...
+
+// NTPSuboptionSrvAddr is the value of NTP_SUBOPTION_SRV_ADDR according to RFC 5908.
+const (
+	NTPSuboptionSrvAddrCode = OptionCode(1)
+	NTPSuboptionMCAddrCode  = OptionCode(2)
+	NTPSuboptionSrvFQDNCode = OptionCode(3)
+)
+
+// NTPSuboptions is a collection of NTPSuboption.
+type NTPSuboptions Options
```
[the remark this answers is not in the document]
insomniacslk: I think so: `Options` is specific to DHCPv6 options, while `NTPSuboptions` is specific to NTP sub-options. Which reminds me that I should probably also add `type NTPSuboption Option`.

If we don't distinguish, we can have ambiguities where an option has a different meaning based on the context. Is `Option(1)` an Option Client ID, or an NTP Server Address sub-option? That will depend on where the object is found, and while it won't break anything, that's where type systems help.
I am happy to just use `Options` if the RFC defines the NTP sub-options as just regular options though (haven't checked)

dhcpv6/dhcpv6message.go
```diff
@@ -288,6 +288,18 @@ func (mo MessageOptions) DHCP4oDHCP6Server() *OptDHCP4oDHCP6Server {
 	return nil
 }
 
+// NTP returns the NTP Server option as defined by RFC 5908.
+func (mo MessageOptions) NTP() *OptNTPServer {
```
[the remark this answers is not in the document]
insomniacslk: There are three different types of sub-options in an NTP Server option, hence returning the whole thing. I can think of two alternatives:
1) leave it as it is: return the option, and let the user filter from `o.Suboptions`
2) Add more functions, like `NTPSrvAddr() []net.IP`, `NTPMCAddr() []net.IP`, and `NTPSrvFQDN() string`

What do you think?

dhcpv6/option_ntp_server.go
insomniacslk: I think it's a good idea, but I would leave it for further discussion and to a separate PR

dhcpv6/option_ntp_server.go
insomniacslk: I am not sure what's right to do here. The original goal of the library was to be a scapy-like thing, where generic options and flexible parsing were first-class citizens. Then we decided that the library should do what's right for the user.

With `GenericOption` we break that a little bit, because we don't implement every possible option, and there are custom vendor options too, so we don't have a choice.
With the NTP Server option we know that only three sub-options are supported by the standard, and I think we should error out. If at any point a new option is added, I think we can reactively add support to it.

Thoughts?

dhcpv6/option_ntp_server.go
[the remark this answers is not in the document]
pmazzini: What I am worried about is that this is not consistent.

* The [dhcpv6 vendor option](https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/option_vendor_opts.go#L15) does not define a different options type.
* The [dhcpv4 relay option](https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/option_relay_agent_information.go#L10) does not define a different options type.
* The [bsdp vendor option](https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/bsdp/option_vendor_specific_information.go#L13) does not specify a different options type.

I think maybe we should define a different subset of Option codes instead (like suggested below).

dhcpv6/dhcpv6message.go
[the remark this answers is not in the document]
pmazzini: I am ok either way. I would go with option 2 and only implement the function for subtype 1, which is likely the one that going to be used the most. The user can get the rest anyway. However, I don't have a strong opinion.

dhcpv6/option_ntp_server.go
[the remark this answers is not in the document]
pmazzini: I would not create a new Options type and we can create a new subset of option types later on.

dhcpv6/option_ntp_server.go
pmazzini: I'd rather not be that strict and use the same approach we use for the [main options parsing](https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/options.go#L107).

In the same way we parse "main" options we know nothing about, I think that approach should be followed here.

If a new sub option is added in the future, this will break the parsing of the whole packet which is not nice (vs reactively adding support for it).

dhcpv6/dhcpv6message.go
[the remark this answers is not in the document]
insomniacslk: sounds good, I'll implement `NTPServers() []net.IP` which returns the sub-option 1 servers, and has a more descriptive name. Let me know if you prefer otherwise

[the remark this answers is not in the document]
insomniacslk: one more thing: the RFC states that a packet can have multiple NTP server options. I'll implement this function to iterate over all the NTP server options it finds, and concatenate the server addresses in order

insomniacslk:
rebased

pmazzini:
Except for the `NTPSuboptions` type which I think we should be removing, this looks good to me.

Do you have a strong opinion about keeping it?

insomniacslk:
> Except for the `NTPSuboptions` type which I think we should be removing, this looks good to me.
> 
> Do you have a strong opinion about keeping it?

No, I literally forgot to address that feedback, doing it now

insomniacslk:
done in the last push

pmazzini (approved):
lgtm

PR #437, opened 2021-08-10
use iana EntID

pmazzini:


pmazzini:
cc @insomniacslk

insomniacslk (commented):
The code looks good to me. Can we just call it `EnterpriseID`? I am all for short names, but EntID feels too ambiguous

pmazzini:
> Can we just call it EnterpriseID?

Ok, updated.

Issue #308, opened 2019-07-29
Use github.com/cenkalti/backoff for retries


rjoleary:
Use github.com/cenkalti/backoff to replace the backoff algorithm in https://github.com/insomniacslk/dhcp/pull/296.

The backoff package is well tested and gives the user more control over which backoff method and parameters to use.

rjoleary:
I can work on this.

insomniacslk:
Thanks for working on this! On a phone now and can't remember what was wrong with the current retry system, are you addressing a particular problem that hit you?

rjoleary:
There'd be two main benefits for using this library. First, it would give more control/options over how the backoff is performed (especially with the BackOff interface). Secondly, this backoff library is well tested.

rjoleary:
It doesn't look like this repository has a vendor directory. How would you want to make the dependency on github.com/cenkalti/backoff ?

hugelgupf:
I'd just make it without managing it, same as with u-root so far.

When u-root is ready we should go to modules?

pmazzini:
> I can work on this.

@rjoleary Are you looking at this?

heinz-boehmer:
I'll take over this.

PR #439, opened 2021-08-11
nclient6: Use cenkalti/backoff/v4 library to replace backoff algorith…

heinz-boehmer:
…m in [#296](https://github.com/insomniacslk/dhcp/pull/296).


dhcpv6/nclient6/client.go
```diff
@@ -456,7 +457,7 @@ func (c *Client) SendAndRead(ctx context.Context, dest *net.UDPAddr, msg *dhcpv6
 			case <-c.done:
 				return ErrNoResponse
 
-			case <-time.After(timeout):
+			case <-time.After(c.timeout):
```
rjoleary, line 460: I think this doesn't quite work since `c.timeout` remains constant throughout all the iterations instead of doubling each iteration.

dhcpv6/nclient6/client.go
```diff
@@ -470,7 +471,7 @@ func (c *Client) SendAndRead(ctx context.Context, dest *net.UDPAddr, msg *dhcpv6
 				}
 			}
 		}
-	})
+	}, backoff.WithMaxRetries(backoff.NewExponentialBackOff(), c.retry))
```
rjoleary, line 474: You should probably also do a `backoff.WithContext(` here as well.

heinz-boehmer: Done.

dhcpv6/nclient6/client.go
heinz-boehmer: Changed this so that the timeout doubles every iteration.

pmazzini:
Fixes #308

Issue #454, opened 2021-12-13
panic parsing UDP packet


smira:
This issue is random, but looks like the library can panic when it receives some (malformed) packet. I don't have the packet dump unfortunately, as the panic happens in the PID 1 process and it crashes whole node.

The panic itself seems to be coming from the line https://github.com/insomniacslk/dhcp/blob/465dd6c35f6c983206a5bc21fbfff4819ce21eb0/dhcpv4/nclient4/ipv4.go#L213, and the panic message is: `panic: runtime error: slice bounds out of range [2:0]`

Looking at the call graph, it seems that the panic originates from the `recieveLoop()` function which runs in a goroutine without panic handler. I'm not sure if the panic got fixed with some later changes to the library, but I propose to add a panic handler to the `receiveLoop` so that it doesn't crash whole Go process. (This can'be be solved outside of the library, as the goroutine is created inside the library). Does this make sense? (I'm happy to work on a PR)

smira:
See also: https://github.com/talos-systems/talos/issues/4567

pmazzini:
Closing after #455

PR #455, opened 2021-12-13
fix: check IP/UDP header size before trying to access it

smira:
This should fix the panic we've seen with the malformed packets on the
wire. Remaining buffer size should be checked before trying to access
the data, otherwise Go might panic on out of bounds slice operation.

Signed-off-by: Andrey Smirnov <andrey.smirnov@talos-systems.com>


smira:
Related to #454

hugelgupf:
Thanks for the fix!

hugelgupf (approved):

PR #456, opened 2022-01-18
dhcp/dhcpv4/nclient4: get it to build on darwin

rminnich:
Signed-off-by: Ronald G. Minnich <rminnich@gmail.com>


hugelgupf:
I counter with https://github.com/insomniacslk/dhcp/pull/457

dhcpv4/nclient4/conn_darwin.go
```diff
@@ -0,0 +1,143 @@
+// Copyright 2018 the u-root Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build go1.12
```
hugelgupf, line 5: I feel like a build tag on the other file instead of copy & pasting it would be preferred? 

https://github.com/insomniacslk/dhcp/pull/457

hugelgupf (changes requested):

dhcpv4/client4/client_darwin.go
```diff
@@ -0,0 +1,3 @@
+// Package client4 is deprecated. Use "nclient4" instead.
+package client4
```
hugelgupf, line 2: I would advise against making an empty package file just to make a package build. A package that is not supported on darwin at all should just not build on darwin. If you don't plan to use client4 and you use nclient4 instead, just leave it unsupported and not building.

dhcpv4/bsdp/client_darwin.go
```diff
@@ -0,0 +1 @@
+package bsdp
```
hugelgupf, line 1: Same comment as client4.

dhcpv4/async/client_darwin.go
```diff
@@ -0,0 +1,2 @@
+package async
```
hugelgupf, line 1: Same comment as client4.

hugelgupf (changes requested):

rminnich:
yeah you're right, will try again, the damn thing just kept refusing to build, and only the empty package fixed it,  but I'll try again

insomniacslk:
I wonder at which point we should just delete client{4,6} and just stick with nclient{4,6}. I can't remember what's missing on our end to migrate, but I'll CC @tfg13 who's in charge for that code

rminnich:
> I wonder at which point we should just delete client{4,6} and just stick with nclient{4,6}. I can't remember what's missing on our end to migrate, but I'll CC @tfg13 who's in charge for that code

you have still got stuff in here that depends on client4. 

That's the only reason I did not just delete it. If you're willing to remove stuff like netboot, this can work.

dhcpv4/client4/client_darwin.go
hugelgupf: Okay, a function that just returns unsupported is also undesired... The 2 desired states from a user point of view are: it either does not compile (compiler says unsupported!) or it compiles and has full support. Why are you trying to make client4 compile if you're not trying to use it? I think just making nclient4 compile is a good enough goal if that's what you're trying to use?

hugelgupf: Compiling but having no functions and compiling but only returning errors are silent failures from a user perspective. It's like Python - it can be interpreted, but you don't know what's wrong until you run it. I'd prefer if we remain in the "not compilable" state for things that we do not support. I guess "compiling but having no functions" is slightly better than "compiling but only returning errors" from that perspective, but I still don't get why you're trying to do it.

rminnich: yeah. this is a hack. I'll let somebody else work out how to do this. This package is wearying.

PR #458, opened 2022-01-21
Change unicast connection of DHCPv4 client as a raw socket

yaocw2020:
If the network manager in the operating system uses DHCP to obtain an IP
for the network card, it will occupy the DHCP client port. In this case
the release function will fail because the unicast connection with UDP
socket need the same port. Use a raw udp socket instead of a datagram
udp socket to solve it.

Signed-off-by: yaocw2020 <yaocanwu@gmail.com>


yaocw2020:
@pmazzini  Could you help review this pr? Moreover, are the unit tests mandatory? It seems hard to add the unit tests for the arp request in this testing environment.

yaocw2020:
@insomniacslk @pmazzini @hugelgupf Could you help review this pr? By the way, there is something abnormal about the integration-tests(dhcpv6)

PR #459, opened 2022-01-21
remove client4/client6 and anything using them

rminnich:
Remove long-deprecated client4/client6 packages that no longer build
on bsd or darwin. This is a long-standing problem.

Remove all things depending on them, which have no users we know of.

Signed-off-by: Ronald G. Minnich <rminnich@gmail.com>


insomniacslk:
CC @tfg13 - we only depend on it for OSF. This should be migrated to nclient{6,4}; tested through our OSF pipeline; released to our customers; and once confirmed that it's working well it can be merged. Also CC @bigvalen as the owner of our internal code that uses this. I can help with the above steps if needed

pmazzini:
Related https://github.com/pinterest/bender/pull/39

pmazzini:
For awareness, the `netboot` package being removed here is used in `r-root`.

https://github.com/u-root/u-root/blob/main/cmds/boot/fbnetboot/main.go#L28

insomniacslk:
if nothing is using it, and client{4,6} are the old and deprecated ones, yes, I think netboot should go

pmazzini:
The `netboot` package is used at least by `lol2`.

insomniacslk:
ack. @rminnich see above

PR #461, opened 2022-02-21
dhcpv4/nclient4: switch to github.com/mdlayher/packet

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

I made and tagged a new library https://github.com/mdlayher/packet for Linux only, see the README for rationale.

Unfortunately this means we also drop support for *BSD again, but AFAICT that code path in package raw was actually broken because the "LinuxSockDGRAM" option never worked on BSD.


dhcpv4/nclient4/conn_unix.go
```diff
@@ -13,22 +12,18 @@ import (
...
+	"golang.org/x/sys/unix"
 )
 
-var (
-	// BroadcastMac is the broadcast MAC address.
-	//
-	// Any UDP packet sent to this address is broadcast on the subnet.
-	BroadcastMac = net.HardwareAddr([]byte{255, 255, 255, 255, 255, 255})
-)
+// BroadcastMac is the broadcast MAC address.
```
mdlayher, line 20: Ah, my editor applies gofumpt. I can revert if you like.

insomniacslk:
CC @hugelgupf since this touches nclient

mdlayher:
Hi there @insomniacslk and @hugelgupf! Any further thoughts? Like I mentioned, I think this was already broken on BSD so I would encourage moving forward with this as an improvement. Dedicated work will be necessary to make a production ready packet sockets equivalent for *BSD, and I lack that expertise at the moment.

pmazzini (approved):

rminnich:
I really wish we would not move to a "linux only" world. I've been battling to get this package working on darwin for years, and it's a shame to see a move away from portability.

mdlayher:
See my first comment:
> AFAICT that code path in package raw was actually broken because the "LinuxSockDGRAM" option never worked on BSD.

But given the lack of movement here, I'm going to close this out and let you all reevaluate. Just note that raw is unmaintained. Thanks.

PR #463, opened 2022-03-31
Extending parse_circuit_id and adding support for Ciena

name29:
Extending support for more interface types in parse_circuit_id
Adding support for DHCPv6 for Ciena

Signed-off-by: Emanuele Fia <name29@fb.com>


dhcpv6/option_userclass_test.go
```diff
@@ -91,3 +91,13 @@ func TestOptUserClassString(t *testing.T) {
 		"String() should contain the list of user classes",
 	)
 }
+
+func TestOptUserClassParseOptUserClassCiena(t *testing.T) {
```
pmazzini, line 95: What is this additional test for? Is it covering something we were not before?

dhcpv6/dhcpv6_test.go
```diff
@@ -133,6 +133,31 @@ func TestFromAndToBytes(t *testing.T) {
 	expected := [][]byte{
 		{01, 0xab, 0xcd, 0xef, 0x00, 0x00, 0x00, 0x00},
 		[]byte("0000\x00\x01\x00\x0e\x00\x01000000000000"),
+		{
```
pmazzini, line 136: What is this additional test for? Is it covering something we were not before?

dhcpv6/option_userclass.go
```diff
@@ -50,5 +50,9 @@ func ParseOptUserClass(data []byte) (*OptUserClass, error) {
 		len := buf.Read16()
 		opt.UserClasses = append(opt.UserClasses, buf.CopyN(int(len)))
 	}
-	return &opt, buf.FinError()
+	var err = buf.FinError()
```
pmazzini, line 53: Modifying the return error message of the user class option does not seem quite related to the topic of the PR.

While I am personally not against it, this makes it inconsistent with what we log for other options.

Why specifically the user class? Can all these "logging improvements" be discussed in a separate PR and in this one only focus on the ZTP changes?

dhcpv6/dhcpv6.go
```diff
@@ -75,7 +75,7 @@ func RelayMessageFromBytes(data []byte) (*RelayMessage, error) {
 	}
 	// TODO: fail if no OptRelayMessage is present.
 	if err := d.Options.FromBytes(buf.Data()); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Unable to parse Options: %v", err)
```
pmazzini, line 78: same as my comment to the user class error return changes

dhcpv6/dhcpv6.go
```diff
@@ -85,7 +85,7 @@ func FromBytes(data []byte) (DHCPv6, error) {
 	buf := uio.NewBigEndianBuffer(data)
 	messageType := MessageType(buf.Read8())
 	if buf.Error() != nil {
-		return nil, buf.Error()
+		return nil, fmt.Errorf("Unable to parse MessageType: %v", buf.Error())
```
pmazzini, line 88: same as my comment to the user class error return changes

dhcpv6/options.go
```diff
@@ -208,10 +208,12 @@ func (o *Options) FromBytesWithParser(data []byte, parser OptionParser) error {
 		// Consume, but do not Copy. Each parser will make a copy of
 		// pertinent data.
 		optData := buf.Consume(length)
-
+		if err := buf.Error(); err != nil {
```
pmazzini, line 211: same, lets discuss error messages on a separate PR and in this one focus on ZTP which if I am not mistake is the one more pressing for you

dhcpv6/ztpv6/parse_remote_id.go
```diff
@@ -8,12 +8,32 @@ import (
 	"github.com/insomniacslk/dhcp/dhcpv6"
 )
 
-var (
+var circuitRegexs = []*regexp.Regexp{
```
pmazzini, line 11: Do you actually need all of the regexes? Are you expecting all of them to have v6 support and be present in a relay? If not, can we only add the ones we are sure about?

Why do we end up having more regexes than v4? Do any of the 2 regexes at the top overlap with some of the newly added ones?

dhcpv6/ztpv6/parse_vendor_options.go
```diff
@@ -66,6 +69,27 @@ func ParseVendorData(packet dhcpv6.DHCPv6) (*VendorData, error) {
...
+		// For Ciena the iana code is 1271
+		// The product type is a number that maps to a Ciena product
+		// The type is used to identified different subtype of the product.
+		// An example can be ‘1271-23422Z11-123’.
+		case strings.HasPrefix(d, strconv.Itoa(int(iana.EnterpriseIDCienaCorporation))):
+			v := strings.Split(d, "-")
+			if len(v) < 3 {
+				return nil, errVendorOptionMalformed
+			}
+			duid := packet.(*dhcpv6.Message).Options.ClientID()
```
pmazzini, line 84: This will panic if the parameter is a relay instead of a message.

I wonder if we should change the function parameter to take a message instead.

dhcpv6/ztpv6/parse_vendor_options.go
```diff
@@ -66,6 +69,27 @@ func ParseVendorData(packet dhcpv6.DHCPv6) (*VendorData, error) {
...
+			if duid.Type != dhcpv6.DUID_EN {
```
pmazzini, line 85: Is this check actually needed? Or is it only to provide additional certainty?

dhcpv6/ztpv6/parse_vendor_options_test.go
```diff
@@ -99,3 +100,50 @@ func TestParseVendorDataWithVendorClass(t *testing.T) {
 		})
 	}
 }
+
+func TestParseVendorDataWithClientId(t *testing.T) {
```
pmazzini, line 104: Can this test be unified with the previous one? It is also parsing the vendor class after all.

dhcpv6/dhcpv6_test.go
name29: I wanted to add an example of packet generated by some a device to ensure that it can be parsed correctly.

If you think that this doesn't add value,  I can remove it :-)

dhcpv6/option_userclass.go
name29: I can move these changes out.

I just added the error messages that helped me troubleshooting some issues I had at the beginning.

dhcpv6/option_userclass_test.go
name29: Stale experiment, let me remove it :-)

dhcpv6/ztpv6/parse_remote_id.go
name29: Ah, I forgot to consolidate the first one with the 4th.

Looking again at this, I don't think I need any of the changes here, so I'm just going to remove it.

dhcpv6/ztpv6/parse_vendor_options.go
name29: Specifications from the vendor is that this parameter is always DUID_EN, this is why I have enforced it here, I'm also not sure what is going to happen if type != DUID_EN and I try to access duid.EnterpriseIdentifier ?

dhcpv6/ztpv6/parse_vendor_options_test.go
name29: Will try to merge them :-)

pmazzini:
Thanks for cleaning up the PR.

dhcpv6/ztpv6/parse_vendor_options.go
```diff
@@ -66,6 +69,27 @@ func ParseVendorData(packet dhcpv6.DHCPv6) (*VendorData, error) {
...
+		// For Ciena the iana code is 1271
+		// The product type is a number that maps to a Ciena product
+		// The type is used to identified different subtype of the product.
+		// An example can be ‘1271-23422Z11-123’.
+		case strings.HasPrefix(d, strconv.Itoa(int(iana.EnterpriseIDCienaCorporation))):
+			v := strings.Split(d, "-")
+			if len(v) < 3 {
+				return nil, errVendorOptionMalformed
+			}
+			innerMessage, err := packet.GetInnerMessage()
```
pmazzini, line 84: The only thing I am not convinced about is how we get the DUID.

In here we get the inner message but we don't do that for the opt16 and opt17 above, so it is not very consistent.

Maybe we should go back to doing `packet.(*dhcpv6.Message)`. Sorry for changing my mind here.

pmazzini:
Maybe it would be good to update the README to include Ciena:
https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/ztpv6/README.md

dhcpv6/ztpv6/parse_vendor_options.go
name29: I personally prefer more this way as we ensure that we are in the inner message 😅

I was tempted to modify also the code that is getting opt16/opt17 but I didn't wanted to break too many things

pmazzini: Yeah, the issue now is that how we get the options is not consistent.

pmazzini (approved):

Issue #466, opened 2022-05-12
RFC 1035 Labels: Support Partial Domain Names as per RFC 4704 Section 4.2


rtpt-erikgeiser:
I found that my VM running Windows 10 (Version 1909) in default configuration sends DHCPv6 solicit messages with an FQDN option where the label only contains a length byte and the name without a trailing zero-length byte (note that the zero-byte after `win10vm` belongs to the vendor class option):

![image](https://user-images.githubusercontent.com/70747329/168074889-98b2b63c-d277-41f7-af31-67493f73feaa.png)

When this label is parsed in `labelsFromBytes()`, the bottom else branch is executed first such that `label` holds `win10vm`. Then, however, the whole data is already consumed, `pos >= len(buf)` is false and the function returns without appending `label` to `labels`.

https://github.com/insomniacslk/dhcp/blob/1ca156eafb9f20f7884eddc2cf610bade5dfb560/rfc1035label/label.go#L93-L140

As far as I understand RFC 1035, this library implements label parsing correctly. In [this section of RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035#section-3.1) it is explicitly stated that "Since every domain name ends with the null label of the root, a domain name is terminated by a length byte of zero". While Wireshark does parse the supposedly malformed label, it calls it "Partial domain name". I don't know if this is normal or if this is due to the missing zero-length byte.

It would be great if this case would also be supported, as this seems to be the default behavior of Windows 10. I've added a PR (#467) with a failing test as a starting point in but I don't really know how the fix should be implemented correctly.

For anyone running into the same issue, I have implemented the following workaround:
```go
if len(fqdn.DomainName.Labels) == 0 {
	rawLabel := fqdn.DomainName.ToBytes()
	if len(rawLabel) > 2 && int(rawLabel[0]) == (len(rawLabel)-1) {
		fqdn.DomainName.Labels = append(fqdn.DomainName.Labels, string(rawLabel[1:]))
	}
}
```

PR #467, opened 2022-05-12
Add Support for Labels Containing Partial Domain Names (RFC 4704 Section 4.2)

rtpt-erikgeiser:
This PR only contains a failing test case for #466. I don't know enough about RFC 1035 (especially compressed ones) to be comfortable implementing a solution that does not break anything.

### DIFF
--- a/rfc1035label/label.go
+++ b/rfc1035label/label.go
@@ -104,8 +104,6 @@ func labelsFromBytes(buf []byte) ([]string, error) {
 			// domain name field as per RFC 4704 Section 4.2
 			if label != "" {
 				labels = append(labels, label)
-
-				return labels, nil
 			}
 
 			break
--- a/rfc1035label/label.go
+++ b/rfc1035label/label.go
@@ -100,6 +100,14 @@ func labelsFromBytes(buf []byte) ([]string, error) {
 
 	for {
 		if pos >= len(buf) {
+			// interpret label without trailing zero-length byte as a partial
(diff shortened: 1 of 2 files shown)


rtpt-erikgeiser:
With this PR, partial domain name labels are now correctly decoded. However, the information whether the label is partial or not is not exposed to the user and marshaling discards the information.

For a complete solution it would probably be necessary to save the information whether it is partial or not for each label and present this to the user somehow. Therefore, I think it is best just to be able to decode partial labels first.

rfc1035label/label.go
```diff
@@ -100,6 +100,14 @@ func labelsFromBytes(buf []byte) ([]string, error) {
 
 	for {
 		if pos >= len(buf) {
+			// interpret label without trailing zero-length byte as a partial
+			// domain name field as per RFC 4704 Section 4.2
+			if label != "" {
+				labels = append(labels, label)
+
+				return labels, nil
```
pmazzini, line 108: Is this return needed here?

pmazzini:
@rtpt-erikgeiser Sorry for the late review. I left a comment.

rfc1035label/label.go
erikgeiser: Looking at it now it should work just as well without it. I can't run the tests right no because I'm on the phone but if they pass without the return we can get rid of it.

rtpt-erikgeiser:
@pmazzini I removed the superfluous return statement.

pmazzini (approved):

PR #469, opened 2022-07-08
dhcpv4: support RENEW requests and option deletion

twelho:
When using `nclient4.RequestFromOffer` with an existing offer to issue lease renewals, `OptionRequestedIPAddress` and `OptionServerIdentifier` must be cleared to comply with [RFC 2131, section 4.3.2] (RENEW). Updating these with modifiers using `dhcpv4.WithOption` does work, but setting the IP addresses to `nil` causes `Options` keep them which results in an invalid DHCP packet that has entries with without values. This fix implements a new `WithoutOption` modifier that can be used to clear those options.

Additionally, it also enables native issuing of renew requests with a new `nclient4.Client.Renew` method using the aforementioned functionality.

[RFC 2131, section 4.3.2]: https://datatracker.ietf.org/doc/html/rfc2131#section-4.3.2

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -382,6 +382,13 @@ func (d *DHCPv4) GetOneOption(code OptionCode) []byte {
 	return d.Options.Get(code)
 }
 
+// DeleteOption deletes an existing option with the given option code.
+func (d *DHCPv4) DeleteOption(code OptionCode) {
+	if d.Options != nil {
+		d.Options.Del(code)
+	}
+}
+
 // UpdateOption replaces an existing option with the same option code with the
 // given one, adding it if not already present.
 func (d *DHCPv4) UpdateOption(opt Option) {
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -231,6 +231,7 @@ func NewInform(hwaddr net.HardwareAddr, localIP net.IP, modifiers ...Modifier) (
 }
 
 // NewRequestFromOffer builds a DHCPv4 request from an offer.
+// It assumes the SELECTING state by default, see Section 4.3.2 in RFC 2131 for more details.
 func NewRequestFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) {
 	return New(PrependModifiers(modifiers,
 		WithReply(offer),
@@ -245,6 +246,21 @@ func NewRequestFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) 
 			OptionDomainName,
 			OptionDomainNameServer,
 		),
+	)...)
+}
+
+// NewRenewFromOffer builds a DHCPv4 RENEW-style request from an offer. RENEW requests have minor
+// changes to their options compared to SELECT requests as specified by RFC 2131, section 4.3.2.
+func NewRenewFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) {
+	return NewRequestFromOffer(offer, PrependModifiers(modifiers,
+		// The server identifier option must not be filled in
+		WithoutOption(OptionServerIdentifier),
+		// The requested IP address must not be filled in
+		WithoutOption(OptionRequestedIPAddress),
+		// The client IP must be filled in with the IP offered to the client
+		WithClientIP(offer.YourIPAddr),
+		// The renewal request must use unicast
+		WithBroadcast(false),
 	)...)
 }
 
--- a/dhcpv4/modifiers_test.go
+++ b/dhcpv4/modifiers_test.go
@@ -44,6 +44,17 @@ func TestWithOptionModifier(t *testing.T) {
 	require.Equal(t, "slackware.it", dnOpt)
 }
 
+func TestWithoutOptionModifier(t *testing.T) {
+	d, err := New(
+		WithOption(OptDomainName("slackware.it")),
+		WithoutOption(OptionDomainName),
+	)
+	require.NoError(t, err)
+
+	require.False(t, d.Options.Has(OptionDomainName))
+	require.Equal(t, "", d.DomainName())
+}
+
 func TestUserClassModifier(t *testing.T) {
 	d, err := New(WithUserClass("linuxboot", false))
 	require.NoError(t, err)
--- a/dhcpv4/nclient4/client.go
+++ b/dhcpv4/nclient4/client.go
@@ -493,6 +493,7 @@ func (e *ErrNak) Error() string {
 }
 
 // RequestFromOffer sends a Request message and waits for an response.
+// It assumes the SELECTING state by default, see Section 4.3.2 in RFC 2131 for more details.
 func (c *Client) RequestFromOffer(ctx context.Context, offer *dhcpv4.DHCPv4, modifiers ...dhcpv4.Modifier) (*Lease, error) {
 	// TODO(chrisko): should this be unicast to the server?
 	request, err := dhcpv4.NewRequestFromOffer(offer, dhcpv4.PrependModifiers(modifiers,
--- a/dhcpv4/nclient4/lease.go
+++ b/dhcpv4/nclient4/lease.go
@@ -3,6 +3,7 @@
 package nclient4
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"time"
@@ -35,4 +36,42 @@ func (c *Client) Release(lease *Lease, modifiers ...dhcpv4.Modifier) error {
 		c.logger.PrintMessage("sent message:", req)
 	}
 	return err
+}
+
+// Renew sends a DHCPv4 request to the server to renew the given lease. The renewal information is
+// sourced from the initial offer in the lease, and the ACK of the lease is updated to the ACK of
+// the latest renewal. This avoids issues with DHCP servers that omit information needed to build a
+// completely new lease from their renewal ACK (such as the Windows DHCP Server).
+func (c *Client) Renew(ctx context.Context, lease *Lease, modifiers ...dhcpv4.Modifier) error {
+	if lease == nil {
+		return fmt.Errorf("lease is nil")
+	}
+
+	request, err := dhcpv4.NewRenewFromOffer(lease.Offer, dhcpv4.PrependModifiers(modifiers,
+		dhcpv4.WithOption(dhcpv4.OptMaxMessageSize(MaxMessageSize)))...)
+	if err != nil {
+		return fmt.Errorf("unable to create a request: %w", err)
+	}
+
+	// Servers are supposed to only respond to Requests containing their server identifier,
+	// but sometimes non-compliant servers respond anyway.
+	// Clients are not required to validate this field, but servers are required to
+	// include the server identifier in their Offer per RFC 2131 Section 4.3.1 Table 3.
+	response, err := c.SendAndRead(ctx, c.serverAddr, request, IsAll(
+		IsCorrectServer(lease.Offer.ServerIdentifier()),
+		IsMessageType(dhcpv4.MessageTypeAck, dhcpv4.MessageTypeNak)))
+	if err != nil {
+		return fmt.Errorf("got an error while processing the request: %w", err)
+	}
+	if response.MessageType() == dhcpv4.MessageTypeNak {
+		return &ErrNak{
+			Offer: lease.Offer,
+			Nak:   response,
+		}
+	}
+
+	// Update the ACK of the lease with the ACK of the latest renewal
+	lease.ACK = response
+
+	return nil
 }
--- a/dhcpv4/nclient4/lease_test.go
+++ b/dhcpv4/nclient4/lease_test.go
@@ -1,4 +1,4 @@
-// this tests nclient4 with lease and release
+// this tests nclient4 with lease, renew and release
 
 package nclient4
 
@@ -236,6 +236,14 @@ func (sll *testServerLeaseList) runTest(t *testing.T) {
 		sll.lastTestSvrErrLock.RLock()
 		keepgoing := chkerr(err, sll.lastTestSvrErr, l.ShouldFail, t)
 		sll.lastTestSvrErrLock.RUnlock()
+
+		if keepgoing {
+			err = clnt.Renew(context.Background(), lease)
+			sll.lastTestSvrErrLock.RLock()
+			keepgoing = chkerr(err, sll.lastTestSvrErr, l.ShouldFail, t)
+			sll.lastTestSvrErrLock.RUnlock()
+		}
+
 		if keepgoing {
 			err = clnt.Release(lease)
 			//this sleep is to make sure release is handled by server
--- a/dhcpv4/dhcpv4_test.go
+++ b/dhcpv4/dhcpv4_test.go
@@ -266,6 +266,43 @@ func TestDHCPv4NewRequestFromOfferWithModifier(t *testing.T) {
 	require.Equal(t, MessageTypeRequest, req.MessageType())
 }
 
+func TestDHCPv4NewRenewFromOffer(t *testing.T) {
+	offer, err := New()
+	require.NoError(t, err)
+	offer.SetBroadcast()
+	offer.UpdateOption(OptMessageType(MessageTypeOffer))
+	offer.UpdateOption(OptServerIdentifier(net.IPv4(192, 168, 0, 1)))
+	offer.UpdateOption(OptRequestedIPAddress(net.IPv4(192, 168, 0, 1)))
+	offer.YourIPAddr = net.IPv4(192, 168, 0, 1)
+
+	// RFC 2131: RENEW-style requests will be unicast
+	var req *DHCPv4
+	req, err = NewRenewFromOffer(offer)
+	require.NoError(t, err)
+	require.Equal(t, MessageTypeRequest, req.MessageType())
+	require.Nil(t, req.GetOneOption(OptionServerIdentifier))
+	require.Nil(t, req.GetOneOption(OptionRequestedIPAddress))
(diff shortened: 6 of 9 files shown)


smira (approved):

twelho:
@insomniacslk @pmazzini could I get a review? This is needed for https://github.com/siderolabs/talos/pull/5897 which is lined up for the 1.2 release of Talos.

pmazzini:
I really don't like making the `Update` method removing an option, from its name it is not clear it will do this.
I also don't like setting the value of options to `nil`, it looks hacky.

How about having a a modifier removing those options or a `NewRenewFromOffer`?

twelho:
I can add a `Del` function to `Options` (similar to `dchpv6`) and corresponding a modifier for it. It might be a good idea to also provide `NewRenewFromOffer` as a shorthand. I will update this PR shortly.

However, it is still possible to cause the library to send invalid DHCP packets with `nil` values in `Option`s. In my opinion this should be prevented somehow. Thoughts?

hugelgupf:
It need not be prevented. Options that have no value are perfectly valid,
and in Go it is often non-idiomatic and non-obvious to treat nil and
[]byte{} as different (both have length 0).

DHCPv4 spec afaik only has the Pad and End options without value, but it
doesn't keep a vendor from specifying and implementing an option without a
value in vendor-specific options, for example. (DHCPv6 has Rapid Commit,
for example.)

twelho:
Hmm, then really what seems to be missing is some kind of input validation for each of the known options as specified by [RFC 2132](https://datatracker.ietf.org/doc/html/rfc2132) in the various `Opt*` functions. That will have to be a separate PR if such functionality is desired.

twelho:
Alright, inspired by @pmazzini's suggestions I present a new approach.

- https://github.com/insomniacslk/dhcp/pull/469/commits/7cd86e1b80be3a3ca310b447519e887b94a2a240 adds the required functionality and a dedicated modifier for option deletion.
- https://github.com/insomniacslk/dhcp/pull/469/commits/575f18f584ca106c1af28f533d245629092f6403 essentially implements `NewRenewFromOffer`, but in a more generic way that covers all four variants of a DHCPREQUEST as specified by RFC 2131.

Could I get a re-review? I will implement unit tests *after* the design has been accepted, but before this PR is merged. Also, if you prefer to have the two functionalities in separate PRs, just let me know.

dhcpv4/dhcpv4.go
```diff
@@ -230,22 +230,48 @@ func NewInform(hwaddr net.HardwareAddr, localIP net.IP, modifiers ...Modifier) (
 	)...)
 }
 
-// NewRequestFromOffer builds a DHCPv4 request from an offer.
-func NewRequestFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) {
```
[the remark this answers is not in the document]
twelho: Yes we can, I wasn't sure how stable the API is here, but that makes sense. I'll still however update the comment to indicate that this explicitly issues a request in the `selecting` state.

dhcpv4/nclient4/client.go
```diff
@@ -492,10 +493,10 @@ func (e *ErrNak) Error() string {
 	return "server rejected request with Nak"
 }
 
-// RequestFromOffer sends a Request message and waits for an response.
-func (c *Client) RequestFromOffer(ctx context.Context, offer *dhcpv4.DHCPv4, modifiers ...dhcpv4.Modifier) (*Lease, error) {
```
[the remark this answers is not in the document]
twelho: Yes, I'll undo the modifications here and update the comment as above.

twelho:
> How about making the renew a method of the lease?
> 
> https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/nclient4/lease.go#L25

This make sense. I don't have time to proto this today anymore, but will look into it implementing it that way tomorrow. How about the approach to clearing options, specifically https://github.com/insomniacslk/dhcp/commit/7cd86e1b80be3a3ca310b447519e887b94a2a240? It is now a bit tangentially related, but could still be useful (since `dhcpv6` has it).

twelho:
8dc422bf899950c56f65968ea86966a0e964c51d now contains the approach where `Renew` is a method of `Client` next to `Release`. This is the smallest changeset I could come up with, the architecture is not very flexible to enable implementing this without almost duplicating `RequestFromOffer` due to the tight coupling between `request` and `offer` in that method.

twelho:
I've managed to integrate `Renew` with the existing `RequestFromOffer` method. The functionality should be the same, but this approach avoids the code duplication and maintenance overhead. Will now write some unit tests for both this and the option removal.

twelho:
Alright, unit tests added. This should now be ready for final review.

dhcpv4/dhcpv4.go
```diff
@@ -230,7 +230,7 @@ func NewInform(hwaddr net.HardwareAddr, localIP net.IP, modifiers ...Modifier) (
 	)...)
 }
 
-// NewRequestFromOffer builds a DHCPv4 request from an offer.
+// NewRequestFromOffer builds a DHCPv4 request in the SELECT state from an offer.
```
pmazzini, line 233: Do we need this clarification? I am not sure if this makes things more clear or more confusing.

dhcpv4/nclient4/client.go
```diff
@@ -492,7 +492,7 @@ func (e *ErrNak) Error() string {
 	return "server rejected request with Nak"
 }
 
-// RequestFromOffer sends a Request message and waits for an response.
+// RequestFromOffer sends a Request message in the SELECT state and waits for a response.
```
pmazzini, line 495: Same: Do we need this clarification? I am not sure if this makes things more clear or more confusing.

dhcpv4/nclient4/lease_test.go
```diff
@@ -236,13 +236,21 @@ func (sll *testServerLeaseList) runTest(t *testing.T) {
...
-			err = clnt.Release(lease)
-			//this sleep is to make sure release is handled by server
-			time.Sleep(time.Second)
+			lease, err = clnt.Renew(context.Background(), lease.Offer)
 			sll.lastTestSvrErrLock.RLock()
-			chkerr(err, sll.lastTestSvrErr, l.ShouldFail, t)
+			keepgoing = chkerr(err, sll.lastTestSvrErr, l.ShouldFail, t)
 			sll.lastTestSvrErrLock.RUnlock()
+
+			if keepgoing {
```
pmazzini, line 246: Do we need this as a nested if? Can it not go outside of the previous if?

dhcpv4/dhcpv4.go
```diff
@@ -262,12 +262,12 @@ func NewReplyFromRequest(request *DHCPv4, modifiers ...Modifier) (*DHCPv4, error
 }
 
 // NewReleaseFromACK creates a DHCPv4 Release message from ACK.
-// default Release message without any Modifer is created as following:
+// default Release message without any Modifiers is created as following:
```
pmazzini, line 265: All these comment updates are unrelated to the RENEW support. Maybe they should be a different PR.

dhcpv4/nclient4/lease.go
```diff
@@ -18,17 +18,17 @@ type Lease struct {
 	CreationTime time.Time
 }
 
-// Release send DHCPv4 release messsage to server, based on specified lease.
-// release is sent as unicast per RFC2131, section 4.4.4.
-// Note: some DHCP server requries of using assigned IP address as source IP,
-// use nclient4.WithUnicast to create client for such case.
+// Release sends a DHCPv4 release message to server, based on the specified lease.
```
pmazzini, line 21: Same: All these comment updates are unrelated to the RENEW support. Maybe they should be a different PR.

dhcpv4/dhcpv4.go
twelho: [RFC 2131](https://datatracker.ietf.org/doc/html/rfc2131#section-4.3.2) states that there are four different DHCPREQUEST types, this function always constructs a request in the SELECT state. This tripped me up when trying to do RENEWs with it, thus I've added the comment.

dhcpv4/nclient4/client.go
twelho: This calls `NewRequestFromOffer`, which does SELECT requests by default. I can however change the wording to be a bit more subtle.

dhcpv4/dhcpv4.go
twelho: I can however change the wording to be a bit more subtle.

dhcpv4/nclient4/lease_test.go
twelho: I think what was intended here is to stop progressing if an error occurs, but make sure to always unlock the mutex, thus the nested `if`s. I can however refactor it to use `continue`s instead to avoid the nesting.

dhcpv4/dhcpv4.go
twelho: Sure, I can remove them. Wanted to just improve the wording as I go, makes it easier to understand while figuring things out.

twelho:
Review feedback assessed.

dhcpv4/nclient4/lease_test.go
```diff
@@ -236,15 +236,24 @@ func (sll *testServerLeaseList) runTest(t *testing.T) {
 		sll.lastTestSvrErrLock.RLock()
 		keepgoing := chkerr(err, sll.lastTestSvrErr, l.ShouldFail, t)
 		sll.lastTestSvrErrLock.RUnlock()
-		if keepgoing {
-			err = clnt.Release(lease)
```
pmazzini, line 240: What is the benefit of this modification?

twelho: Since we need to test `keepgoing` twice (once for `Renew` and again for `Relese` since we need stop going if either of them fails), doing it this way avoids all nesting and makes the two cases homogeneous. This code for `Release` hasn't been removed, it has just moved down...

dhcpv4/nclient4/lease_test.go
```diff
@@ -236,15 +236,24 @@ func (sll *testServerLeaseList) runTest(t *testing.T) {
...
+		if !keepgoing {
+			continue
+		}
+
+		err = clnt.Release(lease)
+		//this sleep is to make sure release is handled by server
+		time.Sleep(time.Second)
+		sll.lastTestSvrErrLock.RLock()
+		chkerr(err, sll.lastTestSvrErr, l.ShouldFail, t)
+		sll.lastTestSvrErrLock.RUnlock()
```
twelho, line 256: ...here.

dhcpv4/nclient4/lease_test.go
pmazzini: I guess my question was: Is there any problem with having it like this to keep it consistent?

```
		if keepgoing {
        		lease, err = clnt.Renew(context.Background(), lease.Offer)
			sll.lastTestSvrErrLock.RLock()
			keepgoing = chkerr(err, sll.lastTestSvrErr, l.ShouldFail, t)
			sll.lastTestSvrErrLock.RUnlock()
		}

		if keepgoing {
	        	err = clnt.Release(lease)
		        //this sleep is to make sure release is handled by server
        		time.Sleep(time.Second)
	        	sll.lastTestSvrErrLock.RLock()
	        	chkerr(err, sll.lastTestSvrErr, l.ShouldFail, t)
        		sll.lastTestSvrErrLock.RUnlock()
		}
```

twelho: Aha, I see now. Yes, I can refactor it to be that way. It might be a bit confusing with `keepgoing` being updated inside the first `if`, but nevertheless it works the same.

twelho: Fixed in 16f5a6e00347d22bc516d4f9c63485d25de4c852.

pmazzini (approved):
lgtm

dhcpv4/modifiers_test.go
```diff
@@ -39,9 +39,19 @@ func TestWithOptionModifier(t *testing.T) {
 	d, err := New(WithOption(OptDomainName("slackware.it")))
 	require.NoError(t, err)
 
-	dnOpt := d.DomainName()
```
pmazzini, line 42: Actually, this change is not related to the title of the PR. Do you mind having these changes separately?

twelho: I can remove them from this PR.

But at this point i would like to see a `CONTRIBUTING.md` or similar guideline explaining for example that PRs must _only_ contain minimal changes that are directly related to the description (and any other project policies). To be clear, I've got nothing against this policy (or any other for that matter), but the contribution workflow has to be clear to contributors _before_ they start their work to avoid wasting everyone's time with these tiny incremental reviews. (And no, it is not obvious as each project does things a bit differently.)

twelho:
Seems like there are no more comments, could this be merged @pmazzini?

I've submitted a follow-up changeset at https://github.com/insomniacslk/dhcp/pull/470 which would also be great to have reviewed and merged together with this. It is more straight forward and should hopefully be easier to review compared to this PR.

dhcpv4/nclient4/client.go
```diff
@@ -478,6 +478,21 @@ func (c *Client) Request(ctx context.Context, modifiers ...dhcpv4.Modifier) (lea
 	return c.RequestFromOffer(ctx, offer, modifiers...)
 }
 
+// Renew sends a lease renewal based on the given offer. RENEW requests only have a minor set of
+// changes to their options compared to SELECT requests as specified by RFC 2131, section 4.3.2.
+func (c *Client) Renew(ctx context.Context, offer *dhcpv4.DHCPv4, modifiers ...dhcpv4.Modifier) (*Lease, error) {
```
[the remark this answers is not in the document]
twelho: The reasoning is that it only requires information from the offer and not the ACK. Taking in a narrower interface allows for better flexibility. For example in Talos we only store the offer from of a lease as the ACK information is parsed right away (see [here](https://github.com/siderolabs/talos/pull/5897/files#diff-009d3627b4be29dee368cb384b84645b5cd590b8eb1febfd1e548599c0bb6d2fR529-R536)). If one has a full lease, they can still very natually call `Renew(ctx, lease.offer)`, as it makes sense to renew the lease based on the information that was initially offered to a client (such as the original IP).

Since `Renew` takes modifiers and is more closely related to `Request` et al. (since it returns a `Lease`), it made sense to assimilate with those functions instead of `Release`, which works quite differently. I consider `Request` and `Renew` to be a pair, for example in Talos the same procedure is used to call both.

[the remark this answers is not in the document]
twelho: This approach turned out to be insufficient with DHCP servers that return incomplete ACKs during lease renewals (so the lease cannot be rotated), so I have now changed `Renew` to live next to and work just like `Release` and take in the entire lease. I managed to do this in a relatively clean way that also allowed for some additional unit testing of `NewRenewFromOffer`.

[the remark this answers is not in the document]
twelho: The new `Renew` approach is in https://github.com/insomniacslk/dhcp/pull/469/commits/aeb78d02bd032611653400cd55e3331a1f839aa9 and the corresponding unit tests in https://github.com/insomniacslk/dhcp/pull/469/commits/feb14de874aeca08e22d0d0a2f7859315646289d.

twelho:
@pmazzini could this be reviewed and merged? I think I have addressed all of the feedback, and don't plan on doing any more changes. This PR and https://github.com/insomniacslk/dhcp/pull/470 are really blocking https://github.com/siderolabs/talos/pull/5897 now.

pmazzini (approved):

pmazzini:
@twelho Sorry for dropping this for a while. Is this ready to be merged?

twelho:
Yes, this PR is ready to be merged from my side. I will fix up https://github.com/insomniacslk/dhcp/pull/470 this weekend as well, apologies for the delays there as I've been quite busy the last couple of weeks.

dhcpv4/nclient4/lease.go
```diff
@@ -36,3 +37,41 @@ func (c *Client) Release(lease *Lease, modifiers ...dhcpv4.Modifier) error {
...
+	}
+	if response.MessageType() == dhcpv4.MessageTypeNak {
+		return &ErrNak{
+			Offer: lease.Offer,
+			Nak:   response,
+		}
+	}
+
+	// Update the ACK of the lease with the ACK of the latest renewal
+	lease.ACK = response
```
pmazzini, line 74: Thinking back about this a new Lease should have been returned instead of modifying the one passed by parameter.

pmazzini: @twelho Would you be able to follow up on this?

twelho: Sure, I'll send a PR for that in the next couple of days.

twelho: https://github.com/insomniacslk/dhcp/pull/478

pmazzini:
Looking back at this, should it have been `NewRenewFromAck` instead of `NewRenewFromOffer`?

twelho:
There's a problem with that, the server may decide to drop the client IP from the ACK response (see `ciaddr` [here](https://www.rfc-editor.org/rfc/rfc2131#section-4.3.1)), which means that a purely ACK-based solution would not know the IP to unicast the renewal request with.

pmazzini:
> There's a problem with that, the server may decide to drop the client IP from the ACK response (see ciaddr [here](https://www.rfc-editor.org/rfc/rfc2131#section-4.3.1)), which means that a purely ACK-based solution would not know the IP to unicast the renewal request with.

All you need in the Review is the `yiaddr`: https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/dhcpv4.go#L260
Am I missing something?

twelho:
Indeed, didn't look at the correct row. The change to `NewRenewFromAck` is now available here: https://github.com/insomniacslk/dhcp/pull/498

PR #470, opened 2022-07-19
dhcpv4: add a way to build an INFORM request from an offer

twelho:
Implements the `NewInformFromOffer` and `InformFromOffer` that allow sending INFORM requests based on the offer of an existing lease, similar to `NewRequestFromOffer` and `RequestFromOffer`. This is needed especially since the required `serverAddr` field is private, which hinders the external use of the current `NewInform` method.

Unit tests are provided for `NewInformFromOffer`, for both its broadcast and unicast variants according to [RFC 2131, Section 4.4.4](https://datatracker.ietf.org/doc/html/rfc2131#section-4.4.4).


twelho:
@pmazzini could this be reviewed and merged as well? As stated in https://github.com/insomniacslk/dhcp/pull/469 this is a blocker for https://github.com/siderolabs/talos/pull/5897 and thus the release of Talos v1.2.

dhcpv4/dhcpv4.go
```diff
@@ -248,6 +248,23 @@ func NewRequestFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error)
 	)...)
 }
 
+// NewInformFromOffer builds a new DHCPv4 Informational message from the given
+// offer with the specified hardware address.
+func NewInformFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) {
+	xid, err := GenerateTransactionID()
```
pmazzini, line 254: I think this can be removed. It is called in the `New` below.

twelho: Good catch, fixed.

dhcpv4/nclient4/client.go
```diff
@@ -524,6 +524,35 @@ func (c *Client) RequestFromOffer(ctx context.Context, offer *dhcpv4.DHCPv4, mod
 	return lease, nil
 }
 
+// InformFromOffer sends an Inform request and waits for a response.
+func (c *Client) InformFromOffer(ctx context.Context, offer *dhcpv4.DHCPv4, modifiers ...dhcpv4.Modifier) (*dhcpv4.DHCPv4, error) {
```
pmazzini, line 528: This looks like a copy of `RequestFromOffer` except it is using `NewInformFromOffer`.
Can it be abstracted out to reuse code?

dhcpv4/dhcpv4.go
```diff
@@ -248,6 +248,17 @@ func NewRequestFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error)
 	)...)
 }
 
+// NewInformFromOffer builds a new DHCPv4 Informational message from the given
+// offer with the specified hardware address.
+func NewInformFromOffer(offer *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) {
```
pmazzini, line 253: With this we end up having `NewInformForInterface`, `NewInform` and `NewInformFromOffer`.
Do we really need all of them?
Do we really need to create an Inform based on an offer?

dhcpv4/nclient4/client.go
twelho: It should be possible to reuse the `c.SendAndRead` part in the middle, so I'll do some refactoring.

twelho: The functionality has now been refactored into `TransactAckNak`.

dhcpv4/dhcpv4.go
twelho: You're right, we don't need all of them. The existing `NewInform` is sufficient, see the example in the updated description. I've removed `NewInformFromOffer`.

twelho:
I've now refactored this to allow using the existing `NewInform` instead of `NewInformFromOffer` (which has been removed) for sending INFORM requests from client applications. While sending the actual request was possible before using `SendAndRead`, its interface is too low-level for external consumers as it delegates internal DHCP tasks, such as verifying server identifiers and ACK/NAK, to the caller. `NewInform` (and others) can be made more usable by exposing the core logic of `RequestFromOffer`, which I have now moved to the new `TransactAckNak` function (better naming suggestions welcome). The example in the updated description shows that this approach provides sufficient abstraction so that client applications don't need to deal with transaction details.

dhcpv4/dhcpv4.go
```diff
@@ -222,8 +222,7 @@ func PrependModifiers(m []Modifier, other ...Modifier) []Modifier {
 // NewInform builds a new DHCPv4 Informational message with the specified
 // hardware address.
 func NewInform(hwaddr net.HardwareAddr, localIP net.IP, modifiers ...Modifier) (*DHCPv4, error) {
-	return New(PrependModifiers(
-		modifiers,
+	return New(PrependModifiers(modifiers,
```
pmazzini, line 225: Why is this changing?

dhcpv4/dhcpv4_test.go
```diff
@@ -355,6 +355,16 @@ func TestNewInform(t *testing.T) {
 	require.True(t, m.ClientIPAddr.Equal(localIP))
 }
 
+func TestDHCPv4NewInformWithModifier(t *testing.T) {
```
pmazzini, line 358: Any reason to add this new test if nothing changed in this code?

dhcpv4/nclient4/lease.go
```diff
@@ -65,8 +65,8 @@ func (c *Client) Renew(ctx context.Context, lease *Lease, modifiers ...dhcpv4.Mo
 	}
 	if response.MessageType() == dhcpv4.MessageTypeNak {
 		return &ErrNak{
-			Offer: lease.Offer,
-			Nak:   response,
+			Request: lease.Offer,
```
pmazzini, line 68: Why is this changing?

dhcpv4/dhcpv4.go
twelho: As mentioned in the updated description, to make it consistent with the neighboring functions such as `NewDiscovery`, `NewRequestFromOffer`, `NewRenewFromOffer` etc. All of them start with `return New(PrependModifiers(modifiers,`.

dhcpv4/dhcpv4_test.go
twelho: The use case in https://github.com/siderolabs/talos/pull/5897 (and the example in the updated description) requires supplying modifiers to `NewInform`, which has to date not been test-validated use case.

dhcpv4/nclient4/lease.go
twelho: `ErrNak` was previously used solely with request operations that supply an Offer, but this assumption is too restrictive for general use of a NAK error, thus the change from `Offer` to `Request`. This particular case however should pass in the request (which is a "superset" of the offer) instead of the offer directly. Fixed now.

dhcpv4/dhcpv4.go
pmazzini: I mean, nothing is changing in here other than removing a line. Is it needed? Can it be avoided to make this PR more clear?

dhcpv4/dhcpv4_test.go
pmazzini: Nothing changed in the `NewInform`. Do we really need a new test?

dhcpv4/nclient4/client.go
```diff
@@ -469,19 +469,35 @@ func (c *Client) DiscoverOffer(ctx context.Context, modifiers ...dhcpv4.Modifier
...
+		ACK:          response,
+		CreationTime: time.Now(),
+	}, nil
 }
 
 // ErrNak is returned if a DHCP server rejected our Request.
 type ErrNak struct {
-	Offer *dhcpv4.DHCPv4
-	Nak   *dhcpv4.DHCPv4
+	Request *dhcpv4.DHCPv4
```
pmazzini, line 499: Unless I am missing something, to me it only makes sense to store what the server returns.

dhcpv4/nclient4/lease.go
```diff
@@ -65,8 +65,8 @@ func (c *Client) Renew(ctx context.Context, lease *Lease, modifiers ...dhcpv4.Mo
 	}
 	if response.MessageType() == dhcpv4.MessageTypeNak {
 		return &ErrNak{
-			Offer: lease.Offer,
-			Nak:   response,
+			Request: request,
```
pmazzini, line 68: Unless I am missing something, to me it only makes sense to store what the server returns.

dhcpv4/dhcpv4.go
twelho: I can avoid it if desired, but in my opinion the tradeoff of making this PR slightly less clear in order to ultimately make the code clearer for all future contributors and maintainers is worth it. I felt like this change is topical for this PR instead of a separate one as I'm working on cleaning up `NewInform`, but in this end it is still your call if we keep it or remove it.

dhcpv4/dhcpv4_test.go
twelho: As I'm improving `NewInform` here, although I'm not changing its functionality (to maintain backwards compatibility), as a consumer of the function I would like to have strong guarantees that it functions properly in all advertised cases. Even though it is empirically known to work properly right now, the test guarantees that modifier support won't be accidentally broken in the future by either a direct or indirect change. I do also not think that `TestDHCPv4NewInformWithModifier` would measurably increase test runtime as it is very lightweight.

dhcpv4/nclient4/client.go
twelho: I can change `ErrNak` to only store `Nak`, but that is a more significant API change than renaming `Offer` to `Request` for any library consumer that might rely on the sent request being present in `ErrNak` for failure analysis etc. If you are sufficiently confident that this hasn't been a supported use case, then I'll go ahead and do the change.

dhcpv4/nclient4/lease.go
twelho: See https://github.com/insomniacslk/dhcp/pull/470#discussion_r967883390.

dhcpv4/dhcpv4_test.go
pmazzini: Is this strictly related to the title of this PR?

twelho: I've updated the title of this PR.

pmazzini: It would make more sense to me to have this in a separate PR instead of having such a long PR title :)

twelho: Right, I can do that as well if you prefer, I was just thinking that that causes a lot of very small PRs which might cause a lot of spam and be hard to keep track of. Regardless, I'll try to find the time in the next days to split this PR up so that every change that can be described in a short PR title gets its own PR.

twelho: Separated: https://github.com/insomniacslk/dhcp/pull/480

dhcpv4/dhcpv4.go
twelho: Separated: https://github.com/insomniacslk/dhcp/pull/481

dhcpv4/nclient4/lease.go
twelho: `Offer` has now been removed as per https://github.com/insomniacslk/dhcp/pull/470#discussion_r967865608.

dhcpv4/nclient4/client.go
twelho: `Offer` has now been removed.

dhcpv4/nclient4/lease.go
twelho: `Offer` has now been removed.

pmazzini:
> This branch has conflicts that must be resolved
> Conflicting files
> dhcpv4/nclient4/lease.go

twelho:
I've resolved the conflicts, could you take one more look at this (now that it's cleaned up) when you have time @pmazzini?

smira:
@pmazzini any chance you could review this one please?

dhcpv4/nclient4/client.go
```diff
@@ -469,19 +469,34 @@ func (c *Client) DiscoverOffer(ctx context.Context, modifiers ...dhcpv4.Modifier
...
+	return &Lease{
+		Offer:        offer,
+		ACK:          response,
+		CreationTime: time.Now(),
+	}, nil
 }
 
 // ErrNak is returned if a DHCP server rejected our Request.
 type ErrNak struct {
-	Offer *dhcpv4.DHCPv4
```
pmazzini, line 483: Why is this being removed?

twelho: `TransactAckNak` is more generic than `RequestFromOffer` for which that error type was originally defined, and does not take in an `offer` to pass into here. The original `offer` from `c.Request` is not passed into `TransactAckNak` (as it is unnecessary), thus I've omitted it from the `ErrNak` struct.

dhcpv4/nclient4/client.go
```diff
@@ -627,6 +609,35 @@ func (c *Client) SendAndRead(ctx context.Context, dest *net.UDPAddr, p *dhcpv4.D
...
+func (c *Client) TransactAckNak(ctx context.Context, request *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, error) {
+	matchers := []Matcher{IsMessageType(dhcpv4.MessageTypeAck, dhcpv4.MessageTypeNak)}
+
+	// Servers are supposed to only respond to Requests containing their server identifier,
+	// but sometimes non-compliant servers respond anyway.
+	// Clients are not required to validate this field. Servers are required to include
+	// the server identifier in their Offer per RFC 2131 Section 4.3.1 Table 3, but it is
+	// not mandatory for e.g. INFORM requests if the client does not explicitly request it.
+	identifier := request.ServerIdentifier()
+	if identifier != nil {
```
pmazzini, line 624: This logic is changing from what we had in `RequestFromOffer`. Is there any reason for it? When will a Servicer Indentifier be nil?

twelho: This change is to facilitate compatbility with INFORM requests. The comment right above the line to it explains the reason for why `ServerIdentifier` can be `nil`:

> Servers are required to include the server identifier in their Offer per RFC 2131 Section 4.3.1 Table 3, but it is not mandatory for e.g. INFORM requests if the client does not explicitly request it.

twelho: If the server has chosen not to send the identifier, it cannot be validated.

pmazzini: This is changing the previous behavior and relaxing the checks. If `RequestFromOffer` does not get a `ServerIdentifier` back, with this change it will not raise an error.

> If the server has chosen not to send the identifier

I thought this was not optional. Do you have any reference to the RFC about this?

pmazzini:
How about abstracting out the code you need to reuse from `RequestFromOffer` but without changing the `Request` function and still leaving a `RequestFromOffer` function.

pmazzini:
Why don't you create an `Inform` method instead of a `TransactAckNak` one?

dhcpv4/nclient4/client.go
twelho: You're right, it seems to not be optional. The server must always send the identifier for ACK/NAK as per RFC 2131 Section 4.3.1 Table 3, and DHCPINFORM follows the ACK parameters as defined in [Section 4.3.5](https://www.rfc-editor.org/rfc/rfc2131#section-4.3.5). Will fix.

twelho:
Good feedback, how about this:

- I've abstracted the needed ACK/NAK handling out to `TransactAckNak`, as is, from the original `RequestFromOffer`, and kept it so that `Request` needs no modifications. This should exactly match the existing behavior. 
- I've now created the `Inform` method in such a way that it should comply to the spec: the original case for making the server identifier check optional is that INFORM requests may not send the server identifier. However, as the identifier is now given to `TransactAckNak` separately, it can be correctly fed from the offer in case of both `Inform` and `Request` (which exactly matches the existing behavior).

Is this kind of structure acceptable?

twelho:
Description and example updated.

dhcpv4/nclient4/client.go
```diff
@@ -627,6 +628,29 @@ func (c *Client) SendAndRead(ctx context.Context, dest *net.UDPAddr, p *dhcpv4.D
 	return response, nil
 }
 
+// TransactAckNak exposes a higher-level API than SendAndRead for sending a given request (any type)
+// and receiving an ACK/NAK response. The server identifier is validated, responses are type checked,
+// and NAK responses are automatically returned as errors.
+func (c *Client) TransactAckNak(ctx context.Context, request *dhcpv4.DHCPv4, serverIdentifier net.IP) (*dhcpv4.DHCPv4, error) {
```
pmazzini, line 634: Does this need to be a public function?

dhcpv4/nclient4/client.go
```diff
@@ -478,10 +478,23 @@ func (c *Client) Request(ctx context.Context, modifiers ...dhcpv4.Modifier) (lea
 	return c.RequestFromOffer(ctx, offer, modifiers...)
 }
 
+// Inform sends an INFORM request based on information received in the given offer.
+// Returns the ACK response from the server on success, and an ErrNak error on failure.
+func (c *Client) Inform(ctx context.Context, offer *dhcpv4.DHCPv4, modifiers ...dhcpv4.Modifier) (*dhcpv4.DHCPv4, error) {
```
pmazzini, line 483: Is the inform always built from an offer?

dhcpv4/nclient4/client.go
twelho: Not necessarily, I've changed it to be private.

dhcpv4/nclient4/client.go
twelho: Essentially yes. The INFORM request requires the client to send its assigned IP in `ciaddr`, and, for consistency, we want to validate that the response to the inquiry comes from the same server that provided the offer (there was one mishap here which I've now fixed, INFORM requests are not allowed to contain the server identifier). Thus, taking in `offer` makes sense, since the client must have received an existing offer before being able to send INFORM requests (otherwise it should send a DISCOVER request).

dhcpv4/nclient4/client.go
```diff
@@ -502,21 +514,9 @@ func (c *Client) RequestFromOffer(ctx context.Context, offer *dhcpv4.DHCPv4, mod
...
 	}
 
-	// Servers are supposed to only respond to Requests containing their server identifier,
-	// but sometimes non-compliant servers respond anyway.
-	// Clients are not required to validate this field, but servers are required to
-	// include the server identifier in their Offer per RFC 2131 Section 4.3.1 Table 3.
-	response, err := c.SendAndRead(ctx, c.serverAddr, request, IsAll(
-		IsCorrectServer(offer.ServerIdentifier()),
-		IsMessageType(dhcpv4.MessageTypeAck, dhcpv4.MessageTypeNak)))
+	response, err := c.transactAckNak(ctx, request, offer.ServerIdentifier())
```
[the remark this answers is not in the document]
twelho: Sure, I can do that.

dhcpv4/nclient4/client.go
pmazzini: The RFC says the client may have obtained the address by some other means. Forcing to pass an offer here does not look right.

twelho: In that case the function needs to take in both the client IP and the server identifier for which it expects a response. I remember now that this is why I made the identifier check optional, but it is also OK for it to be mandatory. Should I keep the function and rename it to `InformFromOffer` and/or have `Inform` take in the client IP and server identifier directly (https://github.com/insomniacslk/dhcp/pull/470#discussion_r944275705)?

pmazzini: Actually this might not be bad - I'll take another look during the week

twelho: Any updates?

pmazzini: If we strict follow the RFC I think we should not force an offer as parameter here.

twelho: Alright, that would mean taking in the server identifier and the offered IP in `Inform`. It would also mean that there won't be anything sensible to put into the `Offer` field of `ErrNak`, should I leave it as `nil`?

twelho: Actually, in order to still use `transactAckNak`, I would need to make it process the server identifier independently of the offer if maintaining `ErrNak` as-is is desired...

pmazzini: Then something looks wrong and maybe we should not try to create a `transactAckNak` function.

In the A part of the DORA it makes sense to get it checked but not here.

twelho: Indeed, I combed through the spec one more time, and came to the same conclusion: INFORM requests are more akin to DISCOVERs that should not use previously acquired information to do validation. Since `c.serverAddr` is known, we can still use it to target the desired server using unicast as stated in the [spec](https://www.rfc-editor.org/rfc/rfc2131#section-4.4.3). Additionally, the only valid response to an INFORM request is an ACK, so we can indeed get rid of `transactAckNak`, which I've now done.

As a result, INFORM handling is now fully standalone. Does it make sense now?

dhcpv4/nclient4/client.go
```diff
@@ -478,6 +478,25 @@ func (c *Client) Request(ctx context.Context, modifiers ...dhcpv4.Modifier) (lea
 	return c.RequestFromOffer(ctx, offer, modifiers...)
 }
 
+// Inform sends an INFORM request using the given local IP.
+// Returns the ACK response from the server on success.
+func (c *Client) Inform(ctx context.Context, localIP net.IP, modifiers ...dhcpv4.Modifier) (*dhcpv4.DHCPv4, error) {
+	request, err := dhcpv4.NewInform(c.InterfaceAddr(), localIP, modifiers...)
```
pmazzini, line 484: Why not directly use `c.ifaceHWAddr` here?

pmazzini:
It is looking good now.

dhcpv4/nclient4/client.go
twelho: Good catch, updated.

twelho:
If everything's good now, can you merge?

pmazzini (approved):

PR #482, opened 2022-10-24
dhcpv6 - Fix DuidFromBytes so that it creates a copy of the underlying bytes

pcsegal:
This pull request changes `DuidFromBytes` so that it copies any pertinent data from the input bytes.
This is intended to make sure that the ClientID and ServerID option parsers make a copy of the pertinent data, since these option parsers make use of `DuidFromBytes`.


pmazzini:
@pcsegal I think this is no longer needed after https://github.com/insomniacslk/dhcp/pull/488.

Do you mind checking?

hugelgupf:
Indeed, this should be fixed as well.

PR #493, opened 2023-02-26
Improving test coverage before lazy parsing

hugelgupf:
Cover every part of message encoding and decoding.

Requires generics. Go 1.18.

New getters for 4RD & vendor class options


hugelgupf:
Stickler doesn't seem to like generics

pmazzini (approved):

PR #508, opened 2023-07-19
[Fuzzing] remove old go-fuzz and use Native Go fuzz

0x34d:
[go-fuzz](https://github.com/dvyukov/go-fuzz) is deprecated in favor of https://go.dev/security/fuzz

* using Native [Go fuzz](https://go.dev/security/fuzz)
* Add fuzzing seed.
* Support of oss-fuzz infrastructure.


Makefile
```diff
@@ -0,0 +1,4 @@
+fuzz:
```
pmazzini, line 1: Do we really need to commit this Makefile?

0x34d: Not actually

pmazzini: Should we run them as part of the CI?

0x34d: >Should we run them as part of the CI? 

My proposal is a little bit longer but it's better for the DHCP project.
See, the DHCP project is used by a lot of [projects](https://github.com/insomniacslk/dhcp#public-projects-that-use-it), so I can integrate this into [oss-fuzz](https://google.github.io/oss-fuzz/) and later [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) can be added.

Steps are:
* Merge this PR.
* Integrate oss-fuzz.
* Implement CIFuzz for DHCP PR.

README.md
```diff
@@ -42,6 +42,15 @@ packages.
 
 See more example code at https://github.com/insomniacslk/exdhcp
 
+# Fuzzing
```
pmazzini, line 45: Do we really need this? This is not different to what the `go-fuzz` docs mention.

0x34d: ReMoved

0x34d:
Request: @pmazzini, may I have your email ID for OSS-Fuzz `primary_contact`?

0x34d:
> Request: @pmazzini, may I have your email ID for OSS-Fuzz `primary_contact`?

USING : https://insomniac.slackware.it/pages/email

PR #516, opened 2023-10-15
fix misleading listen address in documentation

ForestJohnson:
A DHCP server should probably always listen on `0.0.0.0` when we're working on it in development and testing.

Why? 

I believe normal DHCP requests are sent as broadcasts to `255.255.255.255`...   I believe when its broadcasted, every single UDP packet regardless of where it ends up, will have `255.255.255.255` as the destination address. And since `255.255.255.255` != `127.0.0.1`, the UDP listener will ignore it. 

We can tell our UDP listener to listen on `255.255.255.255`  but then it will ONLY receive these "normal" DHCP broadcast packets.   If someone tries to send a datagram to it directly (on a specific IP like `127.0.0.1`,  `192.168.69.1` etc)  it wont be received. 

So i think `0.0.0.0` is appropriate here, considering that it's already being told to listen on a specific interface.


pmazzini:
Please sign the commit. More details in:
https://github.com/insomniacslk/dhcp/pull/516/checks?check_run_id=19055434699

pmazzini (approved):

PR #520, opened 2023-12-04
remove bsdp package

### LINKED ISSUE #116: Add modifiers to BSDP package
As done in dhcpv6 and dhcpv4, let's add modifiers to the packet generation. Assigned to @get9 as our BSDP expert :)

pmazzini:
New Mac's no longer support PXE booting. This code is no longer being used with DHCPLB. Fixes #116


pmazzini:
Fixes #116

insomniacslk:
older macs still support it though. @pmazzini what do we lose leaving this as it is?

pmazzini:
Older Macs have been phased out. I think nobody is making use of this code at the moment.

insomniacslk:
> Older Macs have been phased out. I think nobody is making use of this code at the moment.

Phased out as in they are not sold anymore? If so, I still expect that those devices will be around for years to come

pmazzini:
afaik the hardware is no longer being sold and the software running on it is no longer maintained / supported.

One benefit of cleaning this up is that we will have one less thing to migrate to the new DHCP client. (ref #459 where there was a previous attempt at removing part of this package).

I have't seen users using this code and would rather avoid leaving stale code around. I thought we would benefit from some cleanup but we can leave it if you think somebody may benefit from it.

insomniacslk:
Sounds good, thanks for the details. Go ahead!

insomniacslk (approved):

PR #524, opened 2024-02-25
Add option IPv6OnlyPreferred (RFC 8925)

candlerb:
This DHCP option is needed to support https://github.com/coredhcp/coredhcp/pull/170


pmazzini:
Don't you want to add parsing for this option as well to avoid having to deal with `OptGeneric` in the coredhcp code?

candlerb:
Good point. Helpers added, and tested with a corresponding patched coredhcp.

dhcpv4/dhcpv4.go
```diff
@@ -730,6 +730,22 @@ func (d *DHCPv4) IPAddressRebindingTime(def time.Duration) time.Duration {
 	return time.Duration(dur)
 }
 
+// IPv6OnlyPreferred returns true if this option is present, and the
+// V6ONLY_WAIT duration if it is.
+//
+// The IPv6-Only Preferred option is described by RFC 8925, Section 3.1.
+func (d *DHCPv4) IPv6OnlyPreferred() (bool, time.Duration) {
```
hugelgupf, line 737: I feel the usual pattern is <type>, ok, like in the case of map access.

candlerb: I noticed that MaxMessageSize uses (value, error)

However I'll change to (value, bool) as you suggest.

candlerb:
I think this is in shape for a recheck now, thanks.

dhcpv4/dhcpv4.go
```diff
@@ -730,6 +730,22 @@ func (d *DHCPv4) IPAddressRebindingTime(def time.Duration) time.Duration {
 	return time.Duration(dur)
 }
 
+// IPv6OnlyPreferred returns the V6ONLY_WAIT duration, and a boolean
+// indicating whether this option was present.
+//
+// The IPv6-Only Preferred option is described by RFC 8925, Section 3.1.
+func (d *DHCPv4) IPv6OnlyPreferred() (time.Duration, bool) {
```
pmazzini, line 737: Will this ever return 0, true? If not shouldn't we simply return time.Duration as in `IPAddressRebindingTime`?

dhcpv4/option_ipv6_only_preferred.go
```diff
@@ -0,0 +1,10 @@
+package dhcpv4
+
+import (
+	"time"
+)
+
+// The IPv6-Only Preferred option is described by RFC 8925, Section 3.1
+func OptIPv6OnlyPreferred(d time.Duration) Option {
```
pmazzini, line 8: Maybe we should add this option to the same file where the type `Duration` is defined and rename that file to a more appropriate name.

dhcpv4/dhcpv4.go
candlerb: Yes it makes a difference. Setting the option in the response with value 0, is a signal to the client that it can use IPv6-only for an unlimited time (i.e. never check back for DHCPv4)

Two other possibilities I can think of:
1. Return some sentinel value for -1 meaning "this option was not set" (that's ugly)
2. return (value, error) with a non-nil error if the option is not supplied. That's how `DHCPv4.MaxMessageSize()` works.

dhcpv4/option_ipv6_only_preferred.go
candlerb: Merged as `option_duration.go`

dhcpv4/dhcpv4.go
pmazzini: > Yes it will, and it makes a difference. Setting the option in the response with value 0, is a signal to the client that it can use IPv6-only for an unlimited time

From the RFC I get something different but what matters is that having a zero or not having it makes a difference:

```
If the IPv6-Only Preferred option returned by the server contains a value greater than or equal to MIN_V6ONLY_WAIT, the client SHOULD set the V6ONLY_WAIT timer to that value. Otherwise, the client SHOULD set the V6ONLY_WAIT timer to MIN_V6ONLY_WAIT.
```

pmazzini (dismissed):
lgtm

dhcpv4/dhcpv4.go
candlerb: Yes, on re-reading you are correct. I note that section 3.4 says that V6ONLY_WAIT has a "default" of 1800 seconds; and yet it says twice (in 3.1 and 3.3) that if the server has no explicit configured value then it *must* send a value of zero, and if the client receives a value less than MIN_V6ONLY_WAIT then it should use MIN_V6ONLY_WAIT. So I cannot see any circumstance in which this default value of 1800 would be used.

It might be reasonable to clip the returned value to MIN_V6ONLY_WAIT - although it's a "should" not a "must".

In any case: do you prefer "ok bool" or "err error" for the presence indication?

pmazzini: > In any case: do you prefer "ok bool" or "err error" for the presence indication?

I have no preference.

candlerb: I don't even need it for coredhcp server; only a client needs to look for the presence of the option (the client never sends it to server, only includes it in parameter request list).

As a client, I would probably not care about the difference between a missing option versus a malformed option (and if I did, I would need a sentinel error for "missing").

I'll stick with bool then, but I will change #525 for autoconfigure option to be consistent.

pmazzini (approved):

candlerb:
Thank you!

PR #525, opened 2024-02-25
Option helper for AutoConfigure

candlerb:
This isn't really a fully-fledged pull request: I'm seeking advice on how best to expose the option AutoConfigure from RFC 2563.

What the proposed patch does is implement a Byte option, based on Uint16 from `dhcpv4/option_maximum_dhcp_message_size.go`

However, other possibilities would be:
1. Implement a Bool option
2. Implement an enumeration which is specific to AutoConfigure - so that for example, the String() method would return "Auto-Configure: DoNotAutoConfigure" or "Auto-Configure: AutoConfigure"
3. Do nothing and require clients to use GenericOption directly. However there could be a small helper for setting this generic option, along the lines of `OptClientIdentifier()` in `dhcpv4/option_misc.go`

Based on the [registry](https://www.iana.org/assignments/bootp-dhcp-parameters/bootp-dhcp-parameters.txt), some other examples of single-byte DHCP options are:
* 19 Forward On/Off (RFC 2132)
* 20 SrcRte On/Off (RFC 2132)
* 23 Default IP TTL (RFC 2132)
* ... various other obsolete-looking options with length 1
* Weirdly, option 116 (Auto-Config) is described as having a length of N, not 1.


dhcpv4/dhcpv4.go
```diff
@@ -737,6 +737,13 @@ func (d *DHCPv4) MaxMessageSize() (uint16, error) {
 	return GetUint16(OptionMaximumDHCPMessageSize, d.Options)
 }
 
+// AutoConfigure returns the AutoConfigure option if present.
+//
+// The AutoConfigure option is described by RFC 2563, Section 2.
+func (d *DHCPv4) AutoConfigure() (byte, error) {
```
hugelgupf, line 743: I would define a type for this.

```
type AutoConfiguration byte

const (
  DoNotAutoConfigure AutoConfiguration = 0
  AutoConfigure AutoConfiguration = 1
)
```

AutoConfigure should return that.

It should have a stringer that prints "do not configure" or "configure" so that when Message.String() is used it doesn't just print "0" or "1"

hugelgupf: I wouldn't mind us using generics to make this nicer at some point, but if you're just defining this one, let's do this.

candlerb:
Ready for review

candlerb:
I have changed the AutoConfigure() helper to return (value, bool) instead of (value, error). I do need this one for coredhcp.

Please could you kick off the CI, and if it's happy, I'm good to merge this now.

dhcpv4/options.go
```diff
@@ -344,6 +344,10 @@ func getOption(code OptionCode, data []byte, vendorDecoder OptionDecoder) fmt.St
 		var u Uint16
 		d = &u
 
+	case OptionAutoConfigure:
```
pmazzini, line 347: NIT: maybe sort by option code here

candlerb: Moved OptionAutoConfigure to the right place (but not all the other options are in option code sequence)

pmazzini (approved):

Issue #526, opened 2024-03-11
Build-scoping issue in dhcpv4/nclient4/conn_unix.go (also: BSD support for nclient4)


vista-:
Hey folks,

I recently ran into an issue trying to build code that imports `nclient4` on macOS/BSD-likes:

```
% GOOS=freebsd go build .
../../go/pkg/mod/github.com/insomniacslk/dhcp@v0.0.0-20240227161007-c728f5dd21c8/dhcpv4/nclient4/conn_unix.go:42:59: undefined: unix.ETH_P_IP
```

https://github.com/insomniacslk/dhcp/blob/c728f5dd21c872dcb78cba11556619fdd7b6fc20/dhcpv4/nclient4/conn_unix.go#L42

Upon looking this symbol up, I found that it is build-scoped to Linux platforms only: https://cs.opensource.google/go/x/sys/+/master:unix/zerrors_linux.go;drc=5710a32031f3498e357d801987a07d9638c4fa11;l=1091

which kind of conflicts with the build-scoping of the file itself:
https://github.com/insomniacslk/dhcp/blob/c728f5dd21c872dcb78cba11556619fdd7b6fc20/dhcpv4/nclient4/conn_unix.go#L7

Would it be possible to move this specific function into a Linux build-scoped file?

To give you some context: I'm working on a multi-platform DHCP "swiss knife", and I added support for BSD-likes in my code using the (now deprecated) `github.com/mdlayher/raw` module that you migrated off of. I couldn't find anything like it, and it seems like those wanting to use raw BSD network sockets are stuck with it for now.

If you folks happen to be interested in this code even with this caveat, please let me know and I'll work on "upstreaming" BSD support for the DHCPv4 client. :)

hugelgupf:
I'm for it - both the Linux specific thing and upstreaming

PR #534, opened 2024-05-23
Feature: nclient4 BSD/macOS support

vista-:
This PR adds back BSD/macOS support for nclient4, which was removed when the library switch from using the `raw` package for sending/receiving broadcast UDP packets, instead using the `packet` library (which only supports Linux).

The PR also solves the build-scoping issue described in https://github.com/insomniacslk/dhcp/issues/526

Due to lack of support for raw datagram sockets in BSD, additional functionality had to be added for constructing the raw Ethernet frame. VLAN tag support has also been added to the BSD/macOS-related nclient4 code, which can be utilised to send/receive tagged frames. This can be useful for testing IPoE-based access networks.

Tests have also been added for the Ethernet frame and VLAN tag processing, based on the `gopacket` library.

There are two avenues of improving the code in the PR in the future (but I would prefer to avoid premature optimisation):
- Create a byte array with the byte representation of the VLAN stack + EtherType field in it during initialisation, in `NewRawUDPConn`. This can be inserted at the EtherType offset of the Ethernet frame when sending, and performing a simple byte comparison at the EtherType offset when receiving frames.
- Use `gopacket` for constructing the frame from scratch instead of doing it byte by byte.


vista-:
Force pushed commits with sign-offs.

ggidofalvy-tc:
Ping! :)

vista-:
Ping! :)

PR #535, opened 2024-05-23
Make Option 82 come last

michaelharo:
https://datatracker.ietf.org/doc/html/rfc3046#section-2.1 states:

```
A DHCP relay agent adding a Relay Agent Information field SHALL add
it as the last option (but before 'End Option' 255, if present) in
the DHCP options field of any recognized BOOTP or DHCP packet
forwarded from a client to a server.
```

### DIFF
--- a/dhcpv4/options.go
+++ b/dhcpv4/options.go
@@ -110,8 +110,9 @@ func (o Options) FromBytes(data []byte) error {
 }
 
 const (
-	optPad = 0
-	optEnd = 255
+	optPad       = 0
+	optAgentInfo = 82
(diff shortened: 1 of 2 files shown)


pmazzini:
Can you sign the commit?

https://github.com/insomniacslk/dhcp/pull/535/checks?check_run_id=25338914949

michaelharo:
I think I did that.

insomniacslk:
The code looks good to me but I can't figure out why there is only one check being run

insomniacslk (approved):

PR #536, opened 2024-05-29
Improve support for renewal and rebinding options

michaelharo:


pmazzini (dismissed):

michaelharo:
I'm looking for some feedback on this request. For example, I couldn't decide if the functions should include "Value" in their name.

`dhcpv4.OptRenewTimeValue(...)` or `dhcpv4.OptRenewTime(...)`

Also the access functions on `DHCPv4` already exist and are named like `IPAddressRenewalTime`.  Would it be more clear if the option creation helpers were named with an `IPAddress` prefix so like `OptIPAddressRenewalTime` (also `renew` vs `renewal`?

pmazzini (approved):

michaelharo:
What's the next step for getting this request merged?

PR #537, opened 2024-06-03
misuse of 'err' to check nil

youngifif:
In ConfigureInterface(),the defer closure misuse outter 'err' to check nil ,not 'cerr' inside the closure.

### DIFF
--- a/netboot/netconf.go
+++ b/netboot/netconf.go
@@ -192,7 +192,7 @@ func ConfigureInterface(ifname string, netconf *NetConf) (err error) {
 		return err
 	}
 	defer func() {
-		if cerr := rt.Close(); err != nil {
+		if cerr := rt.Close(); cerr != nil {
 			err = cerr
 		}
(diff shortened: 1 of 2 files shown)


pmazzini (approved):

pmazzini:
Integration tests look to be constantly failing now.

youngifif:
> Integration tests look to be constantly failing now.

I think the testCase in  TestConfigureInterface is not correct.

I use Debian12 official iproute2 tools to input command as follow:
`
sudo ip addr add 10.20.30.40 dev eth0
`
`
sudo ip addr add 10.20.30.40 dev eth0
`

then i got:
`
RTNETLINK answers: File exists
`
apparently netlink doesn't allow add the same address to one dev multiple times.

youngifif:
I was misleading by `cerr`, the original one is correct.
I gonna issue a new issue PR about test case

PR #543, opened 2024-07-25
Add relaxed padding setting to the nclient4

Bogdan-Denisenko:
RFC 2132 says:
`The end option marks the end of valid information in the vendor field.  Subsequent octets should be filled with pad options.`

But it seems that this does not mean that the client should consider packages with non-pad options after the End option invalid, because RFC 2119 explains:
`3. SHOULD   This word, or the adjective "RECOMMENDED", mean that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course.`

So we need a way to accept packets not padded with zeros.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -299,7 +299,15 @@ func NewReleaseFromACK(ack *DHCPv4, modifiers ...Modifier) (*DHCPv4, error) {
 
 // FromBytes decodes a DHCPv4 packet from a sequence of bytes, and returns an
 // error if the packet is not valid.
+// Octets after End option are checked against the pad option.
 func FromBytes(q []byte) (*DHCPv4, error) {
+	return FromBytesWithRelaxedPadding(q, false)
+}
+
+// FromBytesWithRelaxedPadding decodes a DHCPv4 packet from a sequence of bytes, and returns an
+// error if the packet is not valid.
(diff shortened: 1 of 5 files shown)


insomniacslk:
For clarity, maybe call it non-zero padding rather than relaxed padding?

pmazzini:
I am not sure if we should make this an optional parsing or if it should make it the default one. cc @hugelgupf

pmazzini:
> I made it optional just to avoid changing the behavior of the exported functions

Which functions? What is the benefit of having 2 behaviors?

Bogdan-Denisenko:
> Which functions? What is the benefit of having 2 behaviors?

Functions `dhcp4.FromBytes()` and `dhcp4.Options.FromBytes()`. In my opinion it's better to just remove the check for zero-padded packets or at least make such behavior default, because it seems that such packets should be considered valid according to RFC and I don't see any benefits from the possibility to consider such packets invalid. But this would change the behavior of exported (public) functions. Is it consistent with the backward compatibility policy of this project?

Bogdan-Denisenko:
@pmazzini Opened alternative PR with zero padding checking disabled by default - https://github.com/insomniacslk/dhcp/pull/545.

pmazzini:
Continuation in #545

PR #550, opened 2024-10-01
fix: add CLOEXEC flag to socket created directly

smira:
Go runtime adds `CLOEXEC` flags automatically, but when using raw syscalls, we have to do it manually.

Without this flag, file descriptors leak to the child processes.

### DIFF
--- a/internal/xsocket/xsocket.go
+++ b/internal/xsocket/xsocket.go
@@ -1,0 +1,34 @@
+package xsocket
+
+import (
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
+// CloexecSocket creates a new socket with the close-on-exec flag set.
+//
+// If the OS doesn't support the close-on-exec flag, this function will try a workaround.
+func CloexecSocket(domain, typ, proto int) (int, error) {
+	fd, err := socketCloexec(domain, typ, proto)
+	if err == nil {
+		return fd, nil
+	}
+
+	if err == unix.EINVAL || err == unix.EPROTONOSUPPORT {
+		// SOCK_CLOEXEC is not supported, try without it, but avoid racing with fork/exec
+		syscall.ForkLock.RLock()
+		defer syscall.ForkLock.RUnlock()
+
+		fd, err = unix.Socket(domain, typ, proto)
+		if err != nil {
+			return -1, err
+		}
+
+		unix.CloseOnExec(fd)
+
+		return fd, nil
+	}
+
+	return fd, err
+}
--- a/dhcpv4/client4/client.go
+++ b/dhcpv4/client4/client.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/insomniacslk/dhcp/internal/xsocket"
 	"golang.org/x/net/ipv4"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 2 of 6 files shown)


internal/xsocket/xsocket.go
```diff
@@ -0,0 +1,36 @@
...
+// CloexecSocket creates a new socket with the close-on-exec flag set.
+//
+// If the OS doesn't support the close-on-exec flag, this function will try a workaround.
+func CloexecSocket(domain, typ, proto int) (int, error) {
+	fd, err := unix.Socket(domain, typ|unix.SOCK_CLOEXEC, proto)
+	if err == nil {
+		return fd, nil
+	}
+
+	if err == unix.EINVAL || err == unix.EPROTONOSUPPORT {
```
Natolumin, line 18: `EINVAL` seems awfully wide an error to catch for this I think, especially with having to take the forklock in the fallback case. 
Can this be further narrowed per-platform with build tags? Afaict linux and all the go-supported bsds support `SOCK_CLOEXEC`, so the only actually relevant platform would be darwin, non-illumos solaris (before 11.4 according to [net source](https://github.com/golang/go/blob/bae2e968e2daadd39c1bdb1221648361d7277ddc/src/net/sock_cloexec_solaris.go#L7) ), maybe AIX?
It looks like `GOOS=darwin` doesn't even define `unix.SOCK_CLOEXEC` so this workaround won't even compile there (would it be necessary? I'm not sure)

I'd be inclined to not support it at all, I don't think any of these platforms are tested at all anyway

Natolumin (commented):
Nice catch! An inline note about the workaround where I would rather avoid messing with fork as much as possible

internal/xsocket/xsocket.go
smira: Darwin should support CLOEXEC I believe starting with some version, but I haven't tested it. 

There's a similar workaround in Go runtime, and this code is loosely based on https://github.com/mdlayher/socket/blob/9c51a391be6c6bc5b7ce52f328497931c5e97733/conn.go#L254-L303

But I'm happy to drop it and simply do `| unix.O_CLOEXEC`

smira: yes, `SOCK_CLOEXEC` is missing on Darwin, but other constants are also missing, so not sure how much relevant that is:

```
# github.com/insomniacslk/dhcp/dhcpv4/nclient4
dhcpv4/nclient4/conn_unix.go:42:59: undefined: unix.ETH_P_IP
```

smira: @Natolumin do you want me to simplify and just add CLOEXEC directly?

Orzelius: @smira I'm running into this not working on darwin whilst working on https://github.com/siderolabs/talos/issues/10537. FD_CLOEXEC / O_NOINHERIT should be available on nearly all systems so you could refactor this to use them or add CLOEXEC directly

smira: sounds good, let me try to update this

smira: reworked it a bit, now should be closer to what Go stdlib does

Orzelius: Thanks! I've tested this on my branch on mac and it works lovely!

Natolumin: Since this is closely based on the version from mdlayher/socket, and we already import that library in this one, can we simply use that `[socket.Socket](https://github.com/mdlayher/socket/blob/f82fb00aacf0230ab7c5e60e67176bdd178dd027/conn.go#L237)` implementation in here instead of copying it?

While I'm not an active maintainer, I don't think anyone here wants to have to maintain cross-platform socket-API level code for platforms that most of the maintainers don't use and can't test on, so if we can use that upstream lib implementation this would be much easier to merge imo

smira: I tried that, but the problem is that `mdlayher`'s socket is a wrapper around raw file descriptors, so this will be a cascading change all over the codebase. probably a change for good, but I don't know if it will be accepted or not, as it certainly might break in subtle ways

Natolumin (dismissed):
Code lgtm, but I haven't tested on any platform without cloexec, only linux so far

Natolumin (approved):

PR #555, opened 2025-01-08
Add WithContext versions of functions that use contexts.

marksuter:
These new functions are equivalent, but allow callers to pass a context down from main (or equivalent).  This is necessary to comply with certain style guides, for example, https://google.github.io/styleguide/go/decisions#contexts.

We considered replacing dhcpv4.GenerateTransactionID with rand.Read(xid[:]) based on the godoc saying that Read call "never returns an error", but this package is used in embedded situations where it can fail.

### DIFF
--- a/dhcpv4/dhcpv4.go
+++ b/dhcpv4/dhcpv4.go
@@ -146,24 +146,7 @@ func New(modifiers ...Modifier) (*DHCPv4, error) {
 	if err != nil {
 		return nil, err
 	}
-	d := DHCPv4{
-		OpCode:        OpcodeBootRequest,
-		HWType:        iana.HWTypeEthernet,
-		ClientHWAddr:  make(net.HardwareAddr, 6),
-		HopCount:      0,
-		TransactionID: xid,
-		NumSeconds:    0,
-		Flags:         0,
-		ClientIPAddr:  net.IPv4zero,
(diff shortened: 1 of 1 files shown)


dhcpv4/dhcpv4.go
```diff
@@ -133,8 +139,8 @@ func GenerateTransactionID() (TransactionID, error) {
 }
 
 // New creates a new DHCPv4 structure and fill it up with default values. It
-// won't be a valid DHCPv4 message so you will need to adjust its fields.
-// See also NewDiscovery, NewRequest, NewAcknowledge, NewInform and NewRelease.
+// won't be a valid DHCPv4 message so you will need to adjust its fields. See
+// also NewDiscovery, NewRequest, NewAcknowledge, NewInform and NewRelease.
 func New(modifiers ...Modifier) (*DHCPv4, error) {
```
hugelgupf, line 144: can you make New be a wrapper of NewWithContext as well? just so changes to one aren't forgotten in the other.

marksuter: Done with [e0b42be](https://github.com/insomniacslk/dhcp/pull/555/commits/e0b42bed27c5d6babfb99dc400d5c924575a13ed).

hugelgupf (dismissed):

hugelgupf (approved):

PR #559, opened 2025-03-26
Support microsoft-conform user class option in DHCPv6

damyan:
This pull request addresses an issue in the DHCPv6 User Class option (`OptUserClass`) where the `FromBytes` method failed to parse data in the Microsoft-compatible format. The existing implementation adhered strictly to the RFC 8415 specification, which requires each user class to be preceded by a 16-bit length field. However, Microsoft DHCPv6 implementations deviate from this by providing a single user class as raw bytes without a length prefix. This mismatch caused a "buffer too short" error when parsing Microsoft-formatted data, as observed in the otherwise failing test `TestOptUserClassBroken`.

## Changes Made
- **Updated `FromBytes` Method:**
  - The method now attempts to parse the data as multiple user classes with length prefixes (per RFC standards).
  - If a length field exceeds the remaining buffer size, parsing halts at that point.
  - If no valid user classes are parsed (e.g., due to missing or invalid length prefixes), the entire data buffer is treated as a single user class, aligning with Microsoft’s format.
  - This dual-parsing approach ensures compatibility with both RFC-compliant and Microsoft-specific inputs without raising errors.

## Why This Fix Is Necessary
- **Interoperability:** Supporting both RFC and Microsoft formats is essential for seamless interaction with a wide range of DHCPv6 clients, including those in Microsoft environments.
- **Test Reliability:** The previous implementation caused `TestOptUserClassBroken` to fail unnecessarily. This change eliminates the error while maintaining functionality for standard inputs.


dhcpv6/option_userclass.go
```diff
@@ -38,16 +38,26 @@ func (op *OptUserClass) String() string {
 	return fmt.Sprintf("%s: [%s]", op.Code(), strings.Join(ucStrings, ", "))
 }
 
-// FromBytes builds an OptUserClass structure from a sequence of bytes. The
-// input data does not include option code and length bytes.
+// FromBytes builds an OptUserClass structure from a sequence of bytes.
+// The input data does not include option code and length bytes.
```
insomniacslk: Please add a comment here that explains that the function is able to parse both RFC-compliant options and MS-compatible, non-RFC-compliant options

PR #568, opened 2025-12-19
Support windows for dhcp4

QQ2017:
**Description**
This PR adds Windows platform support for the DHCPv4 client packages (client4 and nclient4), as well as necessary supporting changes to xsocket, netboot, and interfaces packages.

**Background**
The library previously relied heavily on Unix-specific APIs (AF_PACKET, SOCK_RAW, unix.Socket, SO_BINDTODEVICE, etc.) which are not available on Windows. This prevented the library from being compiled or used on Windows.

**Changes**
  1. internal/xsocket
  Added //go:build !windows constraint to existing Unix files
  Created xsocket_windows.go with stub implementation
  2. dhcpv4/client4
  Added //go:build !windows constraint to client.go
  Created client_windows.go using standard net.ListenUDP for DHCP exchange
  3. dhcpv4/nclient4
  Created conn_windows.go using standard net.ListenUDP
  4. netboot
  Added //go:build !windows constraint to netconf.go
  Created netconf_windows.go with:
  Full implementation of GetNetConfFromPacketv4 and GetNetConfFromPacketv6
  IfUp using net.InterfaceByName
  ConfigureInterface stub (returning error, as this requires Windows-specific APIs)
  5. interfaces
  Modified bindtodevice_windows.go to return nil (no-op) instead of an error, since Windows implementation uses transaction ID filtering rather than interface binding

**Technical Notes**
Important: The Windows implementation uses standard UDP sockets (net.ListenUDP) instead of raw sockets with ipv4.PacketConn.SetControlMessage, because SetControlMessage is [not implemented on Windows](https://github.com/golang/go/issues/20153) in the golang.org/x/net/ipv4 package.

This means:

The client listens on all interfaces (binding to 0.0.0.0)
Responses are filtered by Transaction ID (which is the standard DHCP behavior)
This is functionally correct for most use cases

**Testing**
Tested on Windows 10/11 with DHCPv4:

  ✅ Compilation successful
  ✅ DORA exchange (Discover → Offer → Request → Ack) works correctly
  ✅ IP address obtained from DHCP server
  Attempt 1 of 4
  DISCOVER → server
  OFFER ← 10.180.2.11
  REQUEST → server  
  ACK ← 10.180.2.11 ✅

**Compatibility**
No breaking changes for existing Unix/Linux/macOS users
All existing tests should continue to pass on Unix platforms
server4 already had Windows support (conn_windows.go), so no changes needed there

**Related**
Fixes compilation on Windows
Enables DHCPv4 client usage on Windows without external dependencies (no Npcap/WinPcap required)

Signed-off-by: QQ2017 <qq10531348@gmail.com>


QQ2017:
@insomniacslk

QQ2017:
@pmazzini

pmazzini:
This feels created by an llm. Was that the case? We don't need support for both `client` and `nclient` do we?

PR #582, opened 2026-07-18
Add guard for when offer is nil

### LINKED ISSUE #581: nclient4.Client.Renew panics when Lease.Offer is nil (when calling lease.Offer.ServerIdentifier())
In `dhcpv4/nclient4/lease.go` at `(*Client).Renew`, there is a missing guard to check if `lease.Offer == nil`. This causes a nil pointer panic when trying to call lease.Offer.ServerIdentifier().

This error occurred due to a bug on my part, but a guard is needed.

ed6767:
fixes #581

### DIFF
--- a/dhcpv4/nclient4/lease.go
+++ b/dhcpv4/nclient4/lease.go
@@ -47,6 +47,10 @@ func (c *Client) Renew(ctx context.Context, lease *Lease, modifiers ...dhcpv4.Mo
 		return nil, fmt.Errorf("lease is nil")
 	}
 
+	if lease.Offer == nil {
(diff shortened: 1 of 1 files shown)


thc1006:
Offer guard looks right. The DCO check is red though, so this needs a `git commit --amend -s` and a force-push before it can merge. One more thing: ACK has the same nil problem. NewRenewFromAck and NewReleaseFromACK read ack.YourIPAddr before anything else, so a nil ACK panics in Renew, and Release too. Might be worth guarding lease.ACK by the Offer check, with a quick nil-field test.

PR #586, opened 2026-08-06
netboot: fix swallowed errors in ConfigureInterface

michaelpesa:
The deferred `rt.Close()` handler checked the wrong variable (`err`, the named return, instead of `cerr`). Any error returned from the function body was overwritten by the (almost always `nil`) `rt.Close()` result. Override the named return error only when it is `nil`.


pmazzini:
Is this something we can unit test?

